PR #237, opened 2026-04-01
fix: CLI crash + auto-generate YAML config templates from schema

viraatc:
## Summary

- **Bug fix**: CLI crash on `--load-pattern` + `--target-qps` (`IndexError`) — `LoadPattern.type` used `alias=` instead of `name=` on `cyclopts.Parameter`
- **Template generation**: `regenerate_templates.py` auto-generates minimal + full YAML templates from Pydantic schema defaults
- **Pre-commit hook**: auto-regenerates templates on schema/config changes; CI validates via `--check` mode

### Templates

Two variants per mode (offline, online, concurrency):
- `_template.yaml` — minimal: only required fields + `<PLACEHOLDER eg: example>` values
- `_template_full.yaml` — all fields with schema defaults, inline `# options:` comments auto-discovered from Enum/Literal annotations, 2 dataset examples (perf + accuracy with parser columns), multiple endpoints

### Init command

`inference-endpoint init offline` always generates the full template. Available types: `offline`, `online`, `concurrency`.

### Tests

- **Hypothesis fuzz**: 4000 random CLI flag combinations across offline/online/concurrency
- **Unit**: all 8 templates (including eval/submission) validate via `from_yaml_file` (auto-discovered with glob)
- **Integration (e2e)**: every generated template is loaded, `<>` placeholders stripped to real values, and run end-to-end against the echo server — verifying templates produce functional configs
- **Init command**: generates template + tested

### Docs updated

- AGENTS.md, CLI_QUICK_REFERENCE.md, DEVELOPMENT.md — all stale template/init references fixed

## Test plan

- [x] `python scripts/regenerate_templates.py --check` passes
- [x] `pytest tests/unit/` — 484 passed
- [x] `pytest tests/integration/commands/test_benchmark_command.py::TestTemplateIntegration` — 6 templates run e2e
- [x] Pre-commit hooks pass (ruff, mypy, prettier, template regeneration)


arekay-nv (approved):
Can we also add a test for this - seems like a change that shouldn't have gone in.

viraatc (commented):
added:
- fuzz tests on CLI
- regenerate default template pre-commit (prev. was just validaing)
- added schema-updated CI job to run fuzz tests and validate tempaltes
- template now includes all supported fields

viraatc (commented):
duplicate

viraatc (commented):
added new schema-updated CI:
1. fuzz tests on CLI in CI 
2. template validated against schema default in CI  

NOTE: template now includes all supported fields

2. was pending items from past.
++ @rashid for thoughts?

nvzhihanj (approved):

PR #275, opened 2026-04-10
feat: move to uv from pip

### LINKED ISSUE #274: feat: Move to uv
Moving to uv will enable better dependency pinning and user experience. 
Plan is to keep limited pip support, but migrate primary functionality to uv.

viraatc:
## What does this PR do?

close #274 

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [x] All tests pass locally

## Checklist

- [x] Pre-commit hooks pass


scripts/Dockerfile.dev
```diff
@@ -5,20 +5,26 @@
...
-    apt-get install -y  --no-install-recommends build-essential procps \
+    apt-get install -y --no-install-recommends build-essential procps \
     && rm -rf /var/lib/apt/lists/*
 
 RUN mkdir /mnt/inference-endpoint
 WORKDIR /mnt/inference-endpoint
-COPY pyproject.toml .
+
+# Copy lockfile + project metadata first for Docker layer caching
+COPY pyproject.toml uv.lock .python-version ./
```
viraatc: [Codex] **critical** (bug): Missing `COPY README.md ./` — `uv_build` requires `README.md` for metadata.

`pyproject.toml` declares `readme = "README.md"` (line 23), so when `uv sync` invokes the build backend, it needs `README.md` to be present. The current `COPY` instructions only bring in `pyproject.toml`, `uv.lock`, `.python-version`, and `src/`, so `docker build -f scripts/Dockerfile.dev .` will fail with a metadata generation error.

**Fix:** Add `COPY README.md ./` before or alongside the existing COPY lines:
```dockerfile
COPY pyproject.toml uv.lock .python-version README.md ./
```

pyproject.toml
```diff
@@ -1,6 +1,20 @@
...
+environments = [
+    "sys_platform == 'linux' and platform_machine == 'x86_64'",
+    "sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "sys_platform == 'darwin' and platform_machine == 'x86_64'",
+    "sys_platform == 'darwin' and platform_machine == 'arm64'",
+]
+
+[tool.uv-build]
+module-root = "src"
+data = {"inference_endpoint" = ["config/templates/*.yaml"]}
```
viraatc: [Claude] **medium** (bug): `data` key installs files to the wheel's data scheme (`sys.prefix/`), not inside the Python package.

The previous setuptools config used `[tool.setuptools.package-data]` which bundled YAML templates *inside* the `inference_endpoint` package at `site-packages/inference_endpoint/config/templates/`. The `data` key in `[tool.uv-build]` instead places them in the data scheme.

`commands/init.py` looks for templates via `Path(__file__).parent.parent / "config" / "templates"` (site-packages path). If uv_build's default VCS-tracked file inclusion already puts them in the package directory, then the `data` key is redundant (and creates a duplicate in the data scheme). If not, the templates won't be found by `init.py`.

**Action:** Verify with `uv build && unzip -l dist/*.whl | grep yaml` that template files appear under `inference_endpoint/config/templates/` in the wheel (not just `.data/data/`).

.pre-commit-config.yaml
```diff
@@ -62,3 +62,11 @@ repos:
         types: [python]
         pass_filenames: true
         exclude: ^(src/inference_endpoint/openai/openai_types_gen.py)$
+
+      - id: uv-lock-check
+        name: Check uv.lock is up-to-date
+        entry: uv lock --check
+        language: python
+        additional_dependencies: [uv]
```
viraatc: [Claude] **medium** (bug): `uv` version is unpinned in `additional_dependencies`.

```yaml
additional_dependencies: [uv]
```

Per the project's pinning policy (AGENTS.md: "pinned to an exact version"), this should pin to match the version used to generate `uv.lock`:
```yaml
additional_dependencies: ["uv==0.7.6"]
```

`uv lock --check` behavior can differ between versions, so an unpinned dependency could cause spurious pre-commit failures or false passes.

scripts/Dockerfile.dev
```diff
@@ -5,20 +5,26 @@
...
 RUN apt-get update && \
-    apt-get install -y  --no-install-recommends build-essential procps \
+    apt-get install -y --no-install-recommends build-essential procps \
     && rm -rf /var/lib/apt/lists/*
 
 RUN mkdir /mnt/inference-endpoint
 WORKDIR /mnt/inference-endpoint
-COPY pyproject.toml .
+
+# Copy lockfile + project metadata first for Docker layer caching
```
viraatc: [Claude] **low** (design): Comment says "Docker layer caching" but `COPY src/ ./src/` on the next line means any source code change invalidates the dependency install layer, defeating the caching optimization.

With `uv_build` as the build backend, `uv sync` needs the source tree present. Consider splitting into two stages with `--no-install-project` for better caching:
```dockerfile
COPY pyproject.toml uv.lock .python-version README.md ./
RUN uv sync --frozen --no-install-project --extra dev --extra test
COPY src/ ./src/
RUN uv sync --frozen --extra dev --extra test
```
This way dependency changes are cached separately from source changes.

.github/workflows/test.yml
```diff
@@ -9,26 +9,16 @@ on:
...
-      - name: Install dependencies
-        run: |
-          python -m pip install pip==26.0.1
-          pip install -e ".[test]"
+          python-version-file: .python-version
 
       - name: Run tests
-        run: |
-          pytest -xv -m "not slow and not performance" --cov=src --cov-report=xml --cov-report=html
+        run: uv run --frozen --extra test pytest -xv -m "not slow and not performance" --cov=src --cov-report=xml --cov-report=html
```
viraatc: [Claude] **low** (design): Unlike the pre-commit workflow (which has an explicit `uv sync --frozen` step), the test workflow relies on `uv run --frozen` to implicitly sync dependencies.

Adding an explicit `uv sync --frozen --extra test` step before `pytest` would make the install step visible in CI logs, making dependency-related failures easier to diagnose:
```yaml
- name: Install dependencies
  run: uv sync --frozen --extra test

- name: Run tests
  run: uv run pytest ...
```

viraatc (commented):

Reviewed by: **Codex** + **Claude** | Depth: **thorough** (3001 lines)

Found **5** issues across **4** files.

viraatc:

Reviewed by: **Codex** + **Claude** | Depth: **thorough** (3001 lines, 9 files)

Found **5** issues across **4** files.

### 🔴 Must Fix (critical)
Issues that will cause build failures or incorrect behavior in production.

| 1 | `scripts/Dockerfile.dev` | 27 | bug | Codex | Missing `COPY README.md` — `uv_build` needs it for metadata, Docker build fails |

### 🟡 Should Fix (medium)
Real issues that trigger under specific conditions or diverge from project policy.

| 2 | `pyproject.toml` | 16 | bug | Claude | `data` key may install templates to data scheme instead of package dir — verify with `uv build && unzip -l dist/*.whl` |
| 3 | `.pre-commit-config.yaml` | 70 | bug | Claude | `additional_dependencies: [uv]` unpinned — should be `["uv==0.7.6"]` per project policy |

### 🔵 Consider (low)

| 4 | `scripts/Dockerfile.dev` | 26 | design | Claude | Misleading "layer caching" comment — `COPY src/` busts the cache; consider `--no-install-project` split |
| 5 | `.github/workflows/test.yml` | 21 | design | Claude | Implicit `uv sync` via `uv run` — explicit step improves CI diagnostics |

---

**Note:** 6 issues from the Claude reviewer were dropped during verification — they referenced files not changed in this PR (schema.py, init.py, test_cli.py, test_benchmark_command.py, main.py). All remaining issues were verified against HEAD source files.

.pre-commit-config.yaml
```diff
@@ -62,3 +62,11 @@ repos:
...
         pass_filenames: true
         exclude: ^(src/inference_endpoint/openai/openai_types_gen.py)$
+
+      - id: uv-lock-check
+        name: Check uv.lock is up-to-date
+        entry: uv lock --check
+        language: python
+        additional_dependencies: ["uv==0.7.6"]
+        pass_filenames: false
+        files: ^pyproject\.toml$
```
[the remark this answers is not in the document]
viraatc: Addressed — widened the files selector to `^(pyproject\.toml|uv\.lock)$`.

pyproject.toml
```diff
@@ -1,6 +1,19 @@
 [build-system]
-requires = ["setuptools==78.1.1", "wheel==0.46.3"]
-build-backend = "setuptools.build_meta"
+requires = ["uv_build==0.7.6"]
+build-backend = "uv_build"
```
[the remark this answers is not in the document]
viraatc: Addressed — added a `build` job to the test workflow that builds the wheel, installs it in a clean venv, and runs `inference-endpoint --help` as a smoke test.

nv-alicheng (approved):

pyproject.toml
```diff
@@ -1,6 +1,19 @@
...
+environments = [
+    "sys_platform == 'linux' and platform_machine == 'x86_64'",
+    "sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "sys_platform == 'darwin' and platform_machine == 'x86_64'",
+    "sys_platform == 'darwin' and platform_machine == 'arm64'",
+]
+
+[tool.uv-build]
+module-root = "src"
+exclude = ["inference_endpoint/evaluation/livecodebench/_server.py"]
```
arekay-nv: [Review by Claude] 🔴 **Blocker — this entire block is silently ignored, and `_server.py` ships in the wheel as a result.**

Two bugs in three lines:

1. **Wrong table name.** Per [uv docs](https://docs.astral.sh/uv/concepts/build-backend/), the backend config table is `[tool.uv.build-backend]`. `[tool.uv-build]` isn't a recognized table — uv silently ignores it and falls back to defaults.
2. **Wrong key name.** The exclusion key is `source-exclude` (applies to sdist + wheel) or `wheel-exclude` (wheel only), not `exclude`.

I verified by building the wheel three ways locally (uv 0.11.6):

| Config | `_server.py` in wheel? |
|---|---|
| `[tool.uv-build]` + `exclude = [...]` (this PR) | **YES (15755 bytes)** |
| `[tool.uv.build-backend]` + `exclude = [...]` | YES (key ignored) |
| `[tool.uv.build-backend]` + `source-exclude = [...]` | NO ✅ |

**Why it matters:** `_server.py` imports `fastapi` at module top (line 29), which is not declared in `project.dependencies` and is not in `uv.lock`. On `main`, `[tool.setuptools.exclude-package-data]` intentionally kept this file out of the wheel for that reason. Any downstream user who `import inference_endpoint.evaluation.livecodebench._server` from an installed wheel hits `ModuleNotFoundError: No module named 'fastapi'`.

The smoke-test CI job (wheel build + `inference-endpoint --help`) doesn't touch this module, so CI stays green while the regression is real.

**Fix:**
```toml
[tool.uv.build-backend]
module-root = "src"
source-exclude = ["inference_endpoint/evaluation/livecodebench/_server.py"]
```

(`module-root = "src"` is the default, so it's optional — keep or drop.)

Recommend extending the smoke test to `python -c "from inference_endpoint.evaluation.livecodebench import _server"` and asserting it raises (or that the file is not present) — that would have caught this.

README.md
```diff
@@ -13,10 +13,23 @@ A high-performance benchmarking tool for LLM inference endpoints, targeting 50k+
...
 pip install .
 ```
 
+After activating the venv, commands work without the `uv run` prefix.
+
+</details>
+
 ```bash
 # Test endpoint connectivity
 inference-endpoint probe \
```
arekay-nv: [Review by Claude] 🟡 **This command will fail for new users following the primary uv path.**

Above, the quick start says `uv sync` (line 16), which creates `.venv/` but doesn't activate it in the user's shell. The commands here — `inference-endpoint probe …`, `benchmark offline …`, etc. — rely on `inference-endpoint` being on PATH. It won't be.

The `<details>` block below handles this for the pip+venv path with an explicit activation note, but the primary uv path doesn't show activation OR `uv run` prefix. Result: copy-paste these commands and hit `command not found`.

Either:
- Prefix each command with `uv run` (matches AGENTS.md and CONTRIBUTING.md style), or
- Add `source .venv/bin/activate` after `uv sync` on line 16.

Same issue on line 58 (`python -m inference_endpoint.testing.echo_server …` and the subsequent `inference-endpoint benchmark …`).

scripts/Dockerfile.dev
```diff
@@ -29,8 +34,14 @@ RUN if ! getent group ${GROUP_ID}; then \
...
 USER appuser
-ENV PATH="/home/appuser/.local/bin:$PATH"
+ENV PATH="/opt/venv/bin:/home/appuser/.local/bin:$PATH"
 
-RUN pip install -e .[dev,test]
+# Install dependencies first (cached unless pyproject.toml/uv.lock change)
+RUN uv sync --frozen --no-install-project --extra dev --extra test
+
+# Copy source and install project
+COPY src/ ./src/
```
arekay-nv: [Review by Claude] 🟡 **`COPY src/` after `USER appuser` (line 39) without `--chown` — files end up root-owned.**

`COPY` respects `USER` for the filesystem it writes into, but the created files themselves default to uid/gid 0 unless `--chown` is specified. Then line 47 runs `uv sync --frozen --extra dev --extra test` as `appuser`, which needs to write install metadata (RECORD, egg-info, `__pycache__`) under `src/`. Reads work via the default 755 on the copied directory, but anything that needs to write under `src/` as appuser fails.

In practice the intended usage (`docker run -v $(pwd):/mnt/inference-endpoint`) volume-mounts over the baked-in copy, papering over this — but the point of the two-step `uv sync` is dependency layer caching, which relies on the baked-in copy being usable on its own.

Suggested:
```dockerfile
COPY --chown=${USER_ID}:${GROUP_ID} src/ ./src/
```

.pre-commit-config.yaml
```diff
@@ -62,3 +62,11 @@ repos:
         types: [python]
         pass_filenames: true
         exclude: ^(src/inference_endpoint/openai/openai_types_gen.py)$
+
+      - id: uv-lock-check
+        name: Check uv.lock is up-to-date
+        entry: uv lock --check
+        language: python
+        additional_dependencies: ["uv==0.7.6"]
```
arekay-nv: [Review by Claude] 🟢 Nit: with `language: python` + `additional_dependencies: ["uv==0.7.6"]`, pre-commit creates a throwaway Python venv per-hook and `pip install`s uv into it just to call `uv lock --check`. Wasteful given uv is already installed in the project venv that runs pre-commit. Consider `language: system` with `entry: uv lock --check` (relies on host `uv`, which is reasonable in a uv-first project). Not a blocker.

pyproject.toml
```diff
@@ -1,6 +1,19 @@
 [build-system]
-requires = ["setuptools==78.1.1", "wheel==0.46.3"]
-build-backend = "setuptools.build_meta"
+requires = ["uv_build==0.7.6"]
+build-backend = "uv_build"
+
+[tool.uv]
+index-url = "https://pypi.org/simple"
```
arekay-nv: [Review by Claude] 🟢 Nit: PyPI (`https://pypi.org/simple`) is uv's default `index-url`, so this line is a no-op. Recommend dropping it to avoid implying configuration where there is none.

arekay-nv (commented):
[Review by Claude] Reviewed locally against `main`. Build direction (setuptools → uv_build, lockfile-based CI, pinned uv) is sound, but there's one blocker: the `[tool.uv-build]` table is silently ignored because the correct name is `[tool.uv.build-backend]`, AND `exclude` should be `source-exclude`. Net effect: `_server.py` (which imports `fastapi`, not a project dep) ships in the wheel where main intentionally excluded it. Verified by building both ways.

| Severity | Issue |
|---|---|
| 🔴 | `[tool.uv-build]` + `exclude` silently ignored → `_server.py` ships with unresolvable `fastapi` import |
| 🟡 | README primary (uv) path shows bare `inference-endpoint` after `uv sync` with no activation or `uv run` |
| 🟡 | Dockerfile.dev: `COPY src/` after `USER appuser` without `--chown` |
| 🟢 | Pre-commit `uv-lock-check` hook uses `language: python` to install uv via pip — wasteful |
| 🟢 | Redundant `index-url = "https://pypi.org/simple"` (PyPI is the default) |
| 🟢 | 23 commits, many `fix:`-the-previous-`fix:` — consider squashing before merge |

Details on the inline comments below.

arekay-nv (approved):
Awesome, thanks!

PR #277, opened 2026-04-10
chore: drop metrics from schema

viraatc:
## What does this PR do?

based on offline discussion, we want to drop metrics from schema (it was unused anyways).

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactor/cleanup

## Testing

- [x] Tests added/updated
- [x] All tests pass locally

## Checklist

- [x] Pre-commit hooks pass

### DIFF
--- a/examples/02_ServerBenchmarking/offline_llama3_8b_cnn.yaml
+++ b/examples/02_ServerBenchmarking/offline_llama3_8b_cnn.yaml
@@ -30,13 +30,6 @@ settings:
   client:
     num_workers: 4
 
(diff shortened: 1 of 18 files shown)


arekay-nv (approved):
Thanks!
Can you also check if the templates were generated or edited - i don't see the generation script updated, so it might be broken if the templates weren't generated.

viraatc:
confirmed - templates checked in by precommit, which was triggered by change to schema.py already.

PR #281, opened 2026-04-14
Add batched publishing to pubsub

nv-alicheng:
## What does this PR do?

Adds a 'send_threshold' to publish EventRecords as a batch to reduce the number of syscalls being invoked by ZMQ.

A large bottleneck was found on ARM when testing via the max throughput test server - msync cannot keep up with the batches. As such, the `msync` call has been *removed* for now:

Implications:
 - Live metrics cannot be added before KVStore / MetricsAggregator is redesigned (i.e. AIPerf style or Prometheus metrics)
 - Higher max throughput: 30k -> 50k.
 - Batching becomes a new 'subscription' topic, so subscribers must subscribe to it - since Publisher does not know what the subscribers are subscribing to, unnecessary records for other topics may appear in the batch. Currently, both major subscriber service subscribe to all topics, so this is not an issue.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [x] Tests added/updated
- [x] All tests pass locally

## Checklist

- [x] Pre-commit hooks pass


src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -533,8 +533,14 @@ async def _run_benchmark_async(
...
                 logger.warning(f"Client cleanup error: {e}")
+            logger.info(
+                "Closing publisher (buffer=%d, pending=%d)...",
+                len(publisher._batch_buffer),
+                len(publisher._pending),
+            )
             publisher.close()
-            await asyncio.to_thread(launcher.wait_for_exit, 10.0)
+            logger.info("Waiting for services to finish processing...")
+            await asyncio.to_thread(launcher.wait_for_exit, None)
```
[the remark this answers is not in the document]
nv-alicheng: This was intentional since on ARM, the MetricsAggregator service can take a while to catch up. After removing the msync call, it no longer seems to be the case. Maybe we can add in a 60s timeout?

src/inference_endpoint/async_utils/transport/zmq/pubsub.py
```diff
@@ -185,15 +231,15 @@ def __init__(
...
         self._socket.setsockopt(zmq.RCVHWM, 0)
 
-        # Subscribe to topics
         if not topics:
             self._socket.setsockopt(zmq.SUBSCRIBE, b"")
         else:
             for topic in topics:
                 self._socket.setsockopt(zmq.SUBSCRIBE, topic.encode("utf-8"))
+            # Always subscribe to batch topic so batched messages are received
+            self._socket.setsockopt(zmq.SUBSCRIBE, BATCH_TOPIC)
```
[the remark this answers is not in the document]
nv-alicheng: This is intentional and is documented as a limitation as a performance tradeoff.

src/inference_endpoint/core/record.py
```diff
@@ -32,6 +32,9 @@
 equal to the length of the longest topic string.
 """
 
+BATCH_TOPIC: Final[bytes] = b"batch".ljust(TOPIC_FRAME_SIZE, b"\0")
```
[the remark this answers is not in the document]
nv-alicheng: This seems like a hallucination to me. There is no such 'batch' topic, this is intentional.

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+
+### Load Strategies
+
+Three load patterns, three implementations — each uses the optimal async primitive
+for its scheduling semantics, validated by benchmarking:
+
+| LoadPatternType   | Strategy              | Mechanism                    | Best At             |
+| ----------------- | --------------------- | ---------------------------- | ------------------- |
+| POISSON           | `TimedIssueStrategy`  | `loop.call_at` (default)     | ≤50k QPS            |
+| POISSON (precise) | `TimedIssueStrategy`  | `run_in_executor(busy_wait)` | Sub-100μs precision |
```
nvzhihanj, line 77: Is this mechanism auto selected, or do people need to explicitly tune them?

nvzhihanj: NVM I see it below

review-council-output.md
```diff
@@ -1,104 +1,65 @@
-# Review Council — PR "Make Loadgen Async" (Round 4)
+# Review Council — Batched ZMQ Publisher
```
nvzhihanj: What's this file? Doesn't seem it should be part of commit

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+
+        case LoadPatternType.CONCURRENCY:
+            return ConcurrencyStrategy(lp.target_concurrency, sample_order)
+```
+
+---
+
+## Benchmark Data Summary
+
+From `.cursor_artifacts/async_lg_benchmarks/` (MaxThroughputServer + real HTTPEndpointClient):
```
nvzhihanj: Likely obsolete

src/inference_endpoint/async_utils/transport/zmq/pubsub.py
```diff
@@ -56,85 +80,96 @@ def __init__(
...
 
-        # Attempt direct send:
-        if not self._buffer:
+        Uses direct per-record send when only 1 record is buffered
+        (avoids batch encoding overhead for single records like ENDED).
+        """
+        if self._batch_buffer:
+            self._flush_batch()
+
+    def _flush_batch(self) -> None:
```
nvzhihanj, line 123: [Claude] **high** (data-integrity): **`_flush_batch()` clears buffer before send — data lost on ZMQ error.** The buffer is swapped out before `_send_frame()` runs:

```python
buf = self._batch_buffer
self._batch_buffer = []          # cleared here
# ...
self._send_frame(...)            # if this raises, buf is gone
```

`zmq.Again` is caught in `_drain_pending`, but `zmq.ZMQError` (e.g., `ETERM` when context is terminated) can escape the handler. Since `buf` goes out of scope, the records are unrecoverable. Consider re-queuing `buf` into `_batch_buffer` on exception, or only clearing after successful send.

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+    runtime_settings: RuntimeSettings
+    dataset: Dataset
+    phase_type: PhaseType = PhaseType.PERFORMANCE
 
-`load_generator/` is the central orchestrator for a benchmark run. It controls **what** to send
-(dataset samples), **when** to send them (load pattern), and **how** to observe the results
-(event hooks feeding the metrics recorder).
 
-## Responsibilities
+class BenchmarkSession:
```
nvzhihanj, line 140: [Claude] **high** (bug): **`BenchmarkSession.__init__` signature is wrong in 3 ways vs actual code (`session.py:224`):**

```python
# Doc (here):
def __init__(self, issuer, event_publisher: EventRecordPublisher, on_sample_complete: Callable[[QueryResult], None] | None = None)

# Actual:
def __init__(self, issuer, event_publisher: EventPublisher, loop: asyncio.AbstractEventLoop,
             on_sample_complete: Callable[[QueryResult | StreamChunk], None] | None = None,
             session_id: str | None = None)
```

1. Missing required `loop` param — callers following this doc get `TypeError`
2. `EventRecordPublisher` → actual is `EventPublisher` (different protocol)
3. `Callable[[QueryResult], None]` → actual is `Callable[[QueryResult | StreamChunk], None]`

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+
+- `WithoutReplacementSampleOrder` — shuffle, exhaust, reshuffle
+- `WithReplacementSampleOrder` — uniform random
+
+Termination is controlled by `BenchmarkSession._should_stop()`, not the iterator.
+
+### SessionResult
+
+```python
+@dataclass(frozen=True, slots=True)
```
nvzhihanj: [Claude] **high** (bug): **`PhaseResult` doc doesn't match `session.py:76`:**

1. Doc: `@dataclass(frozen=True, slots=True)` → Actual: `@dataclass(frozen=True)` (no `slots=True`)
2. Doc includes `report: Report | None` — **this field does not exist** in the actual struct
3. Doc omits `issued_count: int` which IS in the actual struct

The phantom `report` field makes the code example at line 1069 (`pr.report.qps()`) actively misleading — this would raise `AttributeError` at runtime.

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+    def on_query_complete(self, query_id: str) -> None:
+        self._sem.release()
+```
+
+### SampleIssuer (Protocol)
+
+```python
+class SampleIssuer(Protocol):
+    def issue(self, query: Query) -> None: ...
+    def poll(self) -> QueryResult | StreamChunk | None: ...
```
nvzhihanj: [Claude] **medium** (bug): **`SampleIssuer` doc includes `poll()` — not in actual code.** The actual `SampleIssuer` Protocol (`session.py:110-119`) has:

```python
class SampleIssuer(Protocol):
    def issue(self, query: Query) -> None: ...
    async def recv(self) -> QueryResult | StreamChunk | None: ...
    def shutdown(self) -> None: ...
```

No `poll()` method. The doc below says "Uses `recv()` exclusively — no `poll()` spin" (line 260) which contradicts having `poll()` in the protocol.

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+    ctx: BenchmarkContext,
+    loop: asyncio.AbstractEventLoop,
+) -> tuple[SessionResult, ResponseCollector]:
+    collector = ResponseCollector(collect_responses=ctx.collect_responses)
+
+    # Setup: HTTP client, event publisher, service subprocesses
+    client = await HTTPEndpointClient.create(ctx.http_config, loop)
+    issuer = HttpClientSampleIssuer(client)
+    zmq_ctx = ManagedZMQContext()
+    publisher = ZmqEventRecordPublisher(pub_socket_name, zmq_ctx, loop=loop)
```
nvzhihanj: [Claude] **medium** (bug): **CLI Integration code is stale — contradicts this PR's own changes.** The snippet shows:

```python
publisher = ZmqEventRecordPublisher(pub_socket_name, zmq_ctx, loop=loop)
```

But `execute.py:417` (changed in this PR) now uses:

```python
publisher = EventPublisherService(zmq_ctx)
```

Also, `run_benchmark_async` is documented as returning `tuple[SessionResult, ResponseCollector]` but the actual return type is `BenchmarkResult` (a dataclass wrapping session, collector, report).

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+`ResponseCollector.on_complete` drives the progress bar:
+
+```python
+class ResponseCollector:
+    def __init__(self, collect_responses: bool, pbar: tqdm | None = None):
+        self.responses: dict[str, str] = {}
+        self.errors: list[str] = []
+        self.pbar = pbar
+
+    def on_complete(self, result: QueryResult) -> None:
```
nvzhihanj: [Claude] **medium** (bug): **`ResponseCollector` snippet has 3 errors vs `execute.py:99-119`:**

1. Method name: `on_complete` → actual: `on_complete_hook`
2. Missing `count` attribute (actual has `self.count = 0`, incremented each call)
3. Condition `elif self.responses is not None:` → actual: `elif self.collect_responses:` (`responses` is always a dict, never None)

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+    """Result of a single benchmark phase."""
+    name: str
+    phase_type: PhaseType
+    uuid_to_index: dict[str, int]
+    report: Report | None  # Only for PERFORMANCE phases
+    start_time_ns: int
+    end_time_ns: int
+
+
+@dataclass(frozen=True, slots=True)
```
nvzhihanj: [Claude] **low** (bug): Doc: `@dataclass(frozen=True, slots=True)` → Actual (`session.py:87`): `@dataclass(frozen=True)`. Same discrepancy as `PhaseResult` above.

nvzhihanj (commented):

Reviewed by: **Claude** (Codex unavailable) | Depth: **thorough**

Found **11** issues across **4** files.

*10 existing inline comments were already present. Overlapping issues excluded.*

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+The strategy calls `phase_issuer.issue(idx)`. After the phase completes,
+the session reads `phase_issuer.uuid_to_index` and `phase_issuer.issued_count`
+to build the `PhaseResult`.
+
+**UUID generation before Query construction** avoids the old `Sample` catch-22.
+`Query` is a frozen `msgspec.Struct` — all fields set at construction, no mutation.
+
+**`_receive_responses()`** — concurrent coroutine, purely async:
+
+```python
```
nvzhihanj, line 265: [Claude] **low** (bug): **Pseudocode doesn't match `session.py:366-381`:**

- Doc: `while True:` with `if self._done and self._inflight <= 0: break`
- Actual: `while not self._done:` — no inflight check; task is cancelled externally by `run()`
- Doc: `self.issuer` → actual: `self._issuer`

The shutdown contract is fundamentally different: the doc implies self-termination on drain, but the actual task is externally cancelled.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -533,8 +533,14 @@ async def _run_benchmark_async(
                     await http_client.shutdown_async()
             except Exception as e:
                 logger.warning(f"Client cleanup error: {e}")
+            logger.info(
+                "Closing publisher (buffer=%d, pending=%d)...",
```
nvzhihanj, line 537: [Claude] **low** (design): **Accessing private attributes of the publisher in a log statement:**

```python
len(publisher._batch_buffer),   # private attr
len(publisher._pending),        # private attr
```

These are internal to `ZmqEventRecordPublisher` and not part of the `EventPublisherService` public interface. If renamed, this silently raises `AttributeError` at runtime. Consider adding a `publisher.stats` or `publisher.buffered_count` property.

src/inference_endpoint/async_utils/transport/protocol.py
```diff
@@ -278,9 +278,17 @@ def send(self, topic: bytes, payload: bytes) -> None:
         """
         raise NotImplementedError("Subclasses must implement this method.")
 
+    @abstractmethod
+    def flush(self) -> None:
```
nvzhihanj: [Claude] **low** (api-contract): **Inconsistent abstract method pattern.** `send()` (line 279) and `close()` (line 292) use `raise NotImplementedError(...)`, but `flush()` uses `pass`. The docstring says "No-op for unbuffered implementations" — if that's the intent, consider making it a non-abstract method with `pass` body, so unbuffered subclasses don't need to override it.

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+
+- `QueryResult`: publish COMPLETE event, decrement `_inflight`, call `on_sample_complete`,
+  call `strategy.on_query_complete(query_id)` if strategy supports it
+- `StreamChunk(first)`: publish RECV_FIRST event
+- `StreamChunk(non-first)`: publish RECV_NON_FIRST event
+
+**Timestamp fidelity:**
 
-### `SampleIssuer` (abstract base class — implemented externally)
+- ISSUED: `monotonic_ns()` taken immediately before `issuer.issue()`. The ZMQ push is
```
nvzhihanj, line 298: [Claude] **low** (bug): **Timestamp fidelity claim is stale under new batching.** The doc says "The ZMQ push is sync and non-blocking" — but with batched publishing, `publish()` now just buffers the payload. The actual ZMQ send is deferred until threshold or explicit `flush()`. The ISSUED timestamp may not correlate with when the record actually enters the transport.

nvzhihanj:

Reviewed by: **Claude** (Codex unavailable) | Depth: **thorough**

Found **11** issues across **4** files. All line numbers verified against source files at HEAD.

*10 existing inline comments from prior reviews were already present. Overlapping issues excluded.*

### 🔴 Must Fix (high) — 3 issues
Issues that will actively mislead readers or risk data loss.

| 1 | `pubsub.py` | 113 | data-integrity | `_flush_batch()` clears buffer before `_send_frame()` — if ZMQ raises (e.g., `ETERM`), buffered records are permanently lost |
| 2 | `DESIGN.md` | 127 | doc-accuracy | `BenchmarkSession.__init__` missing required `loop` param, wrong `EventRecordPublisher` type (actual: `EventPublisher`), wrong callback type |
| 3 | `DESIGN.md` | 460 | doc-accuracy | `PhaseResult` has phantom `report: Report \| None` field that doesn't exist, missing `issued_count`, wrong decorator — code example at line 1069 would raise `AttributeError` |

### 🟡 Should Fix (medium) — 3 issues
Real inaccuracies or design issues.

| 4 | `DESIGN.md` | 440 | doc-accuracy | `SampleIssuer` protocol includes `poll()` method that doesn't exist in actual code |
| 5 | `DESIGN.md` | 925 | doc-accuracy | CLI Integration snippet uses old `ZmqEventRecordPublisher` — this PR changed it to `EventPublisherService`; return type also wrong |
| 6 | `DESIGN.md` | 982 | doc-accuracy | `ResponseCollector.on_complete` → actual: `on_complete_hook`; missing `count` attr; wrong guard condition |

### 🔵 Consider (low) — 5 issues
Valid improvements, could be follow-ups.

| 7 | `DESIGN.md` | 471 | doc-accuracy | `SessionResult` has `slots=True` in doc but not in code |
| 8 | `DESIGN.md` | 250 | doc-accuracy | `_receive_responses` pseudocode differs from actual — wrong loop condition and shutdown model |
| 9 | `execute.py` | 537 | design | Direct access to private `publisher._batch_buffer` and `._pending` attrs |
| 10 | `protocol.py` | 282 | api-contract | `flush()` uses `pass` body while `send()`/`close()` use `raise NotImplementedError` |
| 11 | `DESIGN.md` | 280 | doc-accuracy | ISSUED timestamp claim stale under batching — `publish()` now buffers, send is deferred |

### Testing gaps (not posted inline — lines outside diff)
- No test for threshold-triggered auto-flush (threshold=1000 path in `pubsub.py:101` is untested)
- No test for `ZmqEventRecordSubscriber._buffer` deque drain — subsequent `receive()` calls after batch decode are untested

---
*🤖 Generated with [Claude Code](https://claude.com/claude-code) — Review Council*

docs/load_generator/DESIGN.md
nvzhihanj: Is this a rebase issue?

docs/load_generator/DESIGN.md
nvzhihanj: Valid, seems like the doc change is lagging behind

docs/load_generator/DESIGN.md
nvzhihanj: Seems valid: https://github.com/mlcommons/endpoints/blob/08d10e8a99be6b20b55fb4a0a5aee127829cf4e1/src/inference_endpoint/load_generator/session.py#L76

I assume the all doc changes need to be reworked on

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+concurrency levels, datasets) against the same server instance within a single session,
+each producing an independent report.
+
+```
+BenchmarkSession.run(phases)
+    |
+    +-- STARTED
+    +-- [saturation]     strategy.execute() → NO drain (keep in-flight saturated)
+    +-- [perf phase 1]   START_PERFORMANCE_TRACKING → strategy.execute() → drain → STOP_PERFORMANCE_TRACKING → snapshot report
+    +-- [saturation]     strategy.execute() → drain
```
arekay-nv: No drain for the second perf phase?

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+**File:** `src/inference_endpoint/load_generator/session.py`
+
+Async orchestrator. Runs phases sequentially on the shared event loop.
+
+```python
+class PhaseType(str, Enum):
+    """Phase types control tracking and reporting behavior."""
+    PERFORMANCE = "performance"  # Tracked, produces a report
+    ACCURACY = "accuracy"        # Untracked, for eval scoring
+    SATURATION = "saturation"    # Untracked, ramp up concurrency before perf phase
```
arekay-nv: Should we just rename this to `WARMUP` to avoid any confusion? Seems like there will be a lot of questions around difference between warmup and saturation so why not use the established term?

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
-- Coordinate graceful shutdown with in-flight drain
+    async def run(self, phases: list[PhaseConfig]) -> SessionResult: ...
+    def stop(self) -> None: ...
+```
+
+**`run(phases)`** lifecycle:
 
-## Component Map
+1. Publish `SessionEventType.STARTED`
+2. Start receiver coroutine (`_receive_responses`)
```
arekay-nv, line 157: Does this constrain us to have a single response handler for all phases? One option can be to create a wrapper for each phase - that way, the response handler, dataset, runtime setting, sample order and load strategy can be grouped together.

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+- `WithoutReplacementSampleOrder` — shuffle, exhaust, reshuffle
+- `WithReplacementSampleOrder` — uniform random
+
+Termination is controlled by `BenchmarkSession._should_stop()`, not the iterator.
+
+### SessionResult
+
+```python
+@dataclass(frozen=True, slots=True)
+class PhaseResult:
```
arekay-nv, line 482: Should this include the dataset as well - might be a good idea if we want to add more details such as sha of the dataset for checking.

src/inference_endpoint/async_utils/event_publisher.py
```diff
@@ -19,42 +19,41 @@
...
-                will block until the message has been successfully sent. In most cases, this should not
-                be turned on, but it is useful for testing, or specifically in the use case where
-                EventRecords are being used as a synchronization mechanism (i.e. sending a specific
-                EventRecord as a STOP signal to subscribers to ensure the ordering of cleanup.)
-            isolated_event_loop (bool): If True, the publisher will be run in a separate event loop.
+            managed_zmq_context: The managed ZMQ context to use.
+            extra_eager: If True, publish() blocks until the message is sent.
+                Useful for testing or when EventRecords are used as a
+                synchronization mechanism (e.g., ENDED as a stop signal).
+            isolated_event_loop: If True, runs on a separate event loop thread.
```
arekay-nv, line 45: missing doc for `send_threshold`

review-council-output.md
arekay-nv, line 1: stray file?

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+
+    S->>B: issue_fn(sample_index)
+    B->>D: load_sample(index)
+    D-->>B: sample_data
+    Note over B: Build Query(id=uuid4().hex, data=load_sample(idx))
+    B->>E: publish(ISSUED, uuid, timestamp_ns)
+    E->>M: ZMQ PUB (EventRecord)
+    B->>I: issue(query)
+    I->>W: ZMQ PUSH (Query)
+    W->>W: HTTP request → endpoint
```
arekay-nv, line 533: Missing the non-first chunk response event. Might be tricky to add, but can add a note that it is missing from the diagram.

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+| `Sample` class                                  | Replaced by `Query` (frozen `msgspec.Struct`) |
+| `Sample.__setattr__` hack                       | UUID generated before `Query` construction    |
+| `SampleEventHandler` singleton                  | Events via `EventPublisher` ZMQ PUB/SUB       |
+| `IssuedSample` dataclass                        | `uuid_to_index` dict on session is sufficient |
+| `Scheduler` class hierarchy                     | Replaced by `LoadStrategy` + factory function |
+| `LoadGenerator` / `SchedulerBasedLoadGenerator` | Replaced by `LoadStrategy`                    |
+| `threading.Thread` in `BenchmarkSession`        | Fully async                                   |
+| `threading.Condition` in `ConcurrencyScheduler` | `asyncio.Semaphore`                           |
+| `HttpClientSampleIssuer._handle_responses`      | Session owns the receive loop                 |
+
```
arekay-nv: Do we need this?

docs/load_generator/DESIGN.md
```diff
@@ -1,152 +1,1086 @@
...
+```
+src/inference_endpoint/load_generator/
+├── __init__.py          # Public exports
+├── session.py           # BenchmarkSession, SessionResult
+├── strategy.py          # LoadStrategy protocol, TimedIssueStrategy,
+│                        #   BurstStrategy, ConcurrencyStrategy,
+│                        #   create_load_strategy()
+├── sample_order.py      # SampleOrder, WithoutReplacement, WithReplacement
+└── delay.py             # poisson_delay_fn, uniform_delay_fn
+```
```
arekay-nv: Move to top?

arekay-nv (approved):
Thanks. Minor cleanup/clarifications. Please update with the perf table.

viraatc (approved):

review-council-output.md
nv-alicheng: Fixed — file removed.

docs/load_generator/DESIGN.md
nv-alicheng: Fixed — reference removed from DESIGN.md.

src/inference_endpoint/async_utils/transport/zmq/pubsub.py
nv-alicheng: Fixed — _flush_batch() now restores buffer on ZMQ error via try/except.

docs/load_generator/DESIGN.md
nv-alicheng: Fixed — poll() removed from SampleIssuer in DESIGN.md.

docs/load_generator/DESIGN.md
nv-alicheng: Fixed — CLI Integration section rewritten to match actual execute.py.

docs/load_generator/DESIGN.md
nv-alicheng: Fixed — SessionResult slots=True removed to match actual code.

docs/load_generator/DESIGN.md
nv-alicheng: Fixed — pseudocode updated to match actual implementation.

src/inference_endpoint/commands/benchmark/execute.py
nv-alicheng: Fixed — replaced with public buffered_count/pending_count properties.

src/inference_endpoint/async_utils/transport/protocol.py
nv-alicheng: Fixed — flush() is now non-abstract with noqa: B027.

docs/load_generator/DESIGN.md
nv-alicheng: Fixed — timestamp fidelity section updated for batched publishing.

src/inference_endpoint/async_utils/event_publisher.py
nv-alicheng: Fixed — send_threshold documented in docstring.

review-council-output.md
nv-alicheng: Fixed — file removed.

docs/load_generator/DESIGN.md
nv-alicheng: Fixed — signature corrected (added loop, session_id, fixed types).

docs/load_generator/DESIGN.md
nv-alicheng: Fixed — PhaseResult corrected (issued_count, no slots=True).

docs/load_generator/DESIGN.md
nv-alicheng: Fixed — ResponseCollector snippet corrected (on_complete_hook, count attr, conditions).

src/inference_endpoint/async_utils/transport/zmq/pubsub.py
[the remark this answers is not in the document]
nv-alicheng: Known limitation, documented in ZmqEventRecordSubscriber docstring. All current subscribers use topics=None. Per-record filtering after decode is ~0.6us/record — negligible compared to processing cost.

src/inference_endpoint/commands/benchmark/execute.py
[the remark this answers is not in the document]
nv-alicheng: Changed from 10s to None because the aggregator was being killed mid-processing at 1M records (took >10s to consume the ZMQ backlog). The aggregator self-terminates on ENDED, so indefinite wait is safe. A hung service would be caught by SIGINT (Ctrl+C).

docs/load_generator/DESIGN.md
nv-alicheng: renamed to Warmup

docs/load_generator/DESIGN.md
nv-alicheng: fixed diagram, there is a drain.

docs/load_generator/DESIGN.md
nv-alicheng: Deferring to follow-up, I agree, should split up results per phase.

docs/load_generator/DESIGN.md
nv-alicheng: Same as above, deferring to followup.

PR #282, opened 2026-04-14
Major refactor to move multi-threaded load generator to async event loops

nv-alicheng:
## What does this PR do?

Major refactor which moves the multi-threaded system to an event loop based one to reduced Python context switching for higher throughput and performance.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [x] Refactor/cleanup

## Related issues

This merges from a side branch which consists of multiple PRs:
- #214 
- #215 
- #216 
- #255 
- #281 

## Testing

- [x] Tests added/updated
- [x] All tests pass locally

## Checklist

- [x] Pre-commit hooks pass


configs/gptoss_test.yaml
arekay-nv, line 1: If this is intended for tests, maybe moving under tests would be better. A `configs` folder would be used for configs needed for the client execution.

tests/unit/async_utils/services/metrics_aggregator/test_kv_store.py
```diff
@@ -0,0 +1,368 @@
...
+        assert stats.max_val == 20.0
+
+    def test_sum_sq(self):
+        stats = SeriesStats([3.0, 4.0])
+        assert stats.sum_sq == pytest.approx(3.0**2 + 4.0**2)
+
+    def test_empty(self):
+        stats = SeriesStats()
+        assert stats.count == 0
+        assert stats.total == 0.0
```
arekay-nv, line 51: what are the expected values for min_val/max_val?

tests/unit/async_utils/services/metrics_aggregator/test_kv_store.py
```diff
@@ -0,0 +1,368 @@
...
+    def test_snapshot(self, tmp_path: Path):
+        store = BasicKVStore(tmp_path / "kv")
+        store.create_key("n_issued", "counter")
+        store.create_key("latency", "series")
+        store.update("n_issued", 42)
+        store.update("latency", 150)
+        store.update("latency", 250)
+
+        snap = store.snapshot()
+        assert snap["n_issued"] == 42
```
arekay-nv, line 122: Can you also add a test where the store is updated after the snapshot and assert that it doesn't change the snapshot.

arekay-nv (commented):
Made a first pass. Will finish up tomorrow.

docs/async_utils/transport/zmq/ready_check_design.md
```diff
@@ -0,0 +1,114 @@
...
+A generic PUSH/PULL readiness protocol that works for any subprocess type:
+
+```
+Main Process                         Subprocess (worker or service)
+┌───────────────────┐               ┌───────────────────────────┐
+│ ReadyCheckReceiver│               │                           │
+│   (PULL, bind)    │               │  1. Initialize transports │
+│                   │               │  2. Subscribe / connect   │
+│   await wait(N)   │◄─── READY ───│  3. send_ready_signal()   │
+│   blocks until N  │    (PUSH)     │     (ctx, path, id)       │
```
arekay-nv, line 22: Does `await(N)` here mean `blocks until N`  or is it in addition to blocking until N signals have arrived. If different, what does `await(N)` mean?

docs/metrics/report_design.md
```diff
@@ -0,0 +1,93 @@
...
+        │
+        ▼
+     Report
+        ├── .display(fn)     → human-readable output
+        ├── .to_json(path)   → JSON serialization
+        ├── .qps             → computed from n_completed / duration
+        └── .tps             → computed from osl total / duration
+```
+
+## Design Principles
```
arekay-nv, line 30: I will try to check the code, but are the metrics reproducible from the events log? Since we are storing the raw counter/series in the kv-store, we would like to trace where a certain entry came from. For instance if we are seeing ttft spiking, we should be able to go back to the events log and build the same distribution. I am pretty sure that is the case, but please verify and add to the design principles - seems important enough to call out.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -57,70 +85,119 @@ class MetricsAggregatorService(ZmqEventRecordSubscriber):
...
+        store = self._kv_store
+
+        self._total_processed += len(records)
+        if self._total_processed - self._last_log_count >= 10000:
+            logger.debug(
+                "Aggregator processed %d records (%d in this batch)",
+                self._total_processed,
+                len(records),
+            )
+            self._last_log_count = self._total_processed
```
arekay-nv, line 158: Is this just for debugging or does it serve any other purpose?

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -57,70 +85,119 @@ class MetricsAggregatorService(ZmqEventRecordSubscriber):
...
                 else:
+                    if ev == SessionEventType.STARTED:
+                        self._session_start_ns = record.timestamp_ns
                     table.handle_session_event(record)
+                    if ev == SessionEventType.STOP_PERFORMANCE_TRACKING:
+                        store.update(
+                            MetricCounterKey.TRACKED_DURATION_NS.value,
+                            table.total_tracked_duration_ns,
+                        )
+                logger.debug("Session event: %s", ev)
```
arekay-nv, line 191: Can we just flatten this to else-if with a duplicated `table.handle_session_event(record)` call - seems cleaner that way.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -135,30 +212,58 @@ async def process(self, records: list[EventRecord]) -> None:
...
-                table.set_field(uuid, "complete_ns", ts, record)
+                # Check if tracked before set_field (which removes the row)
+                is_tracked = table.get_row(uuid) is not None
+                table.set_field(uuid, SampleField.COMPLETE_NS, ts, record)
+                self._total_completed += 1
+                store.update(
+                    MetricCounterKey.TOTAL_SAMPLES_COMPLETED.value,
+                    self._total_completed,
+                )
+                if is_tracked:
```
arekay-nv, line 240: Should we differentiate between tracked and untracked - maybe we track all samples (even warmup) since it would simplify downstream handling. Everything is tracked. I feel like the conditionals might not be worth it - plus we get to exercise the end to end warmup path. Unless there is a non-negligible cost to tracking the warmup samples.

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -112,170 +149,173 @@ def fire(
...
+    Subclass only needs to set metric_name and the required field name.
+    Skips silently if the required field is None (event hasn't occurred yet).
+    """
 
-    def __init__(self, emitter: MetricEmitter):
-        super().__init__("ttft_ns", requires=("issued_ns",))
-        self._emitter = emitter
+    def __init__(self, metric_name: str, kv_store: KVStore, subtract_field: str):
+        super().__init__(metric_name, kv_store, requires=(subtract_field,))
+        self._subtract_field = subtract_field
```
arekay-nv: Can we rename this to `internval_start` or something more descriptive - `subtract_field` captures the mathematical aspect but misses the usage context (i.e time delta).

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -112,170 +149,173 @@ def fire(
...
-        self._pool = tokenize_pool
-        self._loop = loop
+        super().__init__(MetricSeriesKey.ISL, kv_store, tokenize_pool, loop)
 
     def fire(self, ev_rec, row, pre_change):
-        if not isinstance(ev_rec.data, PromptData):
-            return None
-        if ev_rec.data.token_ids is not None:
-            self._emitter.emit(row.sample_uuid, "isl", len(ev_rec.data.token_ids))
+        # Sync fast path: pre-tokenized IDs (SGLang)
```
arekay-nv: Update the comment to indicate that any response where `token_ids` are included will pass through the fast-path - it _happens_ to be SGLang, but we could probably path vllm/trtllm as well.

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -112,170 +149,173 @@ def fire(
...
-
-            async def _compute() -> None:
-                try:
-                    count = await pool.token_count_async(text, loop)
-                    emitter.emit(uuid, "isl", count)
-                except Exception:
-                    logger.exception("ISL tokenization failed for %s", uuid)
-
-            return loop.create_task(_compute())
+        # Async path: tokenize raw text (OpenAI) — handled by base class
```
arekay-nv: Same as above - OpenAI _happens_ to go through this path.

tests/unit/transport/test_zmq_pool_transport.py
```diff
@@ -0,0 +1,180 @@
...
+        loop = asyncio.get_running_loop()
+        await self._create_publisher_and_pool(loop, 3)
+
+    async def test_4_workers(self):
+        loop = asyncio.get_running_loop()
+        await self._create_publisher_and_pool(loop, 4)
+
+    async def test_8_workers(self):
+        loop = asyncio.get_running_loop()
+        await self._create_publisher_and_pool(loop, 8)
```
arekay-nv: Parameterize similar to the `test_pool_only` below.

tests/unit/transport/test_zmq_pool_transport.py
```diff
@@ -0,0 +1,180 @@
...
+
+
```
arekay-nv: This can be consolidated with `TestZmqPoolTransportWithoutPublisher` via a bool `create_publisher` parameter.

tests/unit/metrics/test_report_builder.py
```diff
@@ -0,0 +1,262 @@
...
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.unit
+class TestComputeSummary:
+    def test_empty(self):
+        s = compute_summary(SeriesStats())
+        assert s["total"] == 0
+        assert s["min"] == 0
+        assert s["histogram"]["buckets"] == []
```
arekay-nv, line 42: Please add `max` and `std_dev` behavior.

tests/unit/load_generator/test_sample_order.py
```diff
@@ -0,0 +1,88 @@
...
+        assert sorted(indices) == [0, 1, 2, 3, 4]
+
+    def test_reshuffles_after_exhaustion(self):
+        order = WithoutReplacementSampleOrder(
+            n_samples_in_dataset=3, rng=random.Random(42)
+        )
+        first_pass = [next(order) for _ in range(3)]
+        second_pass = [next(order) for _ in range(3)]
+        assert sorted(first_pass) == [0, 1, 2]
+        assert sorted(second_pass) == [0, 1, 2]
```
arekay-nv: ideally parameterize over values - small, medium, large .

tests/unit/load_generator/test_sample_order.py
```diff
@@ -0,0 +1,88 @@
...
+        assert len(indices) == 100
+        assert all(0 <= i < 2 for i in indices)
+
+    def test_reproducible_with_seed(self):
+        order1 = WithoutReplacementSampleOrder(
+            n_samples_in_dataset=10, rng=random.Random(42)
+        )
+        order2 = WithoutReplacementSampleOrder(
+            n_samples_in_dataset=10, rng=random.Random(42)
+        )
```
arekay-nv, line 69: Nice, but can we also parameterize over different seed values - `42` is a great seed - but sometimes we might want to use a different seed :)

tests/unit/async_utils/services/metrics_aggregator/test_aggregator.py
```diff
@@ -114,95 +50,95 @@ def _streaming_text(*chunks: str) -> TextModelOutput:
...
-                _session(SessionEventType.STARTED, ts=0),
-                _sample(SampleEventType.ISSUED, "s1", ts=100),
+                session_event(SessionEventType.STARTED, ts=0),
+                sample_event(SampleEventType.ISSUED, "s1", ts=100),
             ]
         )
         assert agg._table.get_row("s1") is None
-        assert emitter.emitted == []
+        assert store.get_series_values("ttft_ns") == []
+        assert store.get_series_values("sample_latency_ns") == []
```
arekay-nv: Please add a comment explaining the assertion failure - `"START_PERFORAMCE_TRACKING" not issued - unexpected samples` or something of the sort. This will help debug the failures quickly.

tests/unit/async_utils/services/metrics_aggregator/conftest.py
```diff
@@ -0,0 +1,198 @@
...
+    def __init__(self, delay: float = 0.01) -> None:
+        self._delay = delay
+
+    def token_count(self, text: str) -> int:
+        return len(text.split())
+
+    async def token_count_async(
+        self, text: str, _loop: asyncio.AbstractEventLoop
+    ) -> int:
+        await asyncio.sleep(self._delay)
```
arekay-nv, line 119: suggestion: A slightly more realistic fixture would use the delay per token - so sleep over number of tokens times delay.

arekay-nv (commented):
Part 2 of reviews - next one should wrap it up

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -146,17 +172,37 @@ def enable_streaming(self) -> bool:
...
 
 
-def _load_tokenizer(model_name: str) -> AutoTokenizer | None:
-    """Load HuggingFace tokenizer, warn on failure."""
+def _check_tokenizer_exists(model_name: str) -> bool:
+    """Check if a HuggingFace tokenizer exists for the model (API only, no download).
+
+    Returns True if the model repo exists and has tokenizer files, False otherwise.
+    The actual tokenizer is loaded later by the MetricsAggregator subprocess and
+    by Harmony transforms (each loads their own instance as needed).
```
arekay-nv: Can you update the doc to indicate that this doesn't load the tokenizer and cite examples of the MetricsAggregator and Harmony transforms - the above reads like it is hardwired, but it should be extensible.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -289,115 +320,331 @@ def setup_benchmark(config: BenchmarkConfig, test_mode: TestMode) -> BenchmarkCo
...
+    # Counter keys (from MetricCounterKey enum)
+    for key in [
+        "total_samples_issued",
+        "total_samples_completed",
+        "total_samples_failed",
+        "tracked_samples_issued",
+        "tracked_samples_completed",
+        "tracked_duration_ns",
+        "total_duration_ns",
+    ]:
```
arekay-nv: Should this not be using just the keys from the enum for consistency?

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -289,115 +320,325 @@ def setup_benchmark(config: BenchmarkConfig, test_mode: TestMode) -> BenchmarkCo
...
+    map_path = ctx.report_dir / "sample_idx_map.json"
+    with map_path.open("wb") as f:
+        f.write(msgspec.json.format(msgspec.json.encode(sample_idx_map), indent=2))
+    logger.debug(f"Wrote {map_path}")
+
+    # Copy events.jsonl from tmpfs to report_dir
+    _salvage_tmpfs(ctx.report_dir, tmpfs_dir)
+
+    # Clean up tmpfs
+    shutil.rmtree(tmpfs_dir, ignore_errors=True)
```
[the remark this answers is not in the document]
arekay-nv: +1

arekay-nv (approved):
Awesome work - thanks. 
I will run some perf test offline, but looks good!

docs/async_utils/transport/zmq/ready_check_design.md
nv-alicheng: What do you mean by `await(N)`? It's `await wait(N)` where `wait(N)` blocks until N ready signals have arrived.

docs/metrics/report_design.md
nv-alicheng: They should be, yes.

tests/unit/load_generator/test_async_session.py
```diff
@@ -0,0 +1,873 @@
...
+    SessionResult,
+)
+from inference_endpoint.metrics.metric import Throughput
+
+# ---------------------------------------------------------------------------
+# Test doubles
+# ---------------------------------------------------------------------------
+
+
+class FakeDataset(Dataset):
```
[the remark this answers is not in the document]
nv-alicheng: Intentional.

src/inference_endpoint/load_generator/session.py
```diff
@@ -13,327 +13,460 @@
...
+
+class SampleIssuer(Protocol):
+    """Sends queries to an endpoint and receives responses.
+
+    Matches HTTPEndpointClient's interface: issue (sync ZMQ push),
+    recv (async ZMQ recv), shutdown.
+    """
+
+    def issue(self, query: Query) -> None: ...
+    async def recv(self) -> QueryResult | StreamChunk | None: ...
```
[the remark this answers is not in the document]
nv-alicheng: False positive - protocols are duck-typed not strict. There is no inheritance involved and protocols cannot be instantiated anyway.

configs/gptoss_test.yaml
nv-alicheng: Removed in https://github.com/mlcommons/endpoints/pull/288

tests/unit/async_utils/services/metrics_aggregator/test_kv_store.py
nv-alicheng: If empty, undefined behavior, doesn't matter to check here.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
nv-alicheng: Just for debugging to provide a log that the aggregator is actively computing during session drain in the case where the aggregator is lagging behind.

src/inference_endpoint/commands/benchmark/execute.py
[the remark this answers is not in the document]
nv-alicheng: Removed the redundant `shutil.rmtree` from `_write_scoring_artifacts`. Cleanup is now owned solely by the `finally` block in `run_benchmark`, eliminating the race.

_Generated by Claude_

configs/gptoss_test.yaml
```diff
@@ -0,0 +1,66 @@
...
+metrics:
+  collect:
+    - "throughput"
+    - "latency"
+    - "ttft"
+    - "tpot"
+
+endpoint_config:
+  endpoints:
+    - "http://umbriel-b200-145.nvidia.com:8000"
```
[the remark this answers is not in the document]
nv-alicheng: Replaced the internal hostname with `http://localhost:8000`, and the file was subsequently removed entirely as it wasn't intended to ship (see the thread on this file at line 1).

_Generated by Claude_

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -289,115 +320,325 @@ def setup_benchmark(config: BenchmarkConfig, test_mode: TestMode) -> BenchmarkCo
...
+    ]:
+        reader.register_key(key, "counter")
+    # Series keys (from MetricSeriesKey enum)
+    for key in ["isl", "osl", "sample_latency_ns"]:
+        reader.register_key(key, "series")
+    reader.register_key("tpot_ns", "series", dtype=float)
+    if streaming:
+        for key in ["ttft_ns", "chunk_delta_ns"]:
+            reader.register_key(key, "series")
+    return reader
```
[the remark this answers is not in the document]
nv-alicheng: `_setup_kv_reader` now iterates `MetricCounterKey` and `MetricSeriesKey` instead of hardcoding string literals, so the reader stays in sync with the aggregator automatically.

_Generated by Claude_

configs/gptoss_test.yaml
nv-alicheng: Agreed — the file was local test scaffolding that shouldn't have been committed. Removed entirely.

_Generated by Claude_

tests/unit/async_utils/services/metrics_aggregator/test_kv_store.py
nv-alicheng: Added asserts pinning `stats.min_val == math.inf` and `stats.max_val == -math.inf` on empty `SeriesStats`, with a comment noting that `compute_summary()` is responsible for normalizing these sentinels to 0.

_Generated by Claude_

tests/unit/async_utils/services/metrics_aggregator/test_kv_store.py
nv-alicheng: Added `test_snapshot_is_isolated_from_later_writes`: writes to `n_issued` and `latency` after `snapshot()` and asserts the snapshot values are unchanged.

_Generated by Claude_

docs/metrics/report_design.md
nv-alicheng: Added a 'Reports are reproducible from the event log' section under Design Principles, spelling out the event log as the authoritative record and requiring new metrics to derive only from event fields.

_Generated by Claude_

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
nv-alicheng: Renamed `subtract_field` → `delta_start_fieldname` on `TimeDeltaTrigger`, with the docstring reframed to describe the parameter as the start of a time delta rather than the mathematical operand. Call sites in `TtftTrigger`, `ChunkDeltaTrigger`, and `SampleLatencyTrigger` updated.

_Generated by Claude_

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
nv-alicheng: Reworded to make SGLang an example rather than the defining condition: the fast path triggers whenever `token_ids` is populated, regardless of backend.

_Generated by Claude_

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
nv-alicheng: Same treatment — OpenAI now reads as an example of a backend without `token_ids`, not a defining condition for the async path.

_Generated by Claude_

tests/unit/transport/test_zmq_pool_transport.py
nv-alicheng: Done via the consolidation below — the four `test_{N}_workers` methods are collapsed into a single `@pytest.mark.parametrize('num_workers', [2, 3, 4, 8])`.

_Generated by Claude_

tests/unit/transport/test_zmq_pool_transport.py
nv-alicheng: `TestZmqPoolTransportWithPublisher` and `TestZmqPoolTransportWithoutPublisher` are now a single `TestZmqPoolTransport` class parametrized over `(num_workers, create_publisher)`.

_Generated by Claude_

tests/unit/metrics/test_report_builder.py
nv-alicheng: Added `assert s['max'] == 0` and `assert s['std_dev'] == 0` to `test_empty`.

_Generated by Claude_

tests/unit/load_generator/test_sample_order.py
nv-alicheng: Both `TestWithoutReplacementSampleOrder` and `TestWithReplacementSampleOrder` now parameterize every test over `[3, 100, 10_000]` to exercise small/medium/large dataset sizes.

_Generated by Claude_

tests/unit/async_utils/services/metrics_aggregator/test_aggregator.py
nv-alicheng: Added explanatory messages to the tracking-window asserts in `TestTrackingWindow` — failure output now names the invariant being violated (e.g. 'Sample issued before START_PERFORMANCE_TRACKING must not create a table row').

_Generated by Claude_

src/inference_endpoint/commands/benchmark/execute.py
nv-alicheng: Rewrote the `_check_tokenizer_exists` docstring: explicitly says 'this function is a probe — it never loads or downloads', and reframes MetricsAggregator/Harmony as examples with 'any future plugin with its own tokenization need' to signal extensibility.

_Generated by Claude_

src/inference_endpoint/commands/benchmark/execute.py
nv-alicheng: Addressed alongside the earlier gemini-bot thread on the same function — `_setup_kv_reader` now iterates `MetricCounterKey` and `MetricSeriesKey` instead of hardcoding strings.

_Generated by Claude_

tests/unit/load_generator/test_sample_order.py
nv-alicheng: Can add this in the future, it'll get weird if expected results are checking for proper uses of the RNG seed (i.e. checking exact values)

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
nv-alicheng: I think we can discuss this with finalized rules on how to treat warmup.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
nv-alicheng: I think it's fine? If anything I think it can be
```
if isinstance(ev, SessionEventType):
    ...
```
since Session-level events are treated as control signals. I'll leave it as is for now, I don't think this is too significant.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -289,115 +329,324 @@ def setup_benchmark(config: BenchmarkConfig, test_mode: TestMode) -> BenchmarkCo
...
-    try:
-        api_type: APIType = config.endpoint_config.api_type
-        http_config = config.settings.client.with_updates(
-            endpoint_urls=[urljoin(e, api_type.default_route()) for e in endpoints],
-            api_type=api_type,
-            api_key=config.endpoint_config.api_key,
-            event_logs_dir=ctx.report_dir,
-            cpu_affinity=ctx.affinity_plan,
+    # ZMQ context for event publishing + service launcher
+    with ManagedZMQContext.scoped(io_threads=2) as zmq_ctx:
```
viraatc, line 419: benefit of num-threads=4 was measureable when we had old BenchmarkSession design, i wonder if it still show up in e2e runs?

viraatc (approved):
looks great! thanks!

src/inference_endpoint/commands/benchmark/execute.py
nv-alicheng: Should we consider having this be a tunable value via configs? Or should we just manually test it?

tests/unit/async_utils/services/metrics_aggregator/conftest.py
nv-alicheng: Not high priority, we can add in the future if it somehow does actually cause issues, but in general the tokenizers rust backends are extremely fast anyway.

src/inference_endpoint/commands/benchmark/execute.py
viraatc: fine for now lets merge MR,

we already have a flag in the zmq/trasport.py which this should be hooked up to: 

https://github.com/mlcommons/endpoints/blob/ed1704e5402352ed4fed335722a1a321fcd40c71/src/inference_endpoint/async_utils/transport/zmq/transport.py#L96

might want to move it out of transport.py since its general zmq config not just zmq-transport.

PR #291, opened 2026-04-22
fix: let HTTPClientConfig and template regen work on non-Linux

viraatc:
## Summary

Two independent issues broke `pre-commit run` (and any `HTTPClientConfig()` construction) on macOS. Fixing both:

**(A) `src/inference_endpoint/endpoint_client/config.py`** — `_get_auto_num_workers` now catches `UnsupportedPlatformError` from `get_current_numa_node` / `get_cpus_in_numa_node` and falls back to `min_workers=10`. Matches the existing "NUMA not discoverable" branch. Unbricks the CLI on macOS too.

**(B) `scripts/regenerate_templates.py`** — `_dump_defaults` was documented to avoid running model validators, but it called `default_factory()` on every field, which constructs nested models (and runs their validators) whenever the factory is a `BaseModel` subclass. Now it recurses into `_dump_defaults(factory)` when the factory itself is a `BaseModel` subclass. Factories that dynamically pick a concrete subclass (e.g. `TransportConfig.create_default → ZMQTransportConfig`) are not types, so they still get called and the resulting instance is walked via the existing branch.

Either fix alone unbricks pre-commit on macOS — (A) makes the validator safe to run, (B) stops the script from running it. Together they're belt-and-suspenders: (A) fixes the product, (B) hardens the tooling against future validators adding side effects.

## Template output is unchanged (and platform-independent)

Generated templates are byte-identical before and after this change, and identical across macOS / Linux. `_dump_defaults` emits class-level declared defaults (e.g. `num_workers: -1` — the "auto" sentinel), not the validator-resolved values. Even on Linux before this PR, the resolved value from `HTTPClientConfig()` was silently discarded because the serializer re-walks the class (not the instance). So the validator's NUMA-aware resolution never landed in the template file — it was only ever a runtime concern. This PR just stops the pointless construction.

## Alternatives considered

- **LBYL platform guard** (exporting `IS_LINUX` from `cpu_affinity.py` and checking it at the top of `_get_auto_num_workers`): arguably more idiomatic Python for a non-racy condition, and could replace the `try`/`except` in (A). Kept the current form for minimal diff; happy to swap in a follow-up.
- **Push fallback into `cpu_affinity.py`** (drop `@require_linux` from read-only queries, keep it on mutators like `pin_loadgen`): the most principled shape — "errors for exceptional conditions, sentinels for absent data" — but a broader change that needs `AttributeError` handling for `os.sched_getaffinity` on macOS. Deferred.

## Test plan

- [x] Linux: `pre-commit run --all-files` passes, generated templates are byte-identical
- [x] Simulated non-Linux (`UnsupportedPlatformError` patched to always raise): `HTTPClientConfig()` constructs with `num_workers=10`; regen script completes without entering the NUMA code path
- [ ] Real macOS: `pre-commit run --all-files` on darwin


nv-alicheng:
Verified on MacOS, thanks!

nv-alicheng (approved):

PR #305, opened 2026-05-05
feat: Plug in warmup phase

arekay-nv:
Wires the warmup phase into the standard benchmark execution flow as a config-driven feature, replacing the old ENABLE_WARMUP=1 env-var gate.                                      
                                                                                                                                                                                     
  Config — new WarmupConfig on Settings with CLI shorthands: --warmup, --warmup-requests, --warmup-salt, --warmup-drain, --warmup-seed.                                              
                                                                                                                                                                                     
  Prompt salting — Dataset.with_salt(rng) returns a shallow copy that prepends a seeded [<16-hex>]  prefix on every load_sample() call to defeat KV-cache reuse. Handles             
  plain-string, multimodal (image-first), and pre-tokenized prompts (warns when salting is impossible).                                                                              
                                                                                                                                                                                     
  Execution — warmup is prepended as a PhaseType.WARMUP phase before perf. Independent seeded RNGs keep warmup scheduling reproducible and isolated from the perf phase. The global  
  max_duration_ms timer starts only when the perf phase begins, so warmup time isn't charged against the perf budget. Drain is now config-driven via PhaseConfig.drain_after rather
  than hardcoded.                                                                                                                                                                    
                                                                         
  Session — warmup responses are suppressed from metrics (COMPLETE event publish and _on_sample_complete callback both gated on current phase). _drain_inflight is bounded at 240 s  
  to prevent hangs.
                                                                                                                                                                                     
  Test infrastructure — EchoServer accepts an optional sync/async request_handler; new mock_http_echo_server_factory fixture supports multiple servers with distinct handlers per    
  test.
                                                                                                                                                                                     
  Tests — integration tests for salt and drain behaviour (test_warmup.py), unit tests for Dataset.with_salt() (test_salted_dataset.py), and fixture tests                            
  (test_http_mock_fixtures.py).

<!-- Brief description of the changes -->

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Related issues

<!-- Link any related issues: Closes #123 -->

## Testing

- [x] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [x] Pre-commit hooks pass


src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -411,7 +411,7 @@ def num_samples(self) -> int:
     @classmethod
     def get_dataloader(
         cls,
-        datasets_dir: Path = Path("datasets"),
+        datasets_dir: Path = Path("dataset_cache"),
```
arekay-nv, line 472: **[Review Council — Claude] high · api-contract**

Default `datasets_dir` changed from `Path("datasets")` to `Path("dataset_cache")`:

```python
datasets_dir: Path = Path("dataset_cache"),
```

This is a silent breaking change. Any existing user who relied on the previous default will, after upgrade, fail to find the cached dataframe under `datasets/<DATASET_ID>/<variant>` and will silently re-download/re-generate (potentially many GB and minutes of work). The `.gitignore` change only ignores `dataset_cache/`, leaving any pre-existing `datasets/` cache uncovered.

Suggest: keep `Path("datasets")` as the default, OR add a fallback that reads from `datasets/` if `dataset_cache/` is empty, OR document the migration prominently. At minimum, decouple this rename from the warmup PR — it's an orthogonal change.

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -431,6 +431,56 @@ def get_dataloader(
...
+    def load(
+        self,
+        adapter: "HttpRequestAdapter | None" = None,
+        api_type: APIType | None = None,
+        model_params: ModelParams | None = None,
+        force: bool = False,
+    ) -> None:
+        pass  # Inner dataset already loaded
+
+    def load_sample(self, index: int) -> Any:
```
arekay-nv: **[Review Council — Claude] high · bug**

`SaltedDataset.load_sample()` only mutates `data["prompt"]`, but several adapters bypass the prompt entirely. In `src/inference_endpoint/sglang/adapter.py:90`:

```python
input_tokens = query.data["input_tokens"]
return cls._request_encoder.encode(
    SGLangGenerateRequest(input_ids=input_tokens, ...))
```

When `salt=True` is used with a tokenized dataset (any pipeline that pre-tokenizes — common in MLPerf submissions), the salt prefix is added to the (unused) text but `input_tokens` are unchanged, so the server receives the **same** token-ID sequence for every warmup request. KV-cache reuse is NOT prevented. The class docstring's claim that "each call to load_sample() generates a fresh salt, so reused samples … each receive a distinct salt" is silently violated.

Fix options: (a) detect `input_tokens`/`token_ids` keys and either drop them or re-tokenize after salting; (b) raise an error at warmup setup if `salt=True` and the inner dataset has tokenized fields; (c) explicitly document the restriction.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -347,6 +348,34 @@ def _build_phases(ctx: BenchmarkContext) -> list[PhaseConfig]:
...
+        )
+        warmup_rt = RuntimeSettings(
+            metric_target=ctx.rt_settings.metric_target,
+            reported_metrics=ctx.rt_settings.reported_metrics,
+            min_duration_ms=0,
+            max_duration_ms=None,
+            n_samples_from_dataset=ctx.dataloader.num_samples(),
+            n_samples_to_issue=warmup_cfg.n_requests,
+            min_sample_count=1,
+            rng_sched=random.Random(),
```
arekay-nv: **[Review Council — Both (Codex P2 + Claude)] high · bug**

`rng_sched=random.Random()` and `rng_sample_index=random.Random()` are **unseeded**:

```python
rng_sched=random.Random(),
rng_sample_index=random.Random(),
```

With no seed argument, `random.Random()` is seeded from system time/urandom, so:

1. Warmup sample order is nondeterministic regardless of the perf-phase `runtime.scheduler_random_seed`/`dataloader_random_seed`.
2. Two runs of the same config produce different warmup workloads, defeating the purpose of warmup as a deterministic preconditioning step.
3. With `warmup.n_requests < dataset_size` (or `salt=false`), the warmup phase touches a different sample subset each time, changing which requests prime the server cache and making the perf phase non-reproducible.

The accuracy phase below at line 398 correctly threads `ctx.rt_settings.rng_sched/rng_sample_index` — warmup is the only phase using unseeded RNGs.

Fix: derive the warmup seeds from the perf-phase config so warmup is reproducible, but in a way that doesn't consume perf-phase RNG state (the existing `test_warmup_uses_independent_rng_instances` correctly enforces independence). E.g. seed warmup with `hash((perf_seed, "warmup")) & ((1<<32)-1)`.

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -431,6 +431,56 @@ def get_dataloader(
...
+        data = self._inner.load_sample(index)
+        if not (isinstance(data, dict) and "prompt" in data):
+            return data
+        prompt = data["prompt"]
+        salt = os.urandom(8).hex()
+        if isinstance(prompt, str):
+            return {**data, "prompt": f"[{salt}] {prompt}"}
+        if (
```
arekay-nv: **[Review Council — Both (Codex P2 + Claude)] medium · bug**

Multimodal salting only fires when `prompt[0]` is a text part:

```python
if (isinstance(prompt, list) and prompt and isinstance(prompt[0], dict)
        and prompt[0].get("type") == "text"):
    salted_parts = [{**prompt[0], "text": f"[{salt}] ..."}, *prompt[1:]]
```

Real-world OpenAI-style multimodal prompts often start with an image/audio part and place text at index 1+ ("look at this image, then answer the question"). For those, salting is silently skipped (falls through to the `return data` at line 478) and `warmup.salt=true` does NOT prevent KV-cache reuse for those requests. This breaks the advertised warmup behavior for a common multimodal input shape.

Fix: scan the list for the first dict with `type == "text"` (or salt all text parts) instead of only inspecting `prompt[0]`.

arekay-nv (commented):

Reviewed by **Codex** + **Claude** | Depth: **thorough**

Found **14** issues across **6** files: 0 critical, 3 high, 4 medium, 7 low.

See inline comments for details. A grouped summary will follow as a separate top-level comment.

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -431,6 +431,56 @@ def get_dataloader(
...
+    """Wraps a loaded Dataset, prepending a unique random salt to each prompt on load_sample().
+
+    Each call to load_sample() generates a fresh salt, so reused samples (when
+    n_requests > dataset size) each receive a distinct salt.
+    """
+
+    def __init__(self, inner: Dataset) -> None:
+        # Skip Dataset.__init__ — all state is delegated to inner
+        self._inner = inner
+        self.data = inner.data
```
arekay-nv: **[Review Council — Claude] medium · bug**

`SaltedDataset.__init__` snapshots `inner.data` at construction time:

```python
self._inner = inner
self.data = inner.data  # snapshot, not a view
```

If anything calls `inner.load(force=True)` later, `Dataset.load()` (line 384) reassigns `self.data = df.to_dict(orient="records")` — a new list. `SaltedDataset.data` keeps a reference to the **old** list while `self._inner.load_sample()` reads the new list. Users debugging via `sd.data` see stale rows. Even if reload is rare today, this latent inconsistency tends to surface during refactors.

Fix: make `data` a property that delegates to `self._inner.data` (and the same for `repeats`, which is also snapshotted at line 447).

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -347,6 +348,34 @@ def _build_phases(ctx: BenchmarkContext) -> list[PhaseConfig]:
     """Build the phase list from BenchmarkContext."""
     phases: list[PhaseConfig] = []
 
+    # Warmup phase (optional, before performance)
+    warmup_cfg = ctx.config.settings.warmup
+    if warmup_cfg.enabled:
+        warmup_dataset: Dataset = (
+            SaltedDataset(ctx.dataloader) if warmup_cfg.salt else ctx.dataloader
+        )
+        warmup_rt = RuntimeSettings(
```
arekay-nv: **[Review Council — Claude] medium · design**

The warmup `RuntimeSettings` is built ad-hoc rather than derived from the perf settings:

```python
warmup_rt = RuntimeSettings(
    metric_target=ctx.rt_settings.metric_target,
    reported_metrics=ctx.rt_settings.reported_metrics,
    min_duration_ms=0,
    ...
    load_pattern=LoadPattern(type=LoadPatternType.MAX_THROUGHPUT),
)
```

Two consequences:

1. Every field is duplicated by hand — if `RuntimeSettings` adds a new required/load-bearing field later (e.g. CPU pinning, sample selection mode), the warmup phase will silently miss it and diverge from perf behavior.
2. The load pattern is hardcoded to `MAX_THROUGHPUT`. This makes it impossible to warm up at the same QPS or concurrency the perf run uses, which is the most realistic warmup pattern for production-mirroring benchmarks (the cache state at QPS=N can be very different from the cache state under burst load).

Fix: derive via `dataclasses.replace(ctx.rt_settings, ...)` and let `WarmupConfig` carry an optional `load_pattern` override that defaults to perf's load pattern.

src/inference_endpoint/load_generator/session.py
```diff
@@ -274,12 +273,6 @@ async def run(self, phases: list[PhaseConfig]) -> SessionResult:
             for phase in phases:
```
arekay-nv, line 357: **[Review Council — Claude] medium · api-contract**

The `_WARMUP_ENABLED = os.environ.get("ENABLE_WARMUP") == "1"` gate (and the `ENABLE_WARMUP=1 to enable` log message) was removed. Anyone who scripted `ENABLE_WARMUP=1 inference-endpoint ...` to opt in to warmup will now silently see no behavior change — the env var is ignored, and warmup runs only when the YAML enables it.

Suggested fix: log a one-time warning at session start if `ENABLE_WARMUP` is set in the environment but the YAML doesn't enable warmup; OR document the rename in release notes. (Either way, the silent no-op is the worst case.)

src/inference_endpoint/load_generator/session.py
```diff
@@ -68,6 +65,7 @@ class PhaseConfig:
     runtime_settings: RuntimeSettings
     dataset: Dataset
     phase_type: PhaseType = PhaseType.PERFORMANCE
+    drain_after: bool = True
```
arekay-nv, line 92: **[Review Council — Claude] low · api-contract**

`drain_after: bool = True` was added to `PhaseConfig` with default `True`, but for `PhaseType.WARMUP` the previous behavior was effectively `drain_after=False` (the old `_run_phase` skipped drain for warmup unconditionally). Existing callers that construct `PhaseConfig(..., PhaseType.WARMUP)` directly (e.g. via the Python API or in tests) will now get the **opposite** drain semantics from before unless they pass `drain_after=False` explicitly. The PR's `_build_phases` correctly passes `warmup_cfg.drain` through, but third-party code paths get a silent behavior flip.

Fix: either default `drain_after` based on `phase_type` via a `__post_init__` (warmup → False, others → True), OR make `drain_after` a required keyword for warmup phases.

src/inference_endpoint/load_generator/session.py
```diff
@@ -425,7 +418,10 @@ def _handle_response(self, resp: QueryResult | StreamChunk) -> None:
                     self._drain_event.set()
                 if self._current_strategy:
                     self._current_strategy.on_query_complete(query_id)
-                if self._on_sample_complete:
+                if (
+                    self._on_sample_complete
+                    and self._current_phase_type != PhaseType.WARMUP
```
arekay-nv, line 551: **[Review Council — Claude] low · data-integrity**

The `_on_sample_complete` callback is filtered for warmup, but the `SampleEventType.COMPLETE` event publish at line 396 (just above) is NOT filtered:

```python
self._publisher.publish(EventRecord(
    event_type=SampleEventType.COMPLETE, ..., sample_uuid=query_id, ...
))
```

When `drain_after=False` (warmup default), late-arriving warmup `QueryResult`s race the start of the perf phase. `MetricsTable.update_sample` discards events for unknown sample IDs so this doesn't corrupt counters today — but the events still travel over ZMQ, are decoded, and (if event_logger is enabled) get persisted to the JSONL/SQLite event log. Downstream tooling reading the raw event log will see warmup-named samples interleaved with perf samples without any phase tag.

Fix: tag warmup queries with a phase marker on the `EventRecord` (or the `Query`) so the event logger can filter or annotate, OR add the `_current_phase_type != PhaseType.WARMUP` check around the publish at line 396 as well.

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
```diff
@@ -91,10 +92,16 @@ async def main() -> None:
...
     # (coalesces to 'object' not 'AbstractContextManager[TokenizePool | None]')
+    pool_cm: AbstractContextManager[TokenizePool | None]
     if args.tokenizer:
-        pool_cm: AbstractContextManager[TokenizePool | None] = TokenizePool(
-            args.tokenizer, n_workers=args.tokenizer_workers
-        )
+        try:
+            pool_cm = TokenizePool(args.tokenizer, n_workers=args.tokenizer_workers)
+        except Exception as e:
+            logging.warning(
```
arekay-nv: **[Review Council — Claude] low · error-handling**

Tokenizer load failure is silently downgraded to "no token metrics":

```python
except Exception as e:
    logging.warning(
        f"Failed to load tokenizer '{args.tokenizer}': {e}. "
        "ISL/OSL/TPOT token metrics will be unavailable."
    )
    pool_cm = nullcontext()
```

A user who explicitly passes `--tokenizer X` and thus expects token metrics will instead see a single WARNING line buried in the log, then a silent benchmark run with empty ISL/OSL/TPOT columns. By the time they check the report, the run is over.

Fix: re-raise unless an explicit `--allow-tokenizer-fallback` flag is set, OR at least log at ERROR level and surface the missing-tokenizer state in the final report. Also use the module logger (`logger = logging.getLogger(__name__)`) rather than the root `logging` module.

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -431,6 +431,56 @@ def get_dataloader(
...
+        force: bool = False,
+    ) -> None:
+        pass  # Inner dataset already loaded
+
+    def load_sample(self, index: int) -> Any:
+        data = self._inner.load_sample(index)
+        if not (isinstance(data, dict) and "prompt" in data):
+            return data
+        prompt = data["prompt"]
+        salt = os.urandom(8).hex()
```
arekay-nv: **[Review Council — Claude] low · design**

`salt = os.urandom(8).hex()` on every `load_sample()` ignores the seeded `rng_sample_index` in `RuntimeSettings`. Reproducibility-conscious users who set `dataloader_random_seed` to compare runs will see different salted prompts on each invocation. Warmup metrics aren't reported, but the actual prompts sent to the server (and thus server-side caches/stats) are nondeterministic.

Fix: thread the `rng_sample_index` (or a derived `random.Random`) through `SaltedDataset.__init__` and use `f'{rng.getrandbits(64):016x}'` to keep determinism with a non-zero seed.

tests/unit/commands/test_benchmark.py
```diff
@@ -385,6 +387,378 @@ def test_valid_templates_parse(self, template):
...
+        assert warmup.enabled is False
+        assert warmup.n_requests is None
+
+
+class TestBuildPhases:
+    """Tests for _build_phases() in execute.py."""
+
+    @pytest.fixture
+    def base_rt_settings(self):
+        import random
```
arekay-nv: **[Review Council — Claude] low · design**

`TestBuildPhases` uses lazy/inline imports inside fixtures and ~15 test method bodies (`import random`, `from inference_endpoint.commands.benchmark.execute import _build_phases`, etc.).

AGENTS.md explicitly forbids this:

> All imports must be at the top of the file. Imports inside functions, methods, or conditional blocks (other than `TYPE_CHECKING`) are disallowed.

None of the documented exceptions apply (no circular import, no optional dep, no sandboxing). Move the imports to the top of the file.

tests/integration/commands/test_warmup.py
```diff
@@ -0,0 +1,435 @@
...
+    * Use ``salt=True`` to identify warmup vs perf requests at the server.
+    * Track concurrent in-flight counts inside the server handler.  Since all
+      handlers share a single asyncio event loop, plain Python ints are safe
+      (no actual concurrent mutation — only cooperative interleaving at
+      ``await`` points).
+    * ``run_benchmark`` is synchronous and blocks until the perf phase drains,
+      guaranteeing all overlap events have been recorded before we assert.
+    """
+
+    _DELAY = 0.15  # seconds; long enough for perf to start while warmup is pending
```
arekay-nv, line 292: **[Review Council — Claude] low · testing**

`_DELAY = 0.15` is fragile for the overlap-observation tests:

```python
_DELAY = 0.15  # seconds; long enough for perf to start while warmup is pending
```

On a heavily-loaded CI runner, the entire warmup batch (5 requests) can plausibly round-trip within 150 ms before the perf phase issues, producing a false "no overlap" and flaking `test_drain_false_overlap_observed` / `test_drain_false_max_concurrent_exceeds_single_phase_size`.

Fix: use a controllable handler that explicitly blocks until the test signals release (e.g. via `asyncio.Event`) instead of a fixed sleep. If sticking with a sleep, increase to ≥1s.

tests/integration/commands/test_warmup.py
```diff
@@ -0,0 +1,435 @@
...
+from inference_endpoint.core.types import QueryResult, TextModelOutput
+from inference_endpoint.endpoint_client.config import HTTPClientConfig
+from inference_endpoint.openai.openai_adapter import OpenAIAdapter
+
+# ── helpers ──────────────────────────────────────────────────────────────────
+
+_SALT_RE = re.compile(r"^\[([0-9a-f]{16})\] (.+)$")
+
+_MINIMAL_CLIENT = HTTPClientConfig(
+    num_workers=1, warmup_connections=0, max_connections=10
```
arekay-nv, line 65: **[Review Council — Claude] low · testing**

All warmup integration tests run with `num_workers=1`:

```python
_MINIMAL_CLIENT = HTTPClientConfig(
    num_workers=1, warmup_connections=0, max_connections=10
)
```

None cover the multi-worker warmup path, where ZMQ ordering between worker processes matters and where salted prompts must be distinct across workers. Suggest adding at least one warmup test with `num_workers >= 2` to exercise the multi-process distribution path the production code uses.

arekay-nv:

Reviewed by **Codex** + **Claude** | Depth: **thorough** | HEAD: `56258be`

Found **14** issues across **6** files. Inline comments posted in [this review](https://github.com/mlcommons/endpoints/pull/305#pullrequestreview-4225024755).

Cross-reviewer agreement (boosted confidence): **execute.py:365** (warmup RNG seeding) and **dataset.py:467** (multimodal salt) were flagged by both Codex and Claude.

### 🔴 Must Fix (high)

| 1 | `src/inference_endpoint/dataset_manager/dataset.py` | 414 | api-contract | Claude | Default `datasets_dir` silently changed `datasets/` → `dataset_cache/` (orthogonal to warmup; risks losing existing caches) |
| 2 | `src/inference_endpoint/dataset_manager/dataset.py` | 459 | bug | Claude | `SaltedDataset.load_sample()` mutates `prompt` only, leaving `input_tokens` (used by SGLang adapter) unchanged → KV-cache reuse not actually prevented for tokenized datasets |
| 3 | `src/inference_endpoint/commands/benchmark/execute.py` | 365 | bug | Both | Warmup `rng_sched` / `rng_sample_index` use bare `random.Random()` (unseeded) → warmup sample order is nondeterministic across runs, breaking reproducibility |

### 🟡 Should Fix (medium)

| 1 | `src/inference_endpoint/dataset_manager/dataset.py` | 467 | bug | Both | Multimodal salting only fires when `prompt[0]` is text; image-first prompts (common shape) silently skip salting |
| 2 | `src/inference_endpoint/dataset_manager/dataset.py` | 444 | bug | Claude | `SaltedDataset.data` snapshots `inner.data` at construction; later `inner.load(force=True)` desyncs the wrapper from the inner dataset |
| 3 | `src/inference_endpoint/commands/benchmark/execute.py` | 357 | design | Claude | `warmup_rt` built ad-hoc instead of via `dataclasses.replace`; load pattern hardcoded to `MAX_THROUGHPUT` (no way to warm up at the perf-phase QPS/concurrency) |
| 4 | `src/inference_endpoint/load_generator/session.py` | 273 | api-contract | Claude | `ENABLE_WARMUP=1` env-var escape hatch removed without a deprecation warning; existing scripts now silently no-op |

### 🔵 Consider (low)

| 1 | `src/inference_endpoint/load_generator/session.py` | 68 | api-contract | Claude | `PhaseConfig.drain_after=True` default flips warmup drain semantics for direct callers (was effectively `False`) |
| 2 | `src/inference_endpoint/load_generator/session.py` | 423 | data-integrity | Claude | `drain_after=False` filter only covers `_on_sample_complete`; warmup `COMPLETE` events still publish to ZMQ → event log pollution |
| 3 | `src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py` | 100 | error-handling | Claude | Tokenizer load failure silently downgraded to `nullcontext()`; user-requested `--tokenizer` produces empty ISL/OSL/TPOT with only a warning |
| 4 | `src/inference_endpoint/dataset_manager/dataset.py` | 464 | design | Claude | `os.urandom(8)` salt ignores seeded RNG → salted prompts are nondeterministic even when the user sets `dataloader_random_seed` |
| 5 | `tests/unit/commands/test_benchmark.py` | 471 | design | Claude | `TestBuildPhases` uses lazy/inline imports throughout (~15 sites) — explicit AGENTS.md violation |
| 6 | `tests/integration/commands/test_warmup.py` | 292 | testing | Claude | `_DELAY = 0.15` is fragile for overlap-observation tests — risk of CI flakes; replace with explicit `asyncio.Event` synchronization |
| 7 | `tests/integration/commands/test_warmup.py` | 65 | testing | Claude | All warmup integration tests use `num_workers=1`; multi-worker warmup path (ZMQ ordering, salt distinctness) is uncovered |


---

**Theme summary**

The warmup feature lands the structural plumbing well, but several correctness/reproducibility properties the YAML knobs claim to provide are not actually enforced end-to-end:

- **Reproducibility** — warmup RNGs are unseeded (execute.py:365) and the salt uses `os.urandom` (dataset.py:464). Two runs of the same config produce different warmup workloads, indirectly affecting perf-phase reproducibility when warmup sizes the cache.
- **Salt completeness** — `SaltedDataset` only handles plain-text prompts. Tokenized datasets (SGLang adapter via `input_tokens`) and image-first multimodal prompts silently skip salting, defeating `warmup.salt=true` for those paths.
- **API/back-compat surface** — the `ENABLE_WARMUP` env var was removed without a deprecation warning; the new `PhaseConfig.drain_after=True` default flips warmup drain semantics for direct callers; `datasets_dir` default rename couples an orthogonal cache-directory change to this PR.
- **Test coverage** — single-worker only; the multi-worker ZMQ ordering and salt-distinctness across worker processes is the riskiest part of the architecture and is currently unexercised. The `_DELAY = 0.15` overlap tests are also flake-prone.

🤖 Posted by `/review-council` skill.

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -431,6 +431,56 @@ def get_dataloader(
         return cls(df, transforms=transforms, repeats=num_repeats)
 
 
+class SaltedDataset(Dataset):
```
arekay-nv: **[Review Council — Claude] medium · design**

`SaltedDataset` inherits from `Dataset` without opting out of `__init_subclass__` registration. `Dataset.__init_subclass__` (line 276) registers every concrete subclass under its class name in `Dataset.PREDEFINED`:

```python
cls.DATASET_ID = dataset_id  # → "SaltedDataset"
Dataset.PREDEFINED[dataset_id] = cls
```

`SaltedDataset.__init__` takes `inner: Dataset` (not `dataframe: pd.DataFrame`), so any code that tries to instantiate it through the registry (e.g., `DataLoaderFactory` or the YAML `dataset_id` lookup) will raise a `TypeError` with a confusing error. Fix: pass `dataset_id=None` to explicitly opt out:

```python
class SaltedDataset(Dataset, dataset_id=None):
```

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -431,6 +431,56 @@ def get_dataloader(
...
+            and prompt
+            and isinstance(prompt[0], dict)
+            and prompt[0].get("type") == "text"
+        ):
+            salted_parts = [
+                {**prompt[0], "text": f"[{salt}] {prompt[0]['text']}"},
+                *prompt[1:],
+            ]
+            return {**data, "prompt": salted_parts}
+        return data  # unsupported prompt type — skip salting
```
arekay-nv, line 463: **[Review Council — Claude] medium · bug**

When the prompt is a multimodal list where ALL parts are non-text (e.g., all `image_url`), the loop finds no text part and falls through to `return data` with no warning:

```python
if isinstance(prompt, list) and prompt:
    for i, part in enumerate(prompt):
        if isinstance(part, dict) and part.get("type") == "text":
            ...
            return {**data, "prompt": salted_parts}
return data  # unsupported prompt type — skip salting (silent!)
```

Unlike the `input_tokens` case (which emits a `logger.warning`), this fallback is completely silent. Users who enable `salt=True` for image-only multimodal datasets will not see any indication that KV-cache bypass is not functioning. Add a `logger.warning` consistent with the `input_tokens` path.

src/inference_endpoint/commands/benchmark/execute.py
arekay-nv: **[Review Council — Claude] medium · design**

After fixing the unseeded RNGs, both `rng_sched` and `rng_sample_index` for the warmup phase will be seeded with the same value (`warmup_cfg.random_seed`, default `0`). This means both RNGs start in identical state and produce identical pseudo-random sequences — the scheduler and sample-index selection will be perfectly correlated.

Compare with the perf phase in `RuntimeSettings.from_config` which uses separate fields `scheduler_random_seed` and `dataloader_random_seed`. Consider using `random_seed` for one and `random_seed ^ 0xDEAD` (or adding `warmup_scheduler_random_seed` to `WarmupConfig`) for the other, so their sequences diverge.

tests/unit/dataset_manager/test_salted_dataset.py
```diff
@@ -0,0 +1,218 @@
...
+        sd = SaltedDataset(inner)
+        result = sd.load_sample(0)
+        assert result == {"question": "what is 2+2?", "answer": "4"}
+
+    def test_empty_dict_is_unchanged(self):
+        inner = _make_loaded_dataset([{}])
+        sd = SaltedDataset(inner)
+        assert sd.load_sample(0) == {}
+
+    def test_non_dict_sample_is_returned_as_is(self):
```
arekay-nv, line 159: **[Review Council — Claude] low · testing**

`sd.data = inner.data` is redundant now that `SaltedDataset.data` is a property that delegates to `self._inner.data`. The assignment calls the property setter which sets `self._inner.data = inner.data` — a no-op since they reference the same list. This won't cause a test failure, but it misleads readers into thinking `SaltedDataset.data` is a plain attribute:

```python
inner.data = ["raw string sample"]  # override with non-dict
sd = SaltedDataset(inner)
sd.data = inner.data  # redundant — sd.data property already returns inner.data
```

Remove the `sd.data = inner.data` line.

arekay-nv (commented):
test

src/inference_endpoint/dataset_manager/dataset.py
arekay-nv: **[Review Council — Claude] medium · design**

`SaltedDataset` inherits from `Dataset` without opting out of `__init_subclass__` registration. `Dataset.__init_subclass__` registers every concrete non-abstract subclass under its class name in `Dataset.PREDEFINED`:

```python
cls.DATASET_ID = dataset_id  # → "SaltedDataset"
Dataset.PREDEFINED[dataset_id] = cls
```

`SaltedDataset.__init__` takes `inner: Dataset` (not `dataframe: pd.DataFrame`), so any code that instantiates via the registry will raise a `TypeError`. Fix: opt out explicitly:

```python
class SaltedDataset(Dataset, dataset_id=None):
```

src/inference_endpoint/dataset_manager/dataset.py
arekay-nv, line 463: **[Review Council — Claude] medium · bug**

When the prompt is a multimodal list where ALL parts are non-text (e.g., all `image_url`), the loop finds no text part and falls through to `return data` without salting and **without any warning**:

```python
if isinstance(prompt, list) and prompt:
    for i, part in enumerate(prompt):
        if isinstance(part, dict) and part.get("type") == "text":
            ...
            return {**data, "prompt": salted_parts}
return data  # unsupported prompt type — skip salting (silent!)
```

Unlike the `input_tokens` case (which emits `logger.warning`), this fallback is completely silent. Users who set `salt=True` for image-only datasets will see no indication that KV-cache bypass is not functioning. Add a `logger.warning` here, consistent with the `input_tokens` path.

src/inference_endpoint/commands/benchmark/execute.py
arekay-nv: **[Review Council — Claude] medium · design**

After the fix that seeds the warmup RNGs, both `rng_sched` and `rng_sample_index` will be seeded with the **same value** (`warmup_cfg.random_seed`, default `0`):

```python
rng_sched=random.Random(warmup_cfg.random_seed),
rng_sample_index=random.Random(warmup_cfg.random_seed),
```

Both RNGs start in identical state and produce the same sequence. Scheduler timing and sample selection will be perfectly correlated. Compare with the perf phase which uses separate `scheduler_random_seed` and `dataloader_random_seed` fields. Consider using `random_seed` and `random_seed ^ 0xDEAD` (or a separate `warmup_scheduler_random_seed` field) so the sequences diverge.

tests/unit/dataset_manager/test_salted_dataset.py
arekay-nv, line 159: **[Review Council — Claude] low · testing**

`sd.data = inner.data` is now redundant: `SaltedDataset.data` is a property that delegates to `self._inner.data`. The assignment calls the setter which sets `self._inner.data = inner.data` — a no-op since they already reference the same list.

```python
inner.data = ["raw string sample"]  # override with non-dict
sd = SaltedDataset(inner)
sd.data = inner.data  # redundant — property already returns inner.data
```

Remove the `sd.data = inner.data` line.

arekay-nv:

Reviewed by **Claude** (Codex CLI unavailable) | Depth: **thorough** | HEAD: \`1143d64\`

Found **7** new issues across **5** files. Inline comments posted for 4; 3 are out-of-diff or post-fix.

Cross-reviewer note: Issues already addressed in the previous round (unseeded RNGs, `SaltedDataset.data` snapshot, multimodal salting at index 0, inline imports, ruff-format) are not re-reported.

### 🔴 Must Fix (high)

| 1 | `commands/benchmark/execute.py` | 365 | design | **Inline ↑** After RNG-seed fix: `rng_sched` and `rng_sample_index` both seeded with same `random_seed` value — identical pseudo-random sequences for scheduler and sample selection |

### 🟡 Should Fix (medium)

| 2 | `dataset_manager/dataset.py` | 434 | design | **Inline ↑** `SaltedDataset` is registered in `Dataset.PREDEFINED` under `"SaltedDataset"` via `__init_subclass__`; instantiating via registry raises `TypeError` (incompatible constructor). Fix: `class SaltedDataset(Dataset, dataset_id=None)` |
| 3 | `dataset_manager/dataset.py` | 478 | bug | **Inline ↑** Silent no-salt fallback for image-only multimodal prompts (all parts are `image_url`, no text part exists) — no `logger.warning` emitted, `salt=True` silently no-ops |
| 4 | `load_generator/session.py` | 366 | bug | `_drain_inflight` awaits `self._drain_event.wait()` with no timeout. If a warmup request is dropped (network error, server restart), the drain hangs indefinitely and the benchmark never starts. Only `SIGINT` unblocks it. |

### 🔵 Consider (low)

| 5 | `tests/unit/dataset_manager/test_salted_dataset.py` | 143 | testing | **Inline ↑** `sd.data = inner.data` is redundant now that `data` is a property delegating to `_inner.data` |
| 6 | `config/schema.py` | 413 | api-contract | `random_seed: int = Field(0, ...)` default is implicitly deterministic; `Optional[int] = None` with "use fresh seed when None" semantics would better match `salt=True` cache-busting intent |
| 7 | `config/schema.py` | 395 | api-contract | `WarmupConfig` has no `@cyclopts.Parameter` annotations — warmup is YAML-only with no CLI flags. Undocumented limitation; users testing via CLI will find warmup silently inactive. |

🤖 Posted by `/review-council` skill.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -525,12 +554,31 @@ async def _run_benchmark_async(
...
+        if max_duration_ms is not None:
+
+            def _on_global_timeout() -> None:
+                logger.warning(
+                    "Global experiment timeout reached (%d ms); stopping session.",
+                    max_duration_ms,
+                )
+                session.stop()
+
+            global_timeout_handle = loop.call_later(
```
arekay-nv: **[Review Council — Both (Codex P1 + Claude)] high · api-contract**

The new global timer redefines what `max_duration_ms` means, but only when warmup is enabled, and the schema description was not updated.

```python
max_duration_ms = ctx.rt_settings.max_duration_ms
if max_duration_ms is not None:
    ...
    global_timeout_handle = loop.call_later(
        max_duration_ms / 1000.0, _on_global_timeout
    )
```

`session.py:_make_stop_check` still enforces `max_duration_ms` per-phase against `phase_start_ns`, so we now have **two clocks** running on the same configured value:

- per-phase, from each phase's start (the original semantic — and still what `_build_phases` clears to `None` for warmup/accuracy).
- global, from `loop.call_later` scheduling time (warmup + perf + accuracy combined).

For a user with `warmup.enabled=True` and `max_duration_ms=600_000` (10 min), if warmup takes 90s, the perf phase's effective wall-clock budget collapses from 10 min to ~8.5 min — silently. Worse, an accuracy pass in `TestMode.BOTH` (which sets `max_duration_ms=None` per-phase, expecting unbounded) can now be truncated mid-run by the global timer.

The schema description still reads `"Maximum test duration in ms (0 for no limit)"`, which is the original per-phase meaning. Existing YAML configs will get shorter perf/accuracy phases once they enable warmup, with no documentation indicating why.

Suggested fixes (any one):
1. Schedule the timer at the start of the **performance** phase (e.g., on `START_PERFORMANCE_TRACKING`) so warmup time doesn't eat the perf budget — preserves the per-phase semantic.
2. Expose a separate `warmup.max_duration_ms` to bound only warmup, leave perf-phase `max_duration_ms` untouched.
3. If a global cap is genuinely the goal, rename it (`experiment_timeout_ms`?), keep `max_duration_ms` per-phase, and document the change in `WarmupConfig` and release notes.

src/inference_endpoint/load_generator/session.py
```diff
@@ -363,9 +356,9 @@ async def _run_phase(self, phase: PhaseConfig) -> PhaseResult | None:
     async def _drain_inflight(self, phase_issuer: PhaseIssuer) -> None:
         """Wait for all in-flight responses from this phase to complete.
 
-        Currently, there is no timeout for the drain step. In the future,
-        we can possibly add a dynamic timeout based on the rate of completion
-        throughout the current phase."""
+        Bounded by the global experiment timeout: if the caller schedules a
+        loop.call_later that calls stop(), stop() sets _drain_event, unblocking
+        this wait without leaving it hung indefinitely."""
```
arekay-nv: **[Review Council — Claude] medium · error-handling**

The new docstring on `_drain_inflight` claims a safety guarantee that only holds in some configurations:

```python
async def _drain_inflight(self, phase_issuer: PhaseIssuer) -> None:
    """Wait for all in-flight responses from this phase to complete.

    Bounded by the global experiment timeout: if the caller schedules a
    loop.call_later that calls stop(), stop() sets _drain_event, unblocking
    this wait without leaving it hung indefinitely."""
    ...
    self._drain_event.clear()
    await self._drain_event.wait()
```

The `loop.call_later` is only scheduled in `execute.py:561` if `ctx.rt_settings.max_duration_ms is not None`. `runtime_settings.from_config` maps `max_duration_ms == 0` to `None`, and the schema default is `0` — so the **default** offline/online CLI run has no global timer, and `_drain_inflight` retains its old unbounded `await self._drain_event.wait()`. A single dropped or hanging request will hang the benchmark forever.

The docstring suggests this case is solved; it isn't.

Suggested fixes:
- Make the docstring conditional ("if a global timeout is configured…") and add a note that the drain is otherwise unbounded, OR
- Add an intrinsic drain timeout (a fixed cap like 30–60s, or a multiple of observed completion latency) inside `_drain_inflight` itself so the safety property is a property of the function, not a contract on the caller.

arekay-nv (commented):
## Review Council — Multi-AI Code Review (Round 2)

Reviewed by **Codex** + **Claude** | Depth: **thorough** | HEAD: `c9e80b0` (was `56258be` in round 1)

Re-reviewing after the author's response commits. Most round-1 findings are addressed; this round focuses on **new code** (global timeout, `warmup_random_seed`, `register=False`) and incomplete fixes.

Found **4** new issues across **3** files: 0 critical, 1 high, 1 medium, 2 low. See inline comments — a tiered summary will follow.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -525,12 +554,31 @@ async def _run_benchmark_async(
...
+        global_timeout_handle = None
+        max_duration_ms = ctx.rt_settings.max_duration_ms
+        if max_duration_ms is not None:
+
+            def _on_global_timeout() -> None:
+                logger.warning(
+                    "Global experiment timeout reached (%d ms); stopping session.",
+                    max_duration_ms,
+                )
+                session.stop()
```
arekay-nv, line 729: **[Review Council — Claude] low · concurrency**

The global timer's callback calls `session.stop()` directly, with no guard against firing after cleanup has begun:

```python
def _on_global_timeout() -> None:
    logger.warning(
        "Global experiment timeout reached (%d ms); stopping session.",
        max_duration_ms,
    )
    session.stop()
```

`global_timeout_handle.cancel()` only runs **after** entering the `finally:` block at line 580, so the timer can fire during the gap between `await session.run(phases)` returning and the `cancel()` call running — or worse, during the slow cleanup that follows (`http_client.shutdown_async()`, publisher close, etc.). When that happens, `session.stop()` flips `_stop_requested=True` and `_drain_event.set()` on a session that has already finished. The downstream effect is mostly harmless today (the strategy task is `None`), but the race is real and easy to make harmful with future refactors.

Suggested fix:
```python
def _on_global_timeout() -> None:
    if session._done:  # or expose a public flag
        return
    logger.warning(...)
    session.stop()
```

Also consider moving `global_timeout_handle.cancel()` to immediately after `session.run(phases)` returns (success path), and only run it in `finally` to handle the exception path.

tests/unit/commands/test_benchmark.py
```diff
@@ -385,14 +399,390 @@ def test_valid_templates_parse(self, template):
...
+
+        order_with = [
+            next(create_sample_order(perf_with.runtime_settings)) for _ in range(n_draw)
+        ]
+        order_without = [
+            next(create_sample_order(perf_without.runtime_settings))
+            for _ in range(n_draw)
+        ]
+
+        assert order_with == order_without, (
```
arekay-nv, line 775: **[Review Council — Claude] low · testing**

`test_performance_sample_order_identical_with_and_without_warmup` does not actually exercise the property it claims to test:

```python
def make_rt():
    return RuntimeSettings(
        ...
        rng_sched=random.Random(99),
        rng_sample_index=random.Random(99),
        ...
    )

ctx_with = self._make_ctx(config_with, make_rt(), simple_dataset)
ctx_without = self._make_ctx(config_without, make_rt(), simple_dataset)
```

`make_rt()` is called twice and produces two **separate** `RuntimeSettings` instances with freshly-seeded `Random(99)` RNGs. `_build_phases` in the warmup path uses `dataclass_replace` which substitutes new `Random(seed)` instances on the warmup phase only — it never touches the perf phase's RNGs. So this test only verifies that `_build_phases` doesn't drain the perf RNG, which it never could (it only stores references). It does not verify the regression it claims to guard against: that **executing** the warmup phase leaves the perf-phase RNG untouched.

A tighter test would either:
1. Build phases from a **single shared** `RuntimeSettings`, then advance the warmup phase's `rng_sample_index` (e.g., draw N samples from `create_sample_order(warmup_rt)`), and assert that the perf-phase order matches a control with no warmup, OR
2. Drop this test as redundant with `test_warmup_phase_uses_independent_rngs` (which already asserts `warmup_rt.rng_sched is not perf_rt.rng_sched`). The current implementation gives a false sense of coverage for the actual race the docstring describes.

arekay-nv:
## Review Council — Multi-AI Code Review (Round 2)

Reviewed by **Codex** + **Claude** | Depth: **thorough** | HEAD: `c9e80b0` (was `56258be` in [round 1](https://github.com/mlcommons/endpoints/pull/305#pullrequestreview-4225024755))

Re-review after the author's response commits. **4** new issues across **3** files. Inline comments in [this review](https://github.com/mlcommons/endpoints/pull/305#pullrequestreview-4229521878).

Cross-reviewer agreement (boosted confidence): **execute.py:570** — global timeout semantic change — was the only finding **both** Codex and Claude flagged in round 2.

### 🔴 Must Fix (high)

| 1 | `src/inference_endpoint/commands/benchmark/execute.py` | 570 | api-contract | Both | Global timer makes warmup eat into the per-phase `max_duration_ms` budget; `_make_stop_check` still uses per-phase semantics → two clocks on the same configured value, accuracy phases configured for unbounded duration can be truncated |

### 🟡 Should Fix (medium)

| 1 | `src/inference_endpoint/load_generator/session.py` | 361 | error-handling | Claude | `_drain_inflight` docstring claims a safety guarantee that only holds when `max_duration_ms is not None`; the **default** offline/online run still has unbounded drain on a hung request |

### 🔵 Consider (low)

| 1 | `src/inference_endpoint/commands/benchmark/execute.py` | 568 | concurrency | Claude | `_on_global_timeout` has no `_done` guard; can fire during the cleanup gap between `session.run()` returning and `global_timeout_handle.cancel()` running |
| 2 | `tests/unit/commands/test_benchmark.py` | 775 | testing | Claude | `test_performance_sample_order_identical_with_and_without_warmup` builds two **separate** RuntimeSettings with fresh seeded RNGs — it doesn't actually exercise warmup-perturbing-perf-RNG; gives a false sense of coverage |


---

### Round 1 follow-up status

- ✅ Fixed — `execute.py:365 — warmup RNGs unseeded` — now `random.Random(warmup_random_seed)`
- ✅ Fixed — `dataset.py:467 — multimodal salt only first text` — now scans for first text part at any index
- ✅ Fixed — `dataset.py:444 — `SaltedDataset.data` snapshot` — now `@property` delegating to `_inner.data`
- ✅ Fixed — `test_benchmark.py:471 — lazy imports` — imports moved to top
- ✅ Fixed — `dataset.py:434 — registry pollution` — now `class SaltedDataset(Dataset, register=False)`
- 🟡 Partial — `dataset.py:459 — `input_tokens` not salted` — warning added for `input_tokens`-without-`prompt`; common case (both present) still silently drops salt
- 🟡 Partial — `execute.py:357 — ad-hoc warmup_rt` — now uses `dataclass_replace` ✅ but `LoadPattern(MAX_THROUGHPUT)` is still hardcoded
- ⏸️ Unaddressed — `dataset.py:414 — `datasets/` → `dataset_cache/`` — still a silent breaking default change
- ⏸️ Unaddressed — `session.py:273 — `ENABLE_WARMUP` env-var removed` — no deprecation warning
- ⏸️ Unaddressed — `session.py:68 — `drain_after=True` default flips warmup` — direct callers get opposite semantics
- ⏸️ Unaddressed — `session.py:423 — late warmup `COMPLETE` events still publish to ZMQ`
- ⏸️ Unaddressed — `metrics_aggregator/__main__.py:100 — tokenizer fail silently downgrades`
- ⏸️ Unaddressed — `dataset.py:464 — `os.urandom` salt ignores seeded RNG`
- ⏸️ Unaddressed — `test_warmup.py:292 — `_DELAY=0.15` flake risk`
- ⏸️ Unaddressed — `test_warmup.py:65 — no multi-worker tests`

---

### Theme summary (round 2)

The biggest new concern is the **global wall-clock timer** added to bound the warmup drain. The mechanism works, but it silently changes what `max_duration_ms` means for any user who enables warmup (warmup time now eats into the perf-phase budget, and accuracy phases configured for `None` get truncated by the global cap). Either scope the timer to the perf phase, expose a separate `warmup.max_duration_ms`, or rename the field — but don't double-mean the same configured value.

The drain-safety claim in `_drain_inflight` is also overstated: with the schema default (`max_duration_ms=0` → `None`), no global timer is scheduled and the drain is unbounded — a single hung request can still hang the benchmark.

🤖 Posted by `/review-council` skill.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -525,12 +554,31 @@ async def _run_benchmark_async(
         phases = _build_phases(ctx)
         report: Report | None = None
 
+        # Global wall-clock timeout covers warmup + performance + accuracy phases
```
nv-alicheng: **[Review Council — Codex P1] high · bug**

Global `max_duration_ms` timer covers warmup + perf + accuracy combined, starting before the warmup phase. With `settings.warmup.enabled=true`, a slow warmup eats into the perf phase's budget and can truncate or skip the measured phase entirely — even though `_build_phases()` explicitly gives warmup `max_duration_ms=None`. Existing configs with `max_duration_ms=N` previously meant 'perf runs up to N'; they now mean 'whole session up to N'. Silent semantic shift for benchmark results.

Suggest: arm the timer after warmup completes, or carve out the warmup duration from the perf budget.

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
```diff
@@ -91,10 +92,16 @@ async def main() -> None:
 
     # Using ternary operator causes errors in MyPy object type coalescing
     # (coalesces to 'object' not 'AbstractContextManager[TokenizePool | None]')
+    pool_cm: AbstractContextManager[TokenizePool | None]
     if args.tokenizer:
-        pool_cm: AbstractContextManager[TokenizePool | None] = TokenizePool(
-            args.tokenizer, n_workers=args.tokenizer_workers
-        )
+        try:
+            pool_cm = TokenizePool(args.tokenizer, n_workers=args.tokenizer_workers)
```
nv-alicheng: **[Review Council — Claude] medium · bug**

If `TokenizePool.__init__` fails after the `ThreadPoolExecutor` is created (e.g. a `_get_thread_tokenizer` task raises on a bad tokenizer name or network error), the partially-constructed pool is unreachable but the executor's worker threads keep running — the `except Exception` at line 99 catches the failure and replaces `pool_cm` with `nullcontext()` without calling `pool.close()`. Each retry/restart leaks more daemon threads. Wrap the constructor in a try/except inside `TokenizePool.__init__` that shuts down the executor on failure, or close the pool from the outer except handler before falling back to nullcontext.

src/inference_endpoint/testing/echo_server.py
```diff
@@ -31,6 +33,8 @@
 from inference_endpoint.openai.openai_types_gen import CreateChatCompletionRequest
 from inference_endpoint.utils.logging import setup_logging
 
+RequestHandler = Callable[[web.Request], web.Response | Awaitable[web.Response]]
```
nv-alicheng, line 36: **[Review Council — Claude] low · api-contract**

`RequestHandler = Callable[[web.Request], web.Response | Awaitable[web.Response]]` doesn't allow `None`, but `_dispatch` (line 110) explicitly relies on `None` to fall through to the default echo handler. A user-provided handler that forgets to `return` therefore silently passes through default behavior instead of failing loudly — and the type alias misleads them about what's supported. Either widen the alias to `... | None`, or assert non-None inside `_dispatch` so the error surfaces.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -347,6 +349,33 @@ def _build_phases(ctx: BenchmarkContext) -> list[PhaseConfig]:
...
+        warmup_rt = dataclass_replace(
+            ctx.rt_settings,
+            min_duration_ms=0,
+            max_duration_ms=None,
+            n_samples_from_dataset=ctx.dataloader.num_samples(),
+            n_samples_to_issue=warmup_cfg.n_requests,
+            min_sample_count=1,
+            rng_sched=random.Random(warmup_cfg.warmup_random_seed),
+            rng_sample_index=random.Random(warmup_cfg.warmup_random_seed + 1),
+            load_pattern=LoadPattern(type=LoadPatternType.MAX_THROUGHPUT),
```
nv-alicheng: **[Review Council — Claude] medium · design**

Warmup hardcodes `LoadPattern(type=LoadPatternType.MAX_THROUGHPUT)` regardless of the perf-phase load pattern. For online (Poisson/concurrency) benchmarks this is surprising: a user enabling warmup for an online run gets a synchronous burst of `n_requests` queries firing simultaneously, exercising a different server code path (queue blow-up, batching) than the steady-state pattern they're trying to warm up for. (Distinct from the existing comment on line 357, which is about code duplication; this is a correctness concern for online runs.) Document the constraint in `WarmupConfig.salt`'s docstring or expose the warmup load pattern (default to perf's pattern) so online warmup matches the workload shape.

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -431,6 +432,74 @@ def get_dataloader(
...
+        self._inner.data = value
+
+    def load(
+        self,
+        adapter: "HttpRequestAdapter | None" = None,
+        api_type: APIType | None = None,
+        model_params: ModelParams | None = None,
+        force: bool = False,
+    ) -> None:
+        pass  # Inner dataset already loaded
```
nv-alicheng: **[Review Council — Claude] low · design**

`SaltedDataset.load()` is a no-op (`pass  # Inner dataset already loaded`). This silently ignores `force=True` and assumes the inner is already loaded. If a caller constructs `SaltedDataset(unloaded_inner)` and calls `load()` (the documented entry point on `Dataset`), it returns without error and the first `load_sample(0)` blows up inside `Dataset.load_sample` with `AssertionError: Dataset not loaded`. Either delegate to `self._inner.load(adapter, api_type, model_params, force)` or raise in `__init__` if `inner.data is None`.

tests/unit/commands/test_benchmark.py
```diff
@@ -385,14 +399,390 @@ def test_valid_templates_parse(self, template):
...
+class TestWarmupConfig:
+    """Tests for WarmupConfig schema model."""
+
+    @pytest.mark.unit
+    def test_defaults(self):
+        cfg = WarmupConfig()
+        assert cfg.enabled is False
+        assert cfg.n_requests is None
+        assert cfg.salt is False
+        assert cfg.drain is False
```
nv-alicheng, line 411: **[Review Council — Claude] low · testing**

`test_defaults` asserts `enabled`, `n_requests`, `salt`, `drain` defaults but never checks `warmup_random_seed` (which has a default of 42 in `WarmupConfig` and is plumbed through `execute.py:365` to seed `rng_sched`/`rng_sample_index`). Same gap exists in `test_all_flags_enabled` (line 437) and `test_yaml_roundtrip` (line 445). If a future change silently flips the seed default (e.g. to a non-deterministic source), no test catches it.

tests/unit/dataset_manager/test_salted_dataset.py
```diff
@@ -0,0 +1,218 @@
...
+        # With 8 bytes of randomness we expect all 20 samples to be distinct
+        assert len(salts) == 20
+
+    def test_salt_unique_across_different_indices(self):
+        inner = _make_loaded_dataset([{"prompt": "a"}, {"prompt": "b"}])
+        sd = SaltedDataset(inner)
+        prompt0 = sd.load_sample(0)["prompt"]
+        prompt1 = sd.load_sample(1)["prompt"]
+        # Salts should differ (original prompts differ, salts are random)
+        assert prompt0 != prompt1
```
nv-alicheng: **[Review Council — Claude] low · testing**

`test_salt_unique_across_different_indices` asserts `prompt0 != prompt1` after salting two underlying prompts `"a"` and `"b"`. But the prompts already differ before salting, so the assertion would still pass even if salting were entirely broken (e.g. always emitting an empty salt). Either use the same underlying prompt under different indices, or assert that the salt prefixes (not the full prompt) differ.

tests/unit/commands/test_benchmark.py
```diff
@@ -385,14 +399,390 @@ def test_valid_templates_parse(self, template):
...
+            **_OFFLINE_KWARGS,
+            settings=OfflineSettings(warmup=WarmupConfig(enabled=True)),
+        )
+        ctx = self._make_ctx(config, base_rt_settings, simple_dataset)
+        phases = _build_phases(ctx)
+
+        warmup_rt = phases[0].runtime_settings
+        perf_rt = phases[1].runtime_settings
+        assert warmup_rt.rng_sched is not perf_rt.rng_sched
+        assert warmup_rt.rng_sample_index is not perf_rt.rng_sample_index
```
nv-alicheng, line 721: **[Review Council — Claude] low · testing**

`test_warmup_uses_independent_rng_instances` only checks `is not` identity. This passes even if both warmup RNGs were seeded with the same seed as perf — i.e. distinct objects but identical sequences, which would still cause warmup to consume the same state pattern as perf. Add an assertion comparing the first few drawn values, e.g. `assert [warmup_rt.rng_sched.random() for _ in range(5)] != [perf_rt.rng_sched.random() for _ in range(5)]`, to actually catch the regression the test claims to guard.

tests/unit/dataset_manager/test_salted_dataset.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def test_category_field_preserved(self, loaded_inner):
+        sd = SaltedDataset(loaded_inner)
+        assert sd.load_sample(0)["category"] == "general"
+        assert sd.load_sample(1)["category"] == "ml"
+
+    def test_all_salts_unique_across_full_dataset(self, loaded_inner):
+        sd = SaltedDataset(loaded_inner)
+        prompts = [sd.load_sample(i)["prompt"] for i in range(sd.num_samples())]
+        assert len(set(prompts)) == len(prompts)
```
nv-alicheng, line 251: **[Review Council — Claude] low · testing**

`SaltedDataset.load_sample` has a non-trivial branch at dataset.py:471 that warns when `input_tokens` is present without `prompt`, and a salt-the-prompt branch when both are present (which still leaves `input_tokens` unsalted — see existing comment on dataset.py:467). Neither path is exercised here. Add tests for (a) `{"input_tokens": [...], "prompt": "..."}` (salting still happens but input_tokens unchanged — silent miss for token-based adapters), and (b) `{"input_tokens": [...]}` only (warning fires, data returned unchanged).

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -431,6 +432,74 @@ def get_dataloader(
...
+        if "input_tokens" in data and "prompt" not in data:
+            self.logger.warning(
+                "SaltedDataset: sample has 'input_tokens' but no 'prompt' — "
+                "salt cannot be applied to pre-tokenized input; KV-cache reuse may not be prevented"
+            )
+            return data
+        if "prompt" not in data:
+            return data
+        prompt = data["prompt"]
+        salt = os.urandom(8).hex()
```
nv-alicheng: **[Review Council — Claude] low · data-integrity**

(Distinct from the existing low/design comment on this line about determinism.) The salt format `f"[{salt}] {prompt}"` adds a fixed 18-character (~5-token) prefix to every salted prompt. For users tracking ISL distributions across warmup vs perf — e.g. validating that warmup matches the workload's prompt-shape — this introduces a systematic +5..+10 token bias on every salted warmup sample. For short prompts (<20 tokens) this is a notable distortion. Worth documenting in `WarmupConfig.salt`'s description so users tuning prompt length aren't surprised.

src/inference_endpoint/config/schema.py
```diff
@@ -401,6 +425,7 @@ class Settings(BaseModel):
     runtime: RuntimeConfig = Field(default_factory=RuntimeConfig)
     load_pattern: LoadPattern = Field(default_factory=LoadPattern)
     client: HTTPClientConfig = Field(default_factory=HTTPClientConfig)
+    warmup: WarmupConfig = Field(default_factory=WarmupConfig)
```
nv-alicheng, line 482: **[Review Council — Claude] low · api-contract**

`Settings.warmup: WarmupConfig` has no `cyclopts.Parameter` annotation, so cyclopts auto-generates only dotted CLI flags like `--settings.warmup.enabled=true`, `--settings.warmup.n-requests=...`. Compare with `runtime` and `load_pattern` which expose ergonomic shorthands via field-level `cyclopts.Parameter(alias=...)`. As-is, enabling warmup from the CLI is unwieldy. Add `cyclopts.Parameter(alias="--warmup-...")` aliases on the `WarmupConfig` fields, or document that warmup is YAML-only.

nv-alicheng (commented):
## Review Council — Multi-AI Code Review (run #2)

Found **12** new issues (deduped against the 31+ existing inline comments). Severity: **1** high, **2** medium, **9** low.

See the inline comments for details. A summary table is posted as a separate top-level comment after this review.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -347,6 +349,33 @@ def _build_phases(ctx: BenchmarkContext) -> list[PhaseConfig]:
...
 
+    # Warmup phase (optional, before performance)
+    warmup_cfg = ctx.config.settings.warmup
+    if warmup_cfg.enabled:
+        warmup_dataset: Dataset = (
+            SaltedDataset(ctx.dataloader) if warmup_cfg.salt else ctx.dataloader
+        )
+        warmup_rt = dataclass_replace(
+            ctx.rt_settings,
+            min_duration_ms=0,
```
nv-alicheng, line 453: **[Review Council — Claude] low · bug**

`min_duration_ms=0` combined with `n_samples_to_issue=warmup_cfg.n_requests` (which may be `None`) makes the warmup phase fall back to the implicit 'all dataset samples' path in `total_samples_to_issue()`. That's `dataloader.num_samples()` queries — potentially thousands — issued at MAX_THROUGHPUT in a single burst. There is no documented cap on warmup duration when `n_requests is None`, and the user's intent ('warm up before perf') is silently expanded to 'issue every sample once'. Either clamp warmup's effective sample count, or document this default explicitly in `WarmupConfig.n_requests`'s description.

nv-alicheng:
## Review Council — Multi-AI Code Review (run #2)

Found **12** new issues across **6** files (deduped against 31 existing inline comments).

### 🔴 Must Fix (high)

Issues that produce incorrect benchmark results in normal usage.

| 1 | `src/inference_endpoint/commands/benchmark/execute.py` | 557 | bug | Codex | Global `max_duration_ms` timer covers warmup + perf + accuracy combined; slow warmup eats into perf budget. Silent semantic shift from previous "perf-only" meaning. |

### 🟡 Should Fix (medium)

Real issues that trigger under specific conditions or design flaws that compound.

| 2 | `src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py` | 98 | bug | Claude | `TokenizePool` partial-init failure leaks the executor's worker threads on retry. |
| 3 | `src/inference_endpoint/commands/benchmark/execute.py` | 367 | design | Claude | Warmup hardcoded to `MAX_THROUGHPUT` regardless of perf load pattern — surprising for online (Poisson/concurrency) runs since warmup exercises a different server code path than the steady-state pattern it's meant to warm up for. |

### 🔵 Consider (low)

Valid improvements; could be follow-ups.

| 4 | `src/inference_endpoint/testing/echo_server.py` | 36 | api-contract | Claude | `RequestHandler` type alias doesn't allow `None`, but `_dispatch` relies on `None` for fall-through. Either widen the type or assert non-None. |
| 5 | `src/inference_endpoint/dataset_manager/dataset.py` | 465 | design | Claude | `SaltedDataset.load()` is a no-op — unloaded inner causes opaque `AssertionError` later. Either delegate or raise in `__init__`. |
| 6 | `src/inference_endpoint/dataset_manager/dataset.py` | 480 | data-integrity | Claude | Salt prefix `"[{salt}] "` adds ~5 token overhead — distorts ISL distribution between warmup and perf, especially for short prompts. |
| 7 | `src/inference_endpoint/config/schema.py` | 428 | api-contract | Claude | `WarmupConfig` lacks `cyclopts.Parameter` aliases — only verbose `--settings.warmup.*` flags, unlike sibling `runtime`/`load_pattern`. |
| 8 | `src/inference_endpoint/commands/benchmark/execute.py` | 360 | bug | Claude | `min_duration_ms=0` + `n_requests=None` falls back to "issue every dataset sample at MAX_THROUGHPUT" — undocumented expansion of warmup's intent. |
| 9 | `tests/unit/commands/test_benchmark.py` | 411 | testing | Claude | `test_defaults` doesn't assert the `warmup_random_seed=42` default; same gap in `test_all_flags_enabled` and `test_yaml_roundtrip`. |
| 10 | `tests/unit/commands/test_benchmark.py` | 721 | testing | Claude | `test_warmup_uses_independent_rng_instances` only checks `is not` identity; passes even if both RNGs were seeded identically. |
| 11 | `tests/unit/dataset_manager/test_salted_dataset.py` | 109 | testing | Claude | `test_salt_unique_across_different_indices` compares already-different prompts; passes even if salting is broken. |
| 12 | `tests/unit/dataset_manager/test_salted_dataset.py` | 218 | testing | Claude | No tests exercise the `input_tokens`-present branches of `SaltedDataset.load_sample` (the ones that produce silent token-ID misses for SGLang). |

---

### What we deduped

- **Codex P2** (`SaltedDataset` doesn't salt `input_tokens`) — already covered by [existing comment on `dataset.py:467`](https://github.com/mlcommons/endpoints/pull/305#discussion_r2480091872) (`[Review Council — Claude] high · bug`). Dropped.
- **Claude `test_warmup.py:433`** (`max_concurrent > 5` flakiness) — already covered by [existing comment on `test_warmup.py:292`](https://github.com/mlcommons/endpoints/pull/305) (`[Review Council — Claude] low · testing`), which names that exact test. Dropped.

### Notes on Codex coverage this run

Codex returned only 2 findings on this run (both correctness-grade). The first reproduced an existing high-severity finding (input_tokens salting); the second (the `max_duration_ms` timing bug) is genuinely new and is the most impactful issue in this review. The remaining coverage came from Claude.

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -431,6 +432,74 @@ def get_dataloader(
         return cls(df, transforms=transforms, repeats=num_repeats)
 
 
+class SaltedDataset(Dataset, register=False):
```
nv-alicheng: @arekay-nv Can you clarify the purpose and usage of this? Conceptually a salted dataset is just a normal dataset, but on load_sample, it prepends a randomly generated fixed length salt right? I'm wondering if adding a `salt: bool = False` argument to the base class `.load_sample` function is warranted. A lot of the body of this class is just remaps to the underlying dataset.

src/inference_endpoint/load_generator/session.py
nv-alicheng: I think this is fine to ignore - ENABLE_WARMUP was purely for developers and not used by anything in production (CICD, manual testing / benchmarks, etc).

src/inference_endpoint/dataset_manager/dataset.py
arekay-nv: Moving it will clean up local pre-commit tests which fail due to the presence of `datasets` folder.

src/inference_endpoint/dataset_manager/dataset.py
arekay-nv: I like that approach. Updated. Thanks!

Victor49152:
Tested on example_08/q3vl and the warmup worked as expected and reduced starting overhead nicely.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -438,6 +440,35 @@ def _build_phases(
     """Build the phase list from BenchmarkContext."""
     phases: list[PhaseConfig] = []
 
+    # Warmup phase (optional, before performance)
+    warmup_cfg = ctx.config.settings.warmup
+    if warmup_cfg.enabled:
+        warmup_dataset: Dataset = (
+            ctx.dataloader.with_salt(random.Random(warmup_cfg.warmup_random_seed + 2))
```
viraatc, line 447: qq: whats the +2 here for?

EDIT: i see its to reuse warmup_random_seed

src/inference_endpoint/commands/benchmark/execute.py
viraatc: is `nonlocal session` not needed? 
do we have a test to hit this?

src/inference_endpoint/load_generator/session.py
```diff
@@ -520,26 +525,31 @@ def _handle_response(self, resp: QueryResult | StreamChunk) -> None:
...
-                    sample_uuid=query_id,
-                    conversation_id=conv_id_str,
-                    turn=turn_num,
-                    data=resp.response_output,
+            if self._current_phase_type != PhaseType.WARMUP:
+                self._publisher.publish(
+                    EventRecord(
+                        event_type=SampleEventType.COMPLETE,
+                        timestamp_ns=resp.completed_at
+                        if isinstance(resp.completed_at, int)
```
viraatc, line 533: is the else branch ever actually hit? (resp.completed_at non int)?

src/inference_endpoint/commands/benchmark/execute.py
viraatc, line 447: should we warn if we turn on warmup and no performance dataset is specified for dataloader?

viraatc (approved):
looks great, ty!

src/inference_endpoint/load_generator/session.py
arekay-nv: The else branch might just be for mypy since the `completed_at` can be none or int.

src/inference_endpoint/commands/benchmark/execute.py
arekay-nv: Will address in followup - will probably hit more edge cases.

arekay-nv:
Merging since most unresolved comments are ranked low. Will address remaining in followup once we start collecting numbers with warmup.

Issue #11, opened 2025-11-04
Documentation: Add example of how to ramp up vLLM/SGLang endpoints for benchmarking


nvzhihanj:
As titled.

We can also add this as an example

Issue #10, opened 2025-11-04
Perf: system bottleneck tests


nvzhihanj:
In order to have confidence in the measurement results, we need to know the system we are running client tests on are free of bottlenecks.

Implement a set of tests that can help us catch any issues ahead of launch time.

PR #306, opened 2026-05-05
Change MetricsAggregatorService to publisher to support live metrics without shmem

nv-alicheng:
## What does this PR do?

`shmem` implementation of KVStore in MetricsAggregatorService causes issues on ARM. Several solutions exist:
1. Rewrite shmem implementation in C/C++ where memory fencing primitives are exposed.
2. Restructure the design of the MetricsAggregatorService to be a ZMQ Publisher which publishes metrics at a fixed rate, which other processes, such as a TUI, can subscribe to.

This PR implements (2).

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactor/cleanup

## Testing

- [x] Tests added/updated
- [x] All tests pass locally

## Checklist

- [x] Pre-commit hooks pass


arekay-nv (commented):
_(superseded — see [#pullrequestreview-4237753512](https://github.com/mlcommons/endpoints/pull/306#pullrequestreview-4237753512) for the full review-council output)_

arekay-nv (commented):

Found **12** issues across **7** files (excluding tests, pyproject.toml, uv.lock).

> ⚠️ GitHub's review-comment API is currently returning a persistent `pull_request_review_thread.base` internal error on this PR for inline comments via `line`+`side` and `position`-based posts (the github-code-quality and gemini-code-assist bot comments from yesterday were unaffected). Falling back to a single review with file:line links — clicking each link opens the PR diff at the cited line.

Note: the existing gemini comments on aggregator.py:286/342/353/354 already cover the *encapsulation* concern around `table._in_flight_tasks` direct access, so that issue is intentionally not duplicated here. The cancel-without-await issue (#4 below) is a separate concurrency concern at the same file:line.

Issues that will cause incorrect behavior, data loss, or security problems in production.

| # | File:Line | Reviewer | Category | Summary |
|---|-----------|----------|----------|---------|
| 1 | [`src/inference_endpoint/commands/benchmark/execute.py:482`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/commands/benchmark/execute.py#L482) | Claude | data-integrity | Subscriber late-binding can drop initial ticks (ZMQ slow-joiner) |
| 2 | [`src/inference_endpoint/async_utils/services/metrics_aggregator/publisher.py:168`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/publisher.py#L168) | Claude | performance | `_write_atomic_fallback` runs blocking I/O on the event loop |
| 3 | [`src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py:208`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py#L208) | Codex | performance | Unbounded raw-sample retention in `SeriesSampler` |

### 🟡 Should Fix (medium)

Real issues that trigger under specific conditions or design flaws that will compound.

| # | File:Line | Reviewer | Category | Summary |
|---|-----------|----------|----------|---------|
| 4 | [`src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py:353`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py#L353) | Claude | error-handling | Cancellations not awaited before reading `n_pending` |
| 5 | [`src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py:161`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py#L161) | Claude | bug | `HdrHistogram(low, high, sig_figs)` constructed without `high >= 2*low` check |
| 6 | [`src/inference_endpoint/async_utils/services/metrics_aggregator/subscriber.py:55`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/subscriber.py#L55) | Claude | data-integrity | `conflate=True` default for the Report consumer is fragile |
| 7 | [`src/inference_endpoint/commands/benchmark/execute.py:423`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/commands/benchmark/execute.py#L423) | Codex | data-integrity | Stale `final_snapshot.msgpack` not cleared on report-dir reuse |

### 🔵 Consider (low)

| # | File:Line | Reviewer | Category | Summary |
|---|-----------|----------|----------|---------|
| 8 | [`src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py:281`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py#L281) | Claude | bug | Double `STARTED` reassigns and orphans the tick task |
| 9 | [`src/inference_endpoint/async_utils/services/metrics_aggregator/publisher.py:195`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/publisher.py#L195) | Claude | concurrency | `close()` cancels tick task but doesn't await it |
| 10 | [`src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py:166`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py#L166) | Claude | error-handling | No top-level exception handling around `run_until_complete(main())` |
| 11 | [`src/inference_endpoint/load_generator/session.py:408`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/load_generator/session.py#L408) | Claude | design | ERROR-before-COMPLETE relies on an undocumented publisher ordering contract |
| 12 | [`src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py:132`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py#L132) | Claude | bug | SIGTERM bypasses `publish_final` |

---

## Detailed findings

### 1. 🔴 high · `data-integrity` · [Claude]

[`src/inference_endpoint/commands/benchmark/execute.py:482`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/commands/benchmark/execute.py#L482)

**Subscriber late-binding can drop initial ticks (ZMQ slow-joiner)**

`MetricsSnapshotSubscriber` is constructed (482) and `start()`ed (485) AFTER `launcher.launch()` waits for aggregator readiness. The aggregator can begin publishing as soon as STARTED is observed; ZMQ slow-joiner means snapshots emitted before the SUB handshake completes are dropped. The comment on line 480 acknowledges this. On a busy host the subscriber may also miss COMPLETE if the publisher closes before the subscription warms up, silently degrading to LIVE/DRAINING (incomplete report). **Fix:** move subscriber construction + `start()` BEFORE `launcher.launch()`. Connecting to a not-yet-bound IPC path is fine — ZMQ retries connect transparently.

### 2. 🔴 high · `performance` · [Claude]

[`src/inference_endpoint/async_utils/services/metrics_aggregator/publisher.py:168`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/publisher.py#L168)

**`_write_atomic_fallback` runs blocking I/O on the event loop**

`f.write` + `f.flush` + two `os.fsync` (file + parent dir) + `os.rename` execute synchronously on the aggregator subprocess's main event loop (awaited via `publish_final` from aggregator.py:362). On a busy box `os.fsync` can block tens-to-hundreds of ms — long enough to back-pressure event-record processing if any events are still in flight at finalization. **Fix:** wrap with `await asyncio.to_thread(self._write_atomic_fallback, payload)`.

### 3. 🔴 high · `performance` · [Codex]

[`src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py:208`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py#L208)

**Unbounded raw-sample retention in `SeriesSampler`**

`self._raw.append(value)` retains every observed value in an `array.array` for the full run so the final snapshot can recompute exact percentiles. The prior mmap-backed store spilled this to disk; the new path keeps it in the aggregator subprocess RAM, so memory now scales linearly with run length. A 5-min 50k-QPS run produces ~15M samples (~120MB per int64 series); once latency/ISL/OSL/TTFT/TPOT are all tracked, several hundred MB; longer runs OOM or swap heavily before the report is emitted. **Fix:** periodically spill `_raw` to disk, cap with reservoir sampling, or rely solely on the HDR digest for percentiles.

### 4. 🟡 medium · `error-handling` · [Claude]

[`src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py:353`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py#L353)

**Cancellations not awaited before reading `n_pending`**

After the drain timeout, `t.cancel()` only schedules cancellation at the next await point. `n_pending = len(table._in_flight_tasks)` on the next line therefore reads a count that's still high, and the not-yet-cancelled tasks may still be running when `_publisher.close()` and the loop teardown happen → "Task was destroyed but it is pending\!" warnings on shutdown. **Fix:** after `t.cancel()`, do `await asyncio.gather(*tasks, return_exceptions=True)` (bounded by a short timeout) before recomputing `n_pending`. Independent of the encapsulation point gemini already raised.

### 5. 🟡 medium · `bug` · [Claude]

[`src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py:161`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py#L161)

**`HdrHistogram(low, high, sig_figs)` constructed without `high >= 2*low` check**

`hdrh` requires `high >= 2*low`. Current bounds (1ns–3.6e12ns, 1–1e7 tokens) satisfy this, but a future caller passing e.g. `hdr_low=1, hdr_high=1` (or any `hdr_high < 2*hdr_low`) gets an opaque `ValueError` from deep inside the C library. **Fix:** explicit pre-check after the clamps: `if self._hdr_high < self._hdr_low * 2: raise ValueError(...)`.

### 6. 🟡 medium · `data-integrity` · [Claude]

[`src/inference_endpoint/async_utils/services/metrics_aggregator/subscriber.py:55`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/subscriber.py#L55)

**`conflate=True` default for the Report consumer is fragile**

The docstring argues this is safe because COMPLETE is the publisher's last message, but ZMQ CONFLATE only keeps the most recent unread message. If the subscriber's loop is starved (main process busy in `wait_for_exit`) and the publisher closes before the SUB sees COMPLETE (LINGER expires, IPC socket unlinked), `complete` never fires and `latest` may be a stale LIVE. The 2 s `wait_for_complete` timeout in execute.py:548 papers over most cases, but consider `conflate=False` for the Report consumer (a single producer at a few snapshots/sec, not a TUI).

### 7. 🟡 medium · `data-integrity` · [Codex]

[`src/inference_endpoint/commands/benchmark/execute.py:423`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/commands/benchmark/execute.py#L423)

**Stale `final_snapshot.msgpack` not cleared on report-dir reuse**

`metrics_output_dir = ctx.report_dir / "metrics"` is created (423) but an existing `final_snapshot.msgpack` from a prior run is never removed. If a user reuses `--report-dir` and the new run exits before publishing a fresh final snapshot (SIGTERM, crash), `_load_final_snapshot_from_disk()` decodes the previous run's file and silently builds the new report from stale metrics. **Fix:** unlink any pre-existing `final_snapshot.msgpack` here, OR refuse to decode a snapshot whose timestamp/run-id doesn't match the current run.

### 8. 🔵 low · `bug` · [Claude]

[`src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py:281`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py#L281)

**Double `STARTED` reassigns and orphans the tick task**

`self._publisher.start(...)` is called every time a `SessionEventType.STARTED` record is observed. If two STARTED events ever land in the EventRecord stream (replay, buggy producer, test fixture), `start()` reassigns `_tick_task`; the previous task keeps running until garbage-collected and races with the new task to publish snapshots. **Fix:** guard with `if self._tick_task is None: ...` inside `publisher.start()`, or only call `start()` on the first STARTED in `process()`.

### 9. 🔵 low · `concurrency` · [Claude]

[`src/inference_endpoint/async_utils/services/metrics_aggregator/publisher.py:195`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/publisher.py#L195)

**`close()` cancels tick task but doesn't await it**

`self._tick_task.cancel()` (205) is followed immediately by `self._publisher.close()` (206). If `close()` is invoked from a sync path, the cancelled task may still be runnable when the underlying ZMQ publisher socket is closed → brief window where `_tick` could call `publish()` on a closed transport. `ZmqMessagePublisher.publish()` early-returns on `is_closed` so this isn't fatal, but produces noisy `CancelledError`-during-shutdown traces. **Fix:** make `close()` async-aware (mirror `publish_final`) or document that callers must `await publish_final` (which already cancels the task) before `close`.

### 10. 🔵 low · `error-handling` · [Claude]

[`src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py:166`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py#L166)

**No top-level exception handling around `run_until_complete(main())`**

If `main()` raises (bad CLI args, ZMQ bind failure, tokenizer load error), the subprocess exits with a stack trace but the parent's `ServiceLauncher.wait_for_exit` sees only a non-zero exit code with no diagnostic context propagated upward. **Fix:** wrap with `try/except` + structured logging so failures are surfaced in the parent's logs alongside the exit code.

### 11. 🔵 low · `design` · [Claude]

[`src/inference_endpoint/load_generator/session.py:408`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/load_generator/session.py#L408)

**ERROR-before-COMPLETE relies on an undocumented publisher ordering contract**

The aggregator's `TRACKED_SAMPLES_FAILED` logic (aggregator.py:303–305) assumes ERROR is observed BEFORE COMPLETE removes the row. The ordering is correct here, but it depends on `ZmqMessagePublisher` preserving publish() order through batching AND ZMQ PUB→SUB delivering in order to a single SUB (both currently true). A short comment in session.py noting "publisher must preserve publish-order" would harden this against future transport refactors.

### 12. 🔵 low · `bug` · [Claude]

[`src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py:132`](https://github.com/mlcommons/endpoints/blob/37b20fbe76e5b893657f463647ff2963f7f5f09e/src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py#L132)

**SIGTERM bypasses `publish_final`**

`_finalize()` only sets `_shutdown_event` after `publish_final` and `_publisher.close()` complete, but `__main__.py` doesn't install a SIGTERM handler. On parent-side `launcher.kill_all` (or any external SIGTERM before ENDED arrives), neither pub/sub COMPLETE nor disk fallback runs, leaving the parent's triple-redundant snapshot path empty. **Fix:** `loop.add_signal_handler(SIGTERM, ...)` to flush the final snapshot defensively before exit.

---

_Generated by `/review-council` — Codex `gpt-5.4` (review of git diff against main) + Claude (direct file review with HEAD-source line verification)._

src/inference_endpoint/async_utils/services/metrics_aggregator/snapshot.py
```diff
@@ -0,0 +1,178 @@
...
+
+from enum import Enum
+from typing import ClassVar, Final
+
+import msgspec
+import msgspec.msgpack
+from inference_endpoint.core.record import TOPIC_FRAME_SIZE
+
+
+class SessionState(str, Enum):
```
arekay-nv, line 36: Do we want to add an `initialize` phase - there is not start state which would precede `LIVE`

tests/unit/async_utils/services/metrics_aggregator/test_snapshot.py
```diff
@@ -0,0 +1,125 @@
...
+            percentiles={"50": 100.0, "99": 145.0},
+            histogram=[((50.0, 100.0), 2), ((100.0, 150.0), 3)],
+        )
+        encoded = msgspec.msgpack.encode(stat)
+        decoded = msgspec.msgpack.decode(encoded, type=SeriesStat)
+        assert decoded == stat
+
+
+@pytest.mark.unit
+class TestMetricsSnapshot:
```
arekay-nv, line 67: Can we add some tests to ensure that the state is correctly checked - for instance `LIVE` vs `COMPLETED` vs `PENDING`.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -82,72 +88,156 @@ class MetricCounterKey(str, Enum):
...
 
+# HDR bounds per series — chosen conservatively so realistic benchmark
+# values cannot fall outside [low, high]. Values outside the range are
+# clamped on insert and a warning is logged once per series.
+_NS_HDR_LOW: Final[int] = 1
+_NS_HDR_HIGH: Final[int] = 3_600_000_000_000  # 1 hour in ns
+_TOKEN_HDR_LOW: Final[int] = 1
+_TOKEN_HDR_HIGH: Final[int] = 10_000_000  # 10M tokens
+
+_DRAIN_TIMEOUT_S: Final[float] = 30.0
```
arekay-nv: This might need to be higher.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -82,72 +88,156 @@ class MetricCounterKey(str, Enum):
...
+    mirrors the final snapshot to disk.
     """
 
     def __init__(
         self,
         *args,
-        kv_store: KVStore,
+        registry: MetricsRegistry,
+        publisher: MetricsPublisher,
+        refresh_hz: float,
```
arekay-nv: can we rename this to interval - more consistent with our naming.

src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
```diff
@@ -0,0 +1,423 @@
...
+
+    ``low`` is clamped to ``max(low, 1)`` so the log is well-defined for
+    zero-bound metrics (e.g. token counts starting at 1).
+    """
+    safe_low = max(float(low), 1.0)
+    safe_high = max(float(high), safe_low * 10.0)
+    log_lo = math.log(safe_low)
+    log_hi = math.log(safe_high)
+    step = (log_hi - log_lo) / n_buckets
+    return [math.exp(log_lo + i * step) for i in range(n_buckets + 1)]
```
arekay-nv, line 118: Can we add tests to ensure that the behavior is fixed and any changes are caught by tests. Specifically the internal points/boundaries.

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
```diff
@@ -44,13 +46,37 @@ async def main() -> None:
...
+        "--metrics-output-dir",
+        type=Path,
+        required=True,
+        help="Directory for the final-snapshot disk fallback (created if missing)",
+    )
+    parser.add_argument(
+        "--refresh-hz",
+        type=float,
+        default=4.0,
+        help="Live snapshot publish rate (default: 4.0)",
```
arekay-nv: Suggestion to rename to interval - seems more intuitive.

arekay-nv (approved):
Please do a performance test as well to determine what the limits of the interval and qps are.

nv-alicheng:
**Status update — review-council findings (#4237753512)**

Walking the 12 items against the current pushed state of `prototype/alicheng-zmq-metrics` (HEAD `4110447`):

**Addressed in pushed commits**

- **#1** (`execute.py:482`, subscriber late-binding / slow-joiner) — **Fixed in [`4e31e1b`](https://github.com/mlcommons/endpoints/commit/4e31e1b)**. `MetricsSnapshotSubscriber` is now constructed + `start()`ed *before* `launcher.launch()`, so the SUB handshake completes during the subprocess-spawn window. Connect-before-bind is safe on ZMQ IPC.
- **#2** (`publisher.py:168`, blocking I/O on event loop) — **Fixed in [`4e31e1b`](https://github.com/mlcommons/endpoints/commit/4e31e1b)**. `_write_atomic_fallback` is now wrapped in `asyncio.to_thread(...)` inside `publish_final`, so `f.flush + fsync(file) + fsync(parent) + rename` no longer blocks the aggregator loop.

**Deferred (won't-fix on this PR, tracked)**

- **#3** (`registry.py:208`, unbounded raw-sample retention) — **Deferred** per [`4e31e1b`](https://github.com/mlcommons/endpoints/commit/4e31e1b) commit body. Tracked as a follow-up (`--persist-raw` flag for spill-to-disk). The memory trade is the agreed cost of exact final percentiles in the COMPLETE snapshot; a future PR will add the spill path.

**Outstanding (not yet in pushed commits)**

- **#4** (`aggregator.py:353`, cancellations not awaited before reading `n_pending`)
- **#5** (`registry.py:161`, missing `high >= 2*low` HDR pre-check)
- **#6** (`subscriber.py:55`, `conflate=True` fragile for the Report consumer)
- **#7** (`execute.py:423`, stale `final_snapshot.msgpack` on report-dir reuse)
- **#8** (`aggregator.py:281`, double-STARTED orphans tick task)
- **#9** (`publisher.py:195`, `close()` cancels tick without awaiting)
- **#10** (`__main__.py:166`, no top-level exception handling)
- **#11** (`session.py:408`, undocumented publisher-ordering contract)
- **#12** (`aggregator.py:132`, SIGTERM bypasses `publish_final`)

I'll triage #4–#12 next pass.

src/inference_endpoint/async_utils/transport/protocol.py
```diff
@@ -235,54 +230,76 @@ def cleanup(self) -> None:
...
+
+    The codec is the only type-specific surface in the pub/sub stack. All
+    transport machinery (ZmqMessagePublisher / ZmqMessageSubscriber) operates
+    on (topic_bytes, payload_bytes); the codec is what binds those bytes to
+    a concrete Python type T.
+    """
+
+    def encode(self, item: T) -> tuple[bytes, bytes]:
+        """Return (topic, payload). topic must be exactly TOPIC_FRAME_SIZE bytes."""
+        ...
```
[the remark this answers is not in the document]
nv-alicheng: False positive — ellipses are allowed for Protocol function bodies.

src/inference_endpoint/async_utils/transport/protocol.py
```diff
@@ -235,54 +230,76 @@ def cleanup(self) -> None:
...
+
+    def decode(self, payload: bytes) -> T:
+        """Decode payload back to T. May raise; the caller routes failures
+        through on_decode_error."""
+        ...
```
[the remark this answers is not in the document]
nv-alicheng: False positive — ellipses are allowed for Protocol function bodies.

src/inference_endpoint/async_utils/transport/protocol.py
```diff
@@ -235,54 +230,76 @@ def cleanup(self) -> None:
...
+
+    def on_decode_error(self, payload: bytes, exc: Exception) -> T | None:
+        """Fallback for malformed payloads. Return a sentinel item or None
+        to drop the message."""
+        ...
```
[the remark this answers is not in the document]
nv-alicheng: False positive — ellipses are allowed for Protocol function bodies.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -182,21 +274,35 @@ async def process(self, records: list[EventRecord]) -> None:
...
+                        # Now that we have an event loop running, start the
+                        # publisher tick task. The callable is invoked once
+                        # per tick to capture the live (state, n_pending_tasks)
+                        # pair at each emit.
+                        self._publisher.start(
+                            registry,
+                            self._refresh_hz,
+                            get_runtime_state=lambda: (
+                                self._session_state,
+                                len(table._in_flight_tasks),
```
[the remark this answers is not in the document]
nv-alicheng: **Fixed in [`37300a7`](https://github.com/mlcommons/endpoints/commit/37300a71cb4801cc7861bd1a41c8c2988794ba75)**: aggregator now uses the new `MetricsTable.in_flight_tasks_count` property instead of reaching into `table._in_flight_tasks`.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -232,28 +331,42 @@ async def process(self, records: list[EventRecord]) -> None:
...
-                        MetricCounterKey.TRACKED_SAMPLES_COMPLETED.value,
-                        self._tracked_completed,
-                    )
+                    registry.increment(MetricCounterKey.TRACKED_SAMPLES_COMPLETED.value)
 
         if saw_shutdown:
+            # ENDED has been observed; transition to DRAINING so any tick
+            # that fires before publish_final reflects the new state.
+            self._session_state = SessionState.DRAINING
             logger.info("Draining %d async tasks...", len(table._in_flight_tasks))
```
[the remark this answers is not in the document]
nv-alicheng: **Fixed in [`37300a7`](https://github.com/mlcommons/endpoints/commit/37300a71cb4801cc7861bd1a41c8c2988794ba75)**: drain-log call site now reads `table.in_flight_tasks_count`.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -232,28 +331,42 @@ async def process(self, records: list[EventRecord]) -> None:
...
+                await asyncio.wait_for(table.drain_tasks(), timeout=_DRAIN_TIMEOUT_S)
+            except TimeoutError:
+                logger.warning(
+                    "drain_tasks timed out after %.1fs; some async metrics "
+                    "may be incomplete",
+                    _DRAIN_TIMEOUT_S,
+                )
+                for t in list(table._in_flight_tasks):
+                    if not t.done():
+                        t.cancel()
```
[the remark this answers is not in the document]
nv-alicheng: **Fixed in [`37300a7`](https://github.com/mlcommons/endpoints/commit/37300a71cb4801cc7861bd1a41c8c2988794ba75)**: the per-task cancel loop moved into `MetricsTable.cancel_in_flight_tasks()`, which also returns the cancelled tasks so the aggregator can `await` them (the await landed in [`2b433b1`](https://github.com/mlcommons/endpoints/commit/2b433b14c660f2c830a4a0eff540ee4078590f33) along with council #4).

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -232,28 +331,42 @@ async def process(self, records: list[EventRecord]) -> None:
...
+            n_pending = len(table._in_flight_tasks)
```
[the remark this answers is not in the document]
nv-alicheng: **Fixed in [`37300a7`](https://github.com/mlcommons/endpoints/commit/37300a71cb4801cc7861bd1a41c8c2988794ba75)**: `n_pending = table.in_flight_tasks_count` via the new public property.

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -381,8 +389,8 @@ class MetricsTable:
     Session state is updated via ``handle_session_event``.
     """
 
-    def __init__(self, kv_store: KVStore) -> None:
-        self._kv_store = kv_store
+    def __init__(self, registry: MetricsRegistry) -> None:
+        self._registry = registry
         self._in_flight: dict[str, SampleRow] = {}
         self._triggers: dict[str, list[EmitTrigger]] = {}
         self._in_flight_tasks: set[asyncio.Task] = set()
```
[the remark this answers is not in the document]
nv-alicheng: **Fixed in [`37300a7`](https://github.com/mlcommons/endpoints/commit/37300a71cb4801cc7861bd1a41c8c2988794ba75)**: added `MetricsTable.in_flight_tasks_count` property and `cancel_in_flight_tasks()` method. The cancel method returns the list of cancelled tasks so callers can await them before reading `n_pending` (separately wired in [`2b433b1`](https://github.com/mlcommons/endpoints/commit/2b433b14c660f2c830a4a0eff540ee4078590f33)).

src/inference_endpoint/metrics/report.py
```diff
@@ -24,90 +24,58 @@
...
-        n = stats.count
-        std_dev = math.sqrt((stats.sum_sq - stats.total**2 / n) / (n - 1))
+    if stat.count == 0:
+        return {}
+
+    avg = stat.total / stat.count if stat.count > 0 else 0.0
+    if stat.count > 1:
+        n = stat.count
+        var_num = stat.sum_sq - stat.total * stat.total / n
+        std_dev = math.sqrt(var_num / (n - 1)) if var_num > 0 else 0.0
```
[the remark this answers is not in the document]
nv-alicheng: **Fixed in [`51b4ce1`](https://github.com/mlcommons/endpoints/commit/51b4ce153cf20d4602ec8ee35ef68db0719d6129)**: integer-aggregate series (ns-precision latency) now use `n*sum_sq - total*total` (exact for ints) before the sqrt; float-typed series keep the original form (TPOT magnitudes are small enough that the naive form is fine).

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
nv-alicheng: **Fixed in [`eef11d5`](https://github.com/mlcommons/endpoints/commit/eef11d5ddf7cb127349eae11917d07ef83175231)**: `refresh_hz: float` → `publish_interval_s: float` (seconds), matching the `interval` convention used elsewhere (`check_interval`, `benchmark_httpclient.interval`). The `_s` suffix makes the unit explicit so call sites can't accidentally pass a frequency.

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
nv-alicheng: **Fixed in [`eef11d5`](https://github.com/mlcommons/endpoints/commit/eef11d5ddf7cb127349eae11917d07ef83175231)**: CLI flag `--refresh-hz <Hz>` → `--publish-interval <seconds>` (default 4.0 Hz → 0.25 s, same wire cadence).

src/inference_endpoint/async_utils/services/metrics_aggregator/snapshot.py
nv-alicheng: **Fixed in [`3e32dd6`](https://github.com/mlcommons/endpoints/commit/3e32dd66beda4c75e7257ce1f99893cb7b0e5c19)**: added `SessionState.INITIALIZE` to the wire schema; aggregator now starts in INITIALIZE and transitions to LIVE on the first STARTED. State machine is forward-only (`INITIALIZE → LIVE → DRAINING → COMPLETE`). No INITIALIZE snapshot is emitted on the wire today (the tick task only starts on the first STARTED) — it exists as the well-defined starting point and to carry a future setup-phase tick state.

tests/unit/async_utils/services/metrics_aggregator/test_snapshot.py
nv-alicheng: **Fixed in [`3e32dd6`](https://github.com/mlcommons/endpoints/commit/3e32dd66beda4c75e7257ce1f99893cb7b0e5c19)**: new `TestSessionStateTransitions` class pins the member set, declaration order (consumers can rely on `list(SessionState)` for forward checks), the `complete = state == COMPLETE and n_pending_tasks == 0` rule across every state, and the INITIALIZE round-trip through the codec.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
nv-alicheng: **Fixed in [`d3ba0bc`](https://github.com/mlcommons/endpoints/commit/d3ba0bcbf3eebbf26ac1d29c8e83fc4dde0bb91d)**: `_DRAIN_TIMEOUT_S` default bumped from 30 s → 60 s, and added `--drain-timeout <seconds>` CLI flag plumbed through `MetricsAggregatorService.__init__` as `drain_timeout_s` so long-context tokenize workloads can opt into more headroom without code changes. Analysis: at the system's design point (50k QPS short-context, default 2 workers) the drain finishes in well under a second; the 60 s default covers normal + long-context at default worker count, and `--drain-timeout` covers the 32k-context-at-high-QPS tail.

src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
nv-alicheng: **Fixed in [`225e7ac`](https://github.com/mlcommons/endpoints/commit/225e7ac38b8c84ac3aa5ba61f4ec88f0739a3a86)**: new `TestSeriesSamplerBoundaries` class — 14 tests covering HDR ctor invariants (`high < 2*low` rejected, equality accepted, `low=0` coerced to 1, unsupported dtype rejected), value clamping exactly at `hdr_low`/`hdr_high` (unclamped, no warn), under/over-bound clamping (warn-once per sampler, raw values preserved), float-dtype clamp precision, sig_figs at HDR extremes (1, 5), and rollup edges (count==0 empty histogram, count==1 invariants).

arekay-nv (commented):
# PR Review — ZMQ pub/sub metrics refactor

Five specialised reviews (general code, error-handling, tests, comments, type design) were run in parallel against the diff. Findings below are deduplicated and categorised by priority. Where multiple agents independently converged on the same issue, that is noted — those are the highest-confidence findings.

---

## 🔴 High Priority (correctness — fix before merge)

### H1. Drain-timeout indicator `n_pending_tasks` is always 0 — `state==COMPLETE and n_pending_tasks>0` contract is unenforceable
**File:** `src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py:371-393`

`AGENTS.md` and the `MetricsSnapshot.n_pending_tasks` docstring both document the consumer contract: *drain timeout is detected as `state == COMPLETE and n_pending_tasks > 0`*. But the producer code does `cancel_in_flight_tasks() → await asyncio.gather(*cancelled, return_exceptions=True)`, and each cancelled task's `add_done_callback(self._in_flight_tasks.discard)` fires before `n_pending` is read at line 390. The set is empty by then; the final snapshot reports `n_pending_tasks=0`; `Report.complete` renders **clean success** for a run that actually timed out draining.

Three independent agents (silent-failure, comments, tests) converged on this.

**Fix:** snapshot `len(_in_flight_tasks)` inside the `except TimeoutError:` block *before* cancellation; publish that count. Add a regression test with a forever-blocking `MockTokenizePool` and `drain_timeout_s=0.05` asserting `n_pending_tasks > 0` in the final snapshot.

### H2. `_signal_finalize` task reference is discarded — GC race risks the entire INTERRUPTED delivery path
**File:** `src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py:214`

```python
def _on_sigterm() -> None:
    ...
    loop.create_task(_signal_finalize())  # ← return value discarded
```

Python's `asyncio` docs explicitly warn to save a reference: *\"a task disappearing mid-execution\"*. `_signal_finalize` performs the atomic JSON write and pub/sub send that the entire INTERRUPTED contract depends on — exactly the failure mode the SIGTERM path exists to prevent.

**Fix:** hold the task in a module-level set, remove via `task.add_done_callback`.

### H3. NaN-scrubbed `None` percentiles crash `Report.display()`
**Files:** `src/inference_endpoint/async_utils/services/metrics_aggregator/snapshot.py:239`, `src/inference_endpoint/metrics/report.py:306, 328`

`_scrub_nonfinite` maps non-finite percentiles to `None` to keep `json.dumps(allow_nan=False)` honest, but `report.py:306` does `metric_dict[key] * scale_factor` and `:328` does `val * scale_factor` with no `None` guard. `finalize_benchmark` calls `display()` outside the report-build try/except, so any producer-side NaN takes down the whole `run_benchmark` flow with `TypeError`.

**Fix:** render `N/A` (or skip) for `None` values in `_display_metric`; add a round-trip test that puts NaN into `registry.set_counter`, builds snapshot, runs through `snapshot_to_dict` → `Report.from_snapshot` → `display()` without crashing.

---

## 🟡 Medium Priority (rule violations, untested load-bearing code, doc/code mismatches)

### M1. Docstrings claim SIGINT also calls `publish_final` — it doesn't
**Files:** `publisher.py:91-94, 166-167, 193-194`

All three say *\"the SIGTERM/SIGINT handler in `__main__.py` ... can both call it\"*. But `__main__.py:240-248` installs SIGINT as a log-only no-op handler. Only SIGTERM calls `publish_final`. Change all three to \"SIGTERM\" singular.

### M2. SIGTERM-handler comment misdescribes its trigger
**File:** `__main__.py:186-208`

Comment says the primary reason for the SIGTERM handler is `ServiceLauncher.kill_all` — but `kill_all` calls `proc.kill()` (SIGKILL on POSIX, uncatchable). The handler never fires from that path. Either fix the comment to describe the real trigger (external operator `kill -TERM`), or change `kill_all` to `terminate()` + grace period + `kill()` so it actually exercises the handler.

### M3. \"Disk fallback\" wording inverts the design
**Files:** `execute.py:440`, `__main__.py:65`, `AGENTS.md:202`

All three describe the JSON file as the \"disk fallback\". But `publisher.py:48-55` and the narrative section of `AGENTS.md` call it the **primary** Report source, with pub/sub as the TUI signal. Inverted polarity. Pick one story (the JSON file is primary) and apply it consistently across docs and comments.

### M4. `publish_final` idempotency has no test
**File:** `tests/unit/async_utils/services/metrics_aggregator/test_publisher.py`

`publisher.py:196-198`'s `if self._finalized: return` is the explicit race protector for SIGTERM-vs-ENDED. Both `_signal_finalize` (`__main__.py:216-237`) and the ENDED path (`aggregator.py:399`) call `publish_final`. A regression that removed the `_finalized` guard would double-write the JSON file and the suite would stay green. **Add:** `test_publish_final_idempotent` asserting `publish.call_count == 1` and unchanged file mtime after a second call.

### M5. Drain-timeout recovery path is entirely untested
**File:** `tests/unit/async_utils/services/metrics_aggregator/test_aggregator.py`

`aggregator.py:371-389` is the recovery for blocked tokenize tasks. No test exercises it. Compounds with H1: even after fixing the contract, there is no test to prevent regression. **Add:** integration test with a forever-blocking `MockTokenizePool` + `drain_timeout_s=0.05`.

### M6. `.tmp`-cleanup leg of `_write_atomic_json` has no test
**File:** `tests/unit/async_utils/services/metrics_aggregator/test_publisher.py`

`publisher.py:264-272`'s `try / except BaseException: tmp.unlink(missing_ok=True); raise` is the only thing preventing `.tmp` leak across aggregator restarts. The existing disk-failure test fails before `.tmp` is created. **Add:** patch `os.rename` to raise mid-call, assert no `.tmp` left behind.

### M7. Bare `except asyncio.CancelledError: pass` violates AGENTS.md
**File:** `src/inference_endpoint/load_generator/session.py:299-300`

Project rules require every `except` block to contain a comment or logging. Adjacent blocks at `publisher.py:203-205` and `session.py:338` already comply. Add a one-line comment here.

### M8. Default-to-INTERRUPTED on missing `state` masks malformed snapshots
**File:** `src/inference_endpoint/metrics/report.py:200`

```python
state = snap.get(\"state\", \"interrupted\")
```

A snapshot dict missing `state` is malformed, not interrupted. Defaulting silently rewrites a schema violation as a normal \"user pressed Ctrl-C\" warning. Log at ERROR with the snapshot keys, then either raise or use a distinct `\"malformed\"` sentinel.

### M9. `_load_final_snapshot_from_disk` collapses \"missing\" vs \"corrupt\"
**File:** `src/inference_endpoint/commands/benchmark/execute.py:387-405`

Both file-missing and corrupt-JSON return `None`; downstream log says \"no file on disk\" even when the file exists but failed to parse. Corrupt JSON is a producer-contract violation — should be ERROR level with a distinct message including the parse exception.

### M10. Fallback warning at `execute.py:575-580` says \"may or may not be terminal\"
**File:** `src/inference_endpoint/commands/benchmark/execute.py:575-580`

`metrics_subscriber.latest.state` is already available. A `LIVE` state means the aggregator died mid-run (data lost); `COMPLETE`/`INTERRUPTED` means the disk write failed. These deserve different log messages.

### M11. Add `gc=False` to `CounterStat`, `SeriesStat`, `MetricsSnapshot`
**File:** `src/inference_endpoint/async_utils/services/metrics_aggregator/snapshot.py`

Matches project convention (per AGENTS.md and `core/types.py`). All fields are scalars, frozen dicts of scalars, or lists of frozen-scalar-only Structs — no cycle path exists. One snapshot every publish-interval × N series is high-frequency enough to matter.

---

## 🟢 Low Priority (suggestions)

### Wire schema
- **Rename `MetricsSnapshot.counter` → `seq` / `emit_seq`** — collides namespace-wise with `CounterStat`; docstring already has to disclaim the confusion. Easier to fix before downstream consumers ship.
- **Replace `((lo, hi), count)` histogram tuple with a named `HistogramBucket` Struct** — lossy round-trip through `snapshot_to_dict`; consumers index positionally (`h[0][0]`).

### Encapsulation
- **`MetricsSnapshotSubscriber.latest` should be a `@property`** with private backing field — currently public-mutable, could be silently overwritten.
- **`MetricsTable.is_tracking` / `session_started_ns`** are public mutable attributes — consider `@property` accessors.

### Error handling polish
- **`publisher.start()` should validate `publish_interval_s > 0` before the idempotency early-return** (`publisher.py:120-130`) — current ordering can swallow validation errors.
- **`_write_atomic_json` parent-dir fsync error is misattributed as \"JSON snapshot write failed\"** — file is on disk; only durability is at risk.
- **`MetricsSnapshotCodec.on_decode_error` returns `None` silently** (`snapshot.py:270-277`) — a `msgspec.DecodeError` is a schema bug; add `logger.warning` once with frame size + exc.
- **`http_client.shutdown_async` cleanup uses `f\"{e}\"` (execute.py:548-552)** — bump to `logger.exception` so operators investigating leftover workers have a chain.

### Test polish
- **`test_aggregator_e2e.py` is marked `@pytest.mark.unit` but binds real ZMQ IPC** — should be `@pytest.mark.integration`.
- **`time.sleep(2.0)` readiness wait in `test_signal_handling.py:117, 166`** — `__main__.py` already supports `--readiness-path`/`--readiness-id`; use it.
- **Add negative test for `register_series` rejecting percentiles tuple without 50.0** — `registry.py:388-393` guard currently has no test.
- **Add `_scrub_nonfinite` round-trip test** — inject NaN via `registry.set_counter`, assert `None` in dict + clean `json.dumps(allow_nan=False)`.
- **Add aggregator `_session_state` transition assertions** (INITIALIZE → LIVE → DRAINING) at unit level.
- **`test_publisher.py`: `MetricsPublisher.start()` double-start guard** (`publisher.py:120-126`) has no test.
- **`test_publisher.py:170-176`** has `try: publisher.close() except Exception: pass` — comment suggests author wasn't sure if it should raise. Construct the mock so `close()` succeeds.
- **`test_on_decode_error_drops_malformed` (test_snapshot.py:112-119)** uses `try/except` instead of `pytest.raises` — silently no-ops if msgspec ever stops raising.
- **`test_offline_benchmark`** (integration) should assert `final_snapshot.json` exists and `state == \"complete\"` — given the file is now primary, a regression that never writes it would not be caught.

### Comment hygiene
- **Stale ref: `registry.py:379` says `Report._series_to_metric_dict`** — function is module-level in `metrics/report.py` now.
- **`metrics_table.py:332-339` TpotTrigger docstring contains literal `# NOTE(agents):`** — renders as `#` text in `help()`. Also drop the \"if tokenization throughput becomes a bottleneck, consider merging...\" speculation (per AGENTS.md TODOs must link an issue).
- **`report.py:200` bare TODO** (`# TODO: surface session_started_ns via snapshot`) — link an issue or drop.
- **`subscriber.py:57-60`** \"no current callers do\" parenthetical will rot — rephrase.
- **`aggregator.py:305`** comment \"Now that we have an event loop running\" is misleading (loop is always running in `async def`); the real precondition is STARTED was observed.

### Type design
- **`SampleRow` `gc=False` docstring** should use the standard AT-RISK pattern. It is mutated post-construction; current docstring (\"no mutable container fields\") understates the audit argument.
- **`SeriesSampler._raw: array.array` is unbounded** — documented but not type-expressed. Lifetime is the run, not the snapshot interval.
- **Lift `get_runtime_state: Callable[[], tuple[SessionState, int]]`** to a named `RuntimeState` (NamedTuple or array_like Struct).
- **Co-locate `MetricCounterKey` and `MetricSeriesKey`** (currently `aggregator.py` vs `metrics_table.py`) into a single `keys.py`.

---

## Strengths (worth keeping)

- **Excellent comment density on tricky invariants** — `session.py:411-421` (ERROR-before-COMPLETE), `aggregator.py:284-301` (duplicate-STARTED), `aggregator.py:381-389` (cancel-then-await), `publisher.py:42-65` (dual-path contract). All describe current state with rationale, no history narration.
- **Dual-path final delivery is genuinely independent** — `test_disk_failure_does_not_block_pubsub` (test_publisher.py:140) proves disk failure does not block pub/sub. Exemplary defensive design.
- **`_finalized` flag set before any `await`** (`publisher.py:196-198`) — SIGTERM-vs-ENDED double-publish race is race-free, not just narrow.
- **`test_aggregator_error_handler.py:144-197`** explicitly parametrises the ERROR-before-COMPLETE counter-example with `test_error_after_complete_misses_tracked_failed` — textbook regression-resistant test.
- **HDR `high >= 2*low` pre-check** (`registry.py:160-167`) rejects misconfiguration loudly instead of cryptic HdrHistogram errors.
- **`registry.py:388-393` p50 contract enforcement** rejects misregistration that would silently produce midrange-fallback medians.
- **Disciplined msgspec usage** (frozen + array_like + tagged unions) is correct hot-path shape.
- **Integration tests for SIGINT-no-op and SIGTERM-final-write** are in the right shape and right place.
- **License headers present, no lazy imports, no history-narrating comments, no gitignored refs** across the reviewed surface.

---

## Recommended Action Plan

1. **High priority (H1–H3)** are correctness bugs that materially mislead the user. Fix before merge.
2. **Medium priority (M1–M11)** — tighten the SIGINT/SIGTERM/disk-fallback documentation (M1–M3) and add the missing tests for load-bearing recovery code (M4–M6). These are cheap and currently the docs lie about the design.
3. **Low priority** can be batched into a follow-up PR.
4. Re-run targeted reviews after fixes (`errors` + `tests` aspects).

🤖 Generated by `/pr-review-toolkit:review-pr` (5 specialised agents — code, errors, tests, comments, types).

arekay-nv (approved):
@nv-alicheng 
The three high priority changes seem important and small enough to fix. 
Everything else can be deferred to followup. 
Please merge when ready and collect baseline perf numbers pre/post commit to see if there is any major perf implications. 
Thanks!

PR #307, opened 2026-05-05
fix: OpenAI response decoding defaults, multimodal prompt handling, and tokenizer_name override

BolinSNLHM:
### Summary

Fixes three issues encountered when running the Qwen3-VL-235B-A22B benchmark against vLLM/Dynamo with the Shopify multimodal dataset:

1. **`openai/types.py`** — vLLM omits optional response fields (`content`, `refusal`, `finish_reason`, `usage`, `system_fingerprint`, etc.) that our `msgspec` structs required to be present. Added safe defaults so decoding succeeds regardless of what the server populates.

2. **`session.py`** — The `ShopifyMultimodalFormatter` produces `prompt` as `list[dict]` (OpenAI vision content parts), which the HTTP adapter handles correctly. However, `PhaseIssuer`'s metrics tracking assigned it directly to `PromptData.text` (typed `str | None`), causing a decode failure. Added a type guard to coerce non-string prompts to `None`.

3. **`config/schema.py` + `execute.py`** — When the model is served from a container-local path (e.g. an NVFP4 snapshot under `/root/.cache/huggingface/hub/...`), tokenizer loading fails because the path isn't a valid HF repo ID. Added `model_params.tokenizer_name` to decouple the tokenizer source from the serving model name.

### Testing

Validated end-to-end on GB300 (offline scenario, 1000 samples): all samples completed successfully with no decode errors and full latency/throughput metrics generated.

**Note:** The example YAMLs assume the server is started with `--served-model-name` matching `model_params.name`. When serving from a local NVFP4 checkpoint (the typical NVIDIA submission setup), set `model_params.name` to the actual served path, `tokenizer_name` to the upstream HF repo ID (e.g. `Qwen/Qwen3-VL-235B-A22B-Instruct`), and `endpoint_config.endpoints` to the node IP.

### DIFF
--- a/src/inference_endpoint/commands/benchmark/execute.py
+++ b/src/inference_endpoint/commands/benchmark/execute.py
@@ -179,27 +179,38 @@ class BenchmarkContext:
 
 
 def _check_tokenizer_exists(model_name: str) -> bool:
-    """Check if a HuggingFace tokenizer exists for the model (API only, no download).
+    """Check if a tokenizer exists for the model (local dir or HF repo, no download).
 
-    Returns True if the model repo exists and has tokenizer files, False otherwise.
-    This function is a probe — it never loads or downloads the tokenizer itself.
-    Downstream consumers that need tokenization (e.g. the MetricsAggregator
-    subprocess for ISL/OSL/TPOT, Harmony transforms for prompt preprocessing,
-    and any future plugin with its own tokenization need) each load their own
-    instance as required.
+    Returns True if a tokenizer is available, False otherwise. This function is
+    a probe — it never loads or downloads the tokenizer itself. Downstream
+    consumers that need tokenization (e.g. the MetricsAggregator subprocess
+    for ISL/OSL/TPOT, Harmony transforms for prompt preprocessing, and any
+    future plugin with its own tokenization need) each load their own instance
+    as required.
+
+    ``model_name`` may be a local checkpoint directory (e.g. an NVFP4 snapshot
+    cached under ``/root/.cache/huggingface/hub/...``) or an HF repo ID. Local
+    directories are probed directly; otherwise we ask the HF Hub for the file
+    listing.
     """
     try:
-        info = model_info(model_name)
-        # Check for tokenizer files in the repo
-        siblings = {s.rfilename for s in (info.siblings or [])}
+        local_path = Path(model_name)
+        if local_path.is_dir():
+            siblings = {p.name for p in local_path.iterdir() if p.is_file()}
+        else:
+            info = model_info(model_name)
+            siblings = {s.rfilename for s in (info.siblings or [])}
+
         has_tokenizer = (
             "tokenizer_config.json" in siblings or "tokenizer.json" in siblings
         )
+
         if has_tokenizer:
             logger.info(f"Tokenizer available for model: {model_name}")
         else:
             logger.warning(f"Model {model_name} found but has no tokenizer files")
         return has_tokenizer
+
     except ImportError:
         # huggingface_hub not installed — fall back to assuming it works
         logger.info(
--- a/src/inference_endpoint/load_generator/session.py
+++ b/src/inference_endpoint/load_generator/session.py
@@ -191,8 +191,14 @@ class PhaseIssuer:
         prompt_data: PromptData
         if isinstance(data, dict):
             token_ids = data.get("input_tokens") or data.get("token_ids")
+            # Multimodal datasets store ``prompt`` as a list of OpenAI content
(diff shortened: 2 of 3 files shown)


src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -300,9 +300,16 @@ def setup_benchmark(config: BenchmarkConfig, test_mode: TestMode) -> BenchmarkCo
...
+    # When the serving model name is a local/container path (e.g. an NVFP4
+    # checkpoint cached under /root/.cache/huggingface/hub/...) it is not a
+    # valid HF repo ID and the probe will fail. Allow model_params.tokenizer_name
+    # to override the source so the upstream HF tokenizer can still be used.
     model_name = config.model_params.name
-    tokenizer_name = model_name if _check_tokenizer_exists(model_name) else None
+    tokenizer_source = config.model_params.tokenizer_name or model_name
+    tokenizer_name = (
+        tokenizer_source if _check_tokenizer_exists(tokenizer_source) else None
+    )
```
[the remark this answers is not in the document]
Victor49152: This is a fair point

Victor49152 (changes requested):

examples/08_Qwen3-VL-235B-A22B_Example/online_qwen3_vl_235b_a22b_shopify.yaml
```diff
@@ -6,6 +6,7 @@ timeout: 14400
 
 model_params:
   name: "Qwen/Qwen3-VL-235B-A22B-Instruct"
+  # tokenizer_name: "Qwen/Qwen3-VL-235B-A22B-Instruct"  # Set this if model name is a local/container path
```
wangshangsam: Same comment as https://github.com/mlcommons/endpoints/pull/307/changes#r3210643151

wangshangsam (commented):
The override works for the case where the local checkpoint genuinely doesn't ship tokenizer files, but I think it's papering over a probe bug rather than fixing the root cause.

`_check_tokenizer_exists` (`execute.py:181`) calls `huggingface_hub.model_info(model_name)` first, which only resolves repo IDs. Pass it a path like `/root/.cache/huggingface/hub/...` and it raises → `except Exception` → returns `False` → `tokenizer_name` becomes `None` at `execute.py:305` → `execute.py:465` never passes `--tokenizer` to the metrics aggregator subprocess → ISL/OSL/TPOT silently disabled.

But the downstream loader at `token_metrics.py:79` is `AutoTokenizer.from_pretrained(...)`, which accepts a local directory containing `tokenizer.json` / `tokenizer_config.json` just fine. So there are really two cases:

1. **Local NVFP4 snapshot that contains tokenizer files** — the typical case, since vLLM needs the tokenizer to serve, so it's almost always present in the snapshot. The probe is just being too strict; a short-circuit like "if `Path(model_name).is_dir()` and a tokenizer file exists, return True" would make everything work with no user-visible knob.
2. **Local path with no tokenizer files** — rare (weights-only directories). Here `tokenizer_name` is the right escape hatch.  -> But this is very rarely to happen.

This PR ships (2) but leaves (1) broken — even when the served directory contains a perfectly usable `tokenizer.json`, the user is still forced to set `tokenizer_name`. Could we enable `_check_tokenizer_exists` to support checking whether tokenizer files exist for a model checkpoint in local path?

src/inference_endpoint/load_generator/session.py
```diff
@@ -191,8 +191,14 @@ def issue(self, sample_index: int) -> str | None:
         prompt_data: PromptData
         if isinstance(data, dict):
             token_ids = data.get("input_tokens") or data.get("token_ids")
+            # Multimodal datasets store ``prompt`` as a list of OpenAI content
+            # parts (e.g. [{"type": "text", ...}, {"type": "image_url", ...}])
+            # which the HTTP adapter handles directly. PromptData.text is only
+            # meaningful for ISL on text-only prompts, so coerce non-strings
+            # to None and rely on token_ids when the dataset pre-tokenizes.
```
wangshangsam:
```suggestion
            # which the HTTP adapter handles directly. `PromptData.text` is only
            # meaningful for ISL reporting on text-only prompts.
            # Therefore, setting `text=None` for non-string prompts 
            # means that ISL reporting will be unavailable for multimodal samples.
```

`rely on token_ids when the dataset pre-tokenizes` implies that VLM pre-tokenizes the input which is actually not the case.

wangshangsam (approved):
Nit, but otherwise LGTM!

Victor49152 (approved):

arekay-nv (commented):
Thanks!

PR #314, opened 2026-05-18
Agentic Inference updates

hvagadia:
## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass


docs/MULTI_TURN_QUICKSTART.md
```diff
@@ -41,6 +41,7 @@ datasets:
     type: performance
     path: path/to/your/conversations.jsonl
     multi_turn: # ← Presence of this block enables multi-turn mode
+      mode: independent # ← Per-conv pipelines; no cross-conv turn barrier
```
nvzhihanj: From the description and readme below, it doesn't seem the "cross-conv turn barrier" is something applicable here. What would be alternative modes we are thinking about?

One possible alternative is sub-agent mode, but I assume the trace format itself can address this

docs/MULTI_TURN_QUICKSTART.md
```diff
@@ -41,6 +41,7 @@ datasets:
...
       turn_timeout_s: 300 # ← Max wait for prev turn
```
nvzhihanj, line 44: Will the output be cut off, or will the process just fail?

It makes sense that we need a timeout, but not sure if we need it specifically for multi-turn (under the multi-turn indentation) 🤔 Will a overall per-response timeout be enough for this? cc: @arekay-nv

docs/MULTI_TURN_QUICKSTART.md
```diff
@@ -81,24 +83,58 @@ After the benchmark completes, check the directory configured via `report_dir`:
...
+```
 
 ### Metrics
 
 Currently available:
 
-- **Per-turn metrics**: Latency, TTFT, TPOT recorded per client turn.
-- **Conversation correlation**: events carry `conversation_id` and `turn`
-  directly; group / aggregate as needed in post-processing.
+- **Per-turn metrics**: Latency, TTFT, TPOT for each turn
```
nvzhihanj: Does it make sense to also compute the latency / # tokens per-turn (TBT)?

docs/MULTI_TURN_QUICKSTART.md
```diff
@@ -81,24 +83,58 @@ After the benchmark completes, check the directory configured via `report_dir`:
...
+---
+
+## Conversation Modes Explained
+
+### Independent Mode (Default)
 
-_Note: First-class per-conversation aggregation (e.g., "conversations/sec",
-end-to-end conversation latency) is planned but not yet implemented._
+```yaml
+mode: independent
```
nvzhihanj: Might not need a `mode` if this is the only option

docs/MULTI_TURN_QUICKSTART.md
nvzhihanj, line 144: Is this true for workflow dataset?

nvzhihanj, line 204: I am guessing this is generated by AI, but doesn't seem a real option in benchmarking?

nvzhihanj, line 188: Will system prompt be distinguished in the trace format?

nvzhihanj, line 181: I am thinking whether sub-agent will be supported in this trace format. Okay if it doesn't, but might need a note

docs/MULTI_TURN_QUICKSTART.md
hvagadia: Deleted this file

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Inline accuracy scorer for multi-turn benchmark runs.
```
nvzhihanj: The name of the file needs to be more explicit (and aligned with this summary)

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+
+Scoring rules:
+    Coding turn:   IoU of the multisets of canonical bash exes used in the turn
+    Workflow turn: 1 if `intent: IXXX` code matches gt, else 0
```
nvzhihanj: Should these be 2 separate scorer? IIRC the design was to have modularized accuracy checker for different components, so they can be reused for different dataset (e.g. if we have a longer coding dataset, we can reuse this for that)

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+logger = logging.getLogger("inline_accuracy")
+
+# ---------------------------------------------------------------------------
+# Coding scorer - extract whitelisted exes per turn, score = multiset IoU
+# ---------------------------------------------------------------------------
+
+# Canonical exe whitelist. Synonyms collapse to one canonical name.
+# Anything not in this map is dropped. cd / pwd / echo are intentionally
+# absent: they're navigation/output verbs, not action classes.
+EXE_MAP: dict[str, str] = {
```
nvzhihanj: Does this map haave an original reference (paper/code)? Good to put a reference if it's Aapache2.0 license

examples/09_MultiTurn/kimi_agentic_benchmark.yaml
```diff
@@ -0,0 +1,44 @@
...
+datasets:
+  # Select the dataset to run by updating both `name` and `path`.
+  # Use agentic_coding for coding traces or agentic_workflow for workflow traces.
+  - name: agentic_coding
+    type: performance
+    path: /path/to/agentic_dataset.jsonl
+    multi_turn:
+      turn_timeout_s: 600.0
+      enable_salt: true # add salt after system prompt to prevent cache reuse across trajectories
+      inject_tool_delay: true # add delay before user/tool turns
```
nvzhihanj, line 23: (I might have missed the prev PR) is the tool delay injected dynamically (calced based on distribution), or statically (embedded in the dataset)?

(Same Q for the salt as well)

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
nvzhihanj: @nv-alicheng to suggest as well

nvzhihanj (approved):

examples/09_MultiTurn/kimi_agentic_benchmark.yaml
hvagadia: tool delay is embedded in the dataset from the proposed distribution and used statically by client. Salt is deterministically computed at runtime.

src/inference_endpoint/load_generator/multi_turn_strategy.py
```diff
@@ -210,22 +213,48 @@ async def execute(self, phase_issuer: PhaseIssuerProtocol) -> int:
...
-        if pair is None:
+        turns, cursor = state
+        if cursor >= len(turns):
             del self._active_iters[conv_id]
             self._fill_slot()
             return
 
-        idx, turn = pair
+        idx, turn = turns[cursor]
+        self._active_iters[conv_id] = (turns, cursor + 1)
```
nv-alicheng: [Codex+Claude] **high** (bug): Cursor is advanced to `cursor + 1` at this line **before** the scheduled delay executes.

If a timeout on the preceding in-flight turn triggers `_abort_remaining_turns`, it reads the already-advanced cursor (line 417: `turns, cursor = state`) and iterates `turns[cursor:]` — skipping the turn whose delay was just cancelled. `mark_turn_failed` and `register_skipped`/`_publish_synthetic_failure` are never called for that turn, leaving ConversationManager's completion count off by one and the benchmark missing the failure event entirely.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+            n_perfect += 1
+        elif s == 0.0:
+            n_zero += 1
+    n = len(results)
+    return {
+        "domain": domain,
+        "n_scorable": n,
+        "n_perfect": n_perfect,
+        "n_zero": n_zero,
+        "pass_rate": round(total / n, 4) if n else float("nan"),
```
nv-alicheng: [Codex+Claude] **high** (data-integrity): `float("nan")` is serialized as the bare token `NaN` by `json.dumps` (line 518) without `allow_nan=False`.

`NaN` is not valid JSON per RFC 8259 — `jq`, Go's `encoding/json`, and most strict parsers will refuse to read the output file. Replace `float("nan")` with `None` so the field becomes `null` in JSON, which is valid and clearly signals no data.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+
+def _canonicalize_stage(stage: str) -> str | None:
+    """Return the whitelisted canonical exe of a single pipeline stage, or None."""
+    tokens = stage.split()
+    i = 0
+    while i < len(tokens) and (_ENVKV_RE.match(tokens[i]) or tokens[i] in _WRAPPERS):
+        i += 1
+    if i >= len(tokens):
+        return None
+    leaf = _PATH_LEAF.search(tokens[i]).group(0).lower()
```
nv-alicheng: [Claude] **medium** (bug): `_PATH_LEAF.search(tokens[i]).group(0)` raises `AttributeError` when `tokens[i]` is exactly `"/"` (a bare slash).

`re.compile(r"[^/]+$")` returns `None` for `"/"` because there are no non-slash characters before `$`. This aborts the entire scoring run. Guard against it:
```python
m = _PATH_LEAF.search(tokens[i])
if m is None:
    return None
leaf = m.group(0).lower()
```

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+        "per_turn": results,
+    }
+
+
+# ---------------------------------------------------------------------------
+# Events to model assistants
+# ---------------------------------------------------------------------------
+
+
+def _build_index_to_key(gt_jsonl: Path) -> list[tuple[str, int]]:
```
nv-alicheng: [Claude] **medium** (data-integrity): `_build_index_to_key` silently produces wrong results if the gt JSONL has a different turn count or order than the benchmark run dataset.

The positional index list is used to map `sample_idx_map.json` integer indices to `(conv_id, turn)` keys. If the gt file has extra or missing conversations, all subsequent UUID-to-key mappings past the mismatch point attribute model outputs to the wrong turns — producing incorrect scores with no warning. At minimum, assert or warn when `len(index_to_key)` differs from the number of UUIDs in `uuid_to_index`.

tests/integration/test_multi_turn.py
```diff
@@ -833,3 +835,125 @@ def test_live_history_rejects_tool_turns():
...
+@pytest.mark.asyncio
+async def test_delay_does_not_leak_to_endpoint_payload():
+    received_payloads: list[dict] = []
+
+    class CapturingEchoServer(EchoServer):
+        async def _handle_echo_chat_completions_request(self, request):
+            try:
+                payload = await request.json()
+                received_payloads.append(payload)
+            except Exception:
```
nv-alicheng: [Claude] **low** (error-handling): `except Exception` silently swallows request-parse failures.

`logger = None  # noqa: F841` suppresses the ruff warning but provides zero signal when parsing fails. Per project rules (AGENTS.md), every `except` block must log or include a comment explaining why the error is ignored. A parse failure here would leave `received_payloads` empty with no diagnostic output, making failures very hard to debug.

src/inference_endpoint/openai/types.py
```diff
@@ -128,6 +129,7 @@ class ChatCompletionRequest(
     logit_bias: dict[str, float] | None = None
     user: str | None = None
     chat_template: str | None = None
+    chat_template_kwargs: dict[str, Any] | None = None
```
nv-alicheng, line 133: [Claude] **low** (design): The `gc=False` AT-RISK audit comment at line 112 lists `messages`, `tools`, and `logit_bias` but omits the newly added field at this line.

`chat_template_kwargs: dict[str, Any] | None` is also a mutable container. While the current Kimi use-case values are `dict[str, bool]` (cycle-free), the `Any` type admits cyclic references in general. Add `chat_template_kwargs` to the audit comment so future reviewers know it was considered.

src/inference_endpoint/dataset_manager/multi_turn_dataset.py
```diff
@@ -397,6 +410,11 @@ def _build_metadata(self) -> ConversationMetadata:
                 if val and isinstance(val, str):
                     system_content = val
                     break
+            if self._enable_salt and system_content:
```
nv-alicheng, line 417: [Claude] **low** (design): Salt is silently skipped for conversations that have no system prompt (`system_content` is `None`).

`if self._enable_salt and system_content:` means that when `enable_salt=True` is requested, conversations without a system row receive no cache differentiation — partially defeating the purpose of the flag. At minimum, log a warning at INFO level identifying which conversations were skipped so the caller knows they are unprotected.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+    keys: list[tuple[str, int]] = []
+    by_conv: dict[str, list[dict]] = {}
+    for line in gt_jsonl.read_text().splitlines():
```
nv-alicheng: [Claude] **low** (performance): `gt_jsonl.read_text().splitlines()` loads the entire ground-truth file into memory as a single string before splitting.

For large benchmark datasets this doubles peak memory vs. streaming. `_iter_assistant_turns` (line 268) already uses the correct pattern — iterate the open file handle line-by-line:
```python
with open(gt_jsonl) as fh:
    for line in fh:
        ...
```

nv-alicheng (commented):

Found **11** issues across **7** files (3 couldn't be posted inline — see summary comment).

nv-alicheng:

Found **11** issues across **7** files.

> ⚠️ 3 issues below couldn't be posted as inline comments (lines not in the diff hunk) — they are included here with file:line references.

---

Issues that will cause hangs, data loss, or incorrect behavior in production.

| 1 | `src/inference_endpoint/load_generator/multi_turn_strategy.py` | 326 | bug | Codex | ⚠️ *inline not possible* — Timeout path decrements `inflight` but **never** calls `_drain_event.set()`. If the last in-flight turn times out, `_drain_inflight` in `session.py` waits forever — the benchmark hangs indefinitely. Fix: mirror session.py:540 (`if self._phase_issuer.inflight <= 0: self._drain_event.set()`) in the timeout handler. |
| 2 | `src/inference_endpoint/load_generator/multi_turn_strategy.py` | 232 | bug | Codex+Claude | Cursor advanced to `cursor + 1` before the delayed turn fires. If a timeout aborts the conversation, `_abort_remaining_turns` reads the already-advanced cursor and skips the cancelled delayed turn — `mark_turn_failed` and `_publish_synthetic_failure` are never called for it, silently under-counting failures. |
| 3 | `src/inference_endpoint/load_generator/session.py` | 537 | bug | Codex | ⚠️ *inline not possible* — `completed_uuids` set accumulates every completed + timed-out query ID for the full phase and is never cleared. At 50 k QPS / 600 s this can hold millions of UUID strings and consume several hundred MB before the phase ends. The set's purpose (guard against late double-responses) only requires remembering IDs that were synthetically completed after a timeout — not every successful query. |
| 4 | `examples/09_MultiTurn/accuracy/score_inline_accuracy.py` | 328 | data-integrity | Claude | `float("nan")` serialized as bare `NaN` token by `json.dumps` without `allow_nan=False`. `NaN` is not valid JSON per RFC 8259; `jq`, Go's `encoding/json`, and most strict parsers reject the output file. Replace with `None` → `null`. |

---

### 🟡 Should Fix (medium)

Real issues that produce incorrect results under reachable conditions.

| 5 | `src/inference_endpoint/core/types.py` | 224 | bug | Codex | ⚠️ *inline not possible* — `as_message_parts_after_first_chunk` returns the full `self.tool_calls` for TPOT tokenization, but for streaming tool-call responses some of those tokens arrived in the first chunk (before TTFT). TPOT denominator is systematically inflated for all tool-calling runs. |
| 6 | `examples/09_MultiTurn/accuracy/score_inline_accuracy.py` | 132 | bug | Claude | `_PATH_LEAF.search(tokens[i]).group(0)` raises `AttributeError` when `tokens[i]` is `"/"` (bare slash). `re.compile(r"[^/]+$")` returns `None` for that input; unguarded `.group(0)` aborts the entire scoring run. |
| 7 | `examples/09_MultiTurn/accuracy/score_inline_accuracy.py` | 340 | data-integrity | Claude | `_build_index_to_key` has no guard against a mismatch between the gt JSONL and the benchmark run dataset. A difference in turn count/order silently maps model outputs to wrong turns, producing incorrect scores with no warning. |

---

### 🔵 Consider (low)

Valid improvements, suitable as follow-ups.

| 8 | `tests/integration/test_multi_turn.py` | 939 | error-handling | Claude | `except Exception` swallows parse failures with `logger = None` — no log, no comment. AGENTS.md requires every `except` to explain or log. |
| 9 | `src/inference_endpoint/openai/types.py` | 132 | design | Claude | `gc=False` AT-RISK comment (line 112) lists `messages`, `tools`, `logit_bias` but omits the newly added `chat_template_kwargs: dict[str, Any] \| None`. |
| 10 | `src/inference_endpoint/dataset_manager/multi_turn_dataset.py` | 413 | design | Claude | Salt silently not applied to conversations without a system prompt; no warning logged when `enable_salt=True` is partially defeated. |
| 11 | `examples/09_MultiTurn/accuracy/score_inline_accuracy.py` | 343 | performance | Claude | `gt_jsonl.read_text().splitlines()` loads the entire file into memory. Use line-by-line streaming (same pattern as `_iter_assistant_turns` line 268). |

---

🤖 Generated by `/review-council` (Codex + Claude, thorough depth).

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+# ---------------------------------------------------------------------------
+# Workflow scorer - extract `intent: IXXX` code, exact match
+# ---------------------------------------------------------------------------
+
+_INTENT_RE = re.compile(r"\bintent:\s*(I\d{3})\b", re.IGNORECASE)
+_BARE_INTENT_RE = re.compile(r"\bI(\d{3})\b")
+
+
+def _extract_intent_code(turn: dict) -> str | None:
+    """Two-stage extractor:
```
viraatc: any OSS implementations / papers / conventions we could link to for this section?

examples/09_MultiTurn/kimi_agentic_benchmark.yaml
```diff
@@ -0,0 +1,44 @@
...
+    min_duration_ms: 0
+
+  load_pattern:
+    type: multi_turn
+    target_concurrency: 8
+
+  # Mandatory: with the default warmup behaviour, every request fails with
+  # ConnectionResetError because uvicorn closes pre-warmed idle sockets after 5s.
+  client:
+    warmup_connections: 0
```
viraatc, line 36: whats the http error code ur seeing on this?

dont remember seeing this in the past with sglang.
(i expect tcp-keepalive is disabled (default))

also max-idle-time should is likely not making a difference here, lets use default if we can.

src/inference_endpoint/load_generator/multi_turn_strategy.py
```diff
@@ -132,8 +131,9 @@ def __init__(
 
         # Event-driven state — populated in execute().
         self._pending_convs: deque[tuple[str, list[tuple[int, int]]]] = deque()
-        self._active_iters: dict[str, Iterator[tuple[int, int]]] = {}
+        self._active_iters: dict[str, tuple[list[tuple[int, int]], int]] = {}
```
viraatc: lets add a typedef/alias, hard to understand what `tuple[list[tuple[int, int]], int]` is supposed to mean here and in `self._pending_convs` as well?

src/inference_endpoint/dataset_manager/factory.py
```diff
@@ -110,10 +110,14 @@ def create_loader(config: DatasetConfig, num_repeats: int = 1, **kwargs) -> Data
...
-        return Dataset.load_from_file(
+        dataloader = Dataset.load_from_file(
             Path(dataset_path),
             transforms=transforms,
             format=format_enum,
             dataset_id=dataset_id,
             num_repeats=num_repeats,
         )
+        if config.multi_turn is not None and config.multi_turn.enable_salt:
+            assert isinstance(dataloader, MultiTurnDataset)
```
arekay-nv: [Claude] **high** (bug): `assert isinstance(dataloader, MultiTurnDataset)` raises `AssertionError` — not a user-facing error — when `enable_salt=True` but the loader is not a `MultiTurnDataset`. Worse, the assert is silently skipped under `python -O`, leading to `AttributeError` on `dataloader.enable_salt()` with no diagnostic. The project convention for config-validation failures is `InputValidationError`. Replace with:

```python
if not isinstance(dataloader, MultiTurnDataset):
    raise InputValidationError(
        "multi_turn.enable_salt requires a multi-turn dataset, got "
        f"{type(dataloader).__name__}"
    )
```

src/inference_endpoint/load_generator/multi_turn_strategy.py
```diff
@@ -210,22 +213,48 @@ async def execute(self, phase_issuer: PhaseIssuerProtocol) -> int:
...
+        ):
+            delay_map = self._dataset_metadata.delay_seconds_by_key
+            delay = float(delay_map.get((conv_id, turn), 0.0))
+
+        if delay > 0.0:
+            assert self._loop is not None
+            handle = self._loop.call_later(
+                delay, self._issue_turn_now, conv_id, idx, turn
+            )
+            self._delay_handles[conv_id] = handle
```
arekay-nv, line 259: [Claude] **high** (concurrency): The early-exit guard in `execute()` (a few lines before this hunk) reads `if not self._active_iters and not self._inflight`. This guard fires incorrectly when every conversation has its *first* turn delayed: `_issue_next_turn` advances the cursor and stores the handle here, but returns before populating `_inflight`. If all `_active_iters` slots are occupied by pending-delay conversations, the guard sees `_inflight` empty and never sets `_all_done`, causing `execute()` to hang at `await self._all_done.wait()`. The guard should also check `not self._delay_handles`:

```python
if not self._active_iters and not self._inflight and not self._delay_handles:
    return phase_issuer.issued_count
```

Similarly, `_fill_slot` (further below) checks `elif not self._active_iters` to set `_all_done` — this will fire prematurely while delays are pending, terminating the run before delayed turns execute.

src/inference_endpoint/dataset_manager/multi_turn_dataset.py
```diff
@@ -417,6 +435,7 @@ def _build_metadata(self) -> ConversationMetadata:
                         "name",
                         "tool_calls",
                         "tool_results",
+                        "reasoning_content",
```
arekay-nv, line 448: [Claude] **medium** (api-contract): `reasoning_content` is unconditionally forwarded in prior-assistant-turn history messages. This field is a Kimi/SGLang-specific extension and is **not** part of the OpenAI Chat Completions API spec for *request* message objects. Any server that validates input strictly (standard vLLM, OpenAI-compatible endpoints, TRT-LLM serve) will reject requests with a 400/422 error when prior assistant turns carry this field.

Since `ChatMessage.reasoning_content` has `omit_defaults=True` and defaults to `None`, the field is only included when the dataset row has a non-null value — so this only triggers for datasets recorded from Kimi runs. But there is no `api_type` gate or config option to strip it, and no documentation warning. Recommend gating on `api_type in (APIType.OPENAI_MSGSPEC, APIType.SGLANG)` or adding a `strip_reasoning_from_history` option to `MultiTurnConfig`.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+            uuid = ev.get("sample_uuid") or ""
+            if not uuid:
+                continue
+            key = uuid_to_key.get(uuid)
+            if key is None:
+                continue
+            data = _parse_model_output(ev.get("data"))
+            if data is None:
+                continue
+            conv_id, client_turn = key
```
arekay-nv: [Claude] **medium** (data-integrity): `derive_model_assistants` assigns assistant turn number as `client_turn + 1` (the turn number of the preceding user/tool row + 1). This works only because `_validate_turn_numbering` enforces that every client turn is immediately followed by an assistant turn at the next sequential turn number. That invariant is not visible at this call site.

If a caller passes a JSONL file from a non-validating source (e.g. a manually constructed or future-format dataset with non-contiguous turn numbers), all turns are silently scored against the wrong ground-truth rows — `missing_in_model` will equal the full dataset size and pass rate will be 0 with no error or warning. A comment citing the coupling to `_validate_turn_numbering`, or a runtime assertion that `client_turn + 1` exists in the gt index, would prevent silent misscoring.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+        default=None,
+        help="Key in sample_idx_map.json (only needed if multiple datasets in one run)",
+    )
+    p.add_argument("--out", type=Path)
+    args = p.parse_args()
+
+    if args.report_dir is not None:
+        model_path = args.report_dir / "model_assistants.jsonl"
+        derive_model_assistants(args.report_dir, args.gt, model_path, args.dataset_name)
+    else:
```
arekay-nv: [Claude] **low** (error-handling): When `--model` is passed directly (not `--report-dir`), `model_path = args.model` is used without checking whether the file exists. `score_run` then calls `_iter_assistant_turns(model_jsonl)` which raises a raw `FileNotFoundError` with a Python traceback. The `--report-dir` path is consistent but the `--model` path should add the same guard:

```python
if not model_path.exists():
    sys.exit(f"model file not found: {model_path}")
```

arekay-nv (commented):

Claude-only review (Codex unavailable — bwrap/pivot_root not permitted in this environment). Depth: **thorough**.

src/inference_endpoint/dataset_manager/factory.py
arekay-nv: [Claude] **high** (bug): `assert isinstance(dataloader, MultiTurnDataset)` raises `AssertionError` — not a user-facing error — when `enable_salt=True` but the loader is not a `MultiTurnDataset`. Worse, the assert is silently skipped under `python -O`, leading to `AttributeError` on `dataloader.enable_salt()` with no diagnostic. The project convention for config-validation failures is `InputValidationError`. Replace with:

```python
if not isinstance(dataloader, MultiTurnDataset):
    raise InputValidationError(
        "multi_turn.enable_salt requires a multi-turn dataset, got "
        f"{type(dataloader).__name__}"
    )
```

src/inference_endpoint/load_generator/multi_turn_strategy.py
arekay-nv, line 259: [Claude] **high** (concurrency): The early-exit guard in `execute()` (a few lines before this hunk) reads `if not self._active_iters and not self._inflight`. This guard fires incorrectly when every conversation has its *first* turn delayed: `_issue_next_turn` advances the cursor and stores the handle here, but returns before populating `_inflight`. If all `_active_iters` slots are occupied by pending-delay conversations, the guard sees `_inflight` empty and never sets `_all_done`, causing `execute()` to hang at `await self._all_done.wait()`. The guard should also check `not self._delay_handles`:

```python
if not self._active_iters and not self._inflight and not self._delay_handles:
    return phase_issuer.issued_count
```

Similarly, `_fill_slot` (further below) checks `elif not self._active_iters` to set `_all_done` — this will fire prematurely while delays are pending, terminating the run before delayed turns execute.

src/inference_endpoint/dataset_manager/multi_turn_dataset.py
arekay-nv, line 448: [Claude] **medium** (api-contract): `reasoning_content` is unconditionally forwarded in prior-assistant-turn history messages. This field is a Kimi/SGLang-specific extension and is **not** part of the OpenAI Chat Completions API spec for *request* message objects. Any server that validates input strictly (standard vLLM, OpenAI-compatible endpoints, TRT-LLM serve) will reject requests with a 400/422 error when prior assistant turns carry this field.

Since `ChatMessage.reasoning_content` has `omit_defaults=True` and defaults to `None`, the field is only included when the dataset row has a non-null value — so this only triggers for datasets recorded from Kimi runs. But there is no `api_type` gate or config option to strip it, and no documentation warning. Recommend gating on `api_type in (APIType.OPENAI_MSGSPEC, APIType.SGLANG)` or adding a `strip_reasoning_from_history` option to `MultiTurnConfig`.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
arekay-nv: [Claude] **medium** (data-integrity): `derive_model_assistants` assigns assistant turn number as `client_turn + 1` (the turn number of the preceding user/tool row + 1). This works only because `_validate_turn_numbering` enforces that every client turn is immediately followed by an assistant turn at the next sequential turn number. That invariant is not visible at this call site.

If a caller passes a JSONL file from a non-validating source (e.g. a manually constructed or future-format dataset with non-contiguous turn numbers), all turns are silently scored against the wrong ground-truth rows — `missing_in_model` will equal the full dataset size and pass rate will be 0 with no error or warning. A comment citing the coupling to `_validate_turn_numbering`, or a runtime assertion that `client_turn + 1` exists in the gt index, would prevent silent misscoring.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
arekay-nv: [Claude] **low** (error-handling): When `--model` is passed directly (not `--report-dir`), `model_path = args.model` is used without checking whether the file exists. `score_run` then calls `_iter_assistant_turns(model_jsonl)` which raises a raw `FileNotFoundError` with a Python traceback. The `--report-dir` path is consistent but the `--model` path should add the same guard:

```python
if not model_path.exists():
    sys.exit(f"model file not found: {model_path}")
```

arekay-nv (commented):

Claude-only review (Codex unavailable — bwrap/pivot_root not permitted in this environment). Depth: **thorough**.

src/inference_endpoint/dataset_manager/factory.py
arekay-nv: [Claude] **high** (bug): `assert isinstance(dataloader, MultiTurnDataset)` raises `AssertionError` when `enable_salt=True` but the loader is not a `MultiTurnDataset`. The assert is silently skipped under `python -O`, leading to `AttributeError` on `dataloader.enable_salt()` with no diagnostic. The project convention for config-validation failures is `InputValidationError`:

```python
if not isinstance(dataloader, MultiTurnDataset):
    raise InputValidationError(
        "multi_turn.enable_salt requires a multi-turn dataset, got "
        f"{type(dataloader).__name__}"
    )
```

src/inference_endpoint/load_generator/multi_turn_strategy.py
arekay-nv, line 259: [Claude] **high** (concurrency): The early-exit guard in `execute()` reads `if not self._active_iters and not self._inflight`. When every conversation has its first turn delayed, `_issue_next_turn` advances the cursor and stores the handle here, but returns before populating `_inflight`. The guard then sees `_inflight` empty and `_all_done` is never set, causing `execute()` to hang at `await self._all_done.wait()`. Fix: also check `not self._delay_handles`.

The same issue affects `_fill_slot`: `elif not self._active_iters` fires prematurely while delays are pending, terminating the run before delayed turns execute.

src/inference_endpoint/dataset_manager/multi_turn_dataset.py
arekay-nv, line 448: [Claude] **medium** (api-contract): `reasoning_content` is unconditionally forwarded in prior-assistant-turn history messages. This is a Kimi/SGLang-specific extension, not in the OpenAI Chat Completions spec for request message objects. Servers that validate input strictly (standard vLLM, TRT-LLM serve) will reject with 400/422 when prior assistant turns carry this field. Since `ChatMessage.reasoning_content` defaults to `None` and uses `omit_defaults=True`, this only triggers for datasets recorded from Kimi runs — but there is no `api_type` gate, config option to strip it, or documentation warning.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
arekay-nv: [Claude] **medium** (data-integrity): `client_turn + 1` works only because `_validate_turn_numbering` enforces sequential turn numbering — an invariant not visible here. If called with a non-validating source (manually constructed dataset, future format with non-contiguous turn numbers), all turns are silently scored against the wrong ground-truth rows: `missing_in_model` equals the full dataset size, pass rate is 0, with no error. A comment or runtime assertion would prevent silent misscoring.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
arekay-nv: [Claude] **low** (error-handling): When `--model` is passed directly, `model_path = args.model` is used without checking existence. `score_run` raises a raw `FileNotFoundError` with a traceback. The `--report-dir` path is consistent — the `--model` path needs the same existence check before use.

arekay-nv:

Reviewed by: **Claude** (Codex unavailable — bwrap/pivot_root not permitted in this environment) | Depth: **thorough** | PR: #314 "Agentic Inference updates"

Found **5 new issues** posted as inline comments (3 previously covered by earlier review pass were deduplicated). Commit hygiene: 14 commits, 6 chore/fix — consider squashing before merge.

| 1 | `factory.py` | 121 | bug | `assert isinstance(...)` → must be `InputValidationError`; silently skipped under `python -O` |
| 2 | `multi_turn_strategy.py` | 247 | concurrency | `execute()` early-exit + `_fill_slot` completion check both missing `not self._delay_handles`; first-turn-delay scenario can hang or terminate early |

### 🟡 Should Fix (medium)

| 3 | `multi_turn_dataset.py` | 438 | api-contract | `reasoning_content` unconditionally forwarded in history messages; breaks non-Kimi endpoints (vLLM, TRT-LLM) with 400/422 |
| 4 | `score_inline_accuracy.py` | 461 | data-integrity | `client_turn + 1` implicitly couples to `_validate_turn_numbering` invariant; silent misscoring if invariant not enforced upstream |

### 🔵 Consider (low)

| 5 | `score_inline_accuracy.py` | 507 | error-handling | `--model` path has no existence check; raises raw `FileNotFoundError` traceback vs `--report-dir` path's clean error |

> ⚠️ **Commit hygiene**: This PR has 14 commits including 6 apparent chore/fix commits. Consider squashing before merge.

examples/09_MultiTurn/README.md
```diff
@@ -1,297 +1,109 @@
...
-highest-indexed snapshot per conversation is used.
+Place the dataset under `examples/09_MultiTurn/datasets/` or point the YAML at
+another accessible JSONL path.
 
-### Converting to flat-row format
+## Start A Server
 
-The following commands convert each source snapshot file to the flat-row format required by the benchmarker.
-Run from the repo root:
+Start an SGLang OpenAI-compatible server. This is the standard non-deterministic
```
arekay-nv: what does non-deterministic mean in this context?

examples/09_MultiTurn/README.md
```diff
@@ -1,297 +1,109 @@
...
-## Overview
+## Dataset
 
-Multi-turn conversation benchmarking enables testing realistic conversational AI scenarios where each turn depends on previous responses. The system maintains conversation history and enforces turn sequencing to simulate real-world multi-turn interactions.
-
-## Dataset Format
-
-Multi-turn datasets use JSONL format with the following structure:
+Use flat JSONL with one row per message. Rows for each `conversation_id` must be
+contiguous.
```
arekay-nv: and ordered sequentially in increasing turn order.

examples/09_MultiTurn/README.md
```diff
@@ -1,297 +1,109 @@
...
+  --tool-call-parser kimi_k2 \
+  --host 0.0.0.0 \
+  --port 8000
 ```
 
-### Running agentic benchmarks
-
-After converting the datasets, update the `path` field in the config files and run:
+The default YAML assumes the server is reachable at `http://localhost:8000` and
+the served model name is `/model`.
```
arekay-nv: Does `/model` work as a model name? Probably not something you'd want to do as the client should only care about the model name, not the path. Best to document an example with a separate model name.

examples/09_MultiTurn/README.md
```diff
@@ -1,297 +1,109 @@
...
+  reasoning preservation.
+- `datasets[0].name`: label used in benchmark outputs.
+- `datasets[0].type: performance`: multi-turn datasets are replayed as
+  performance datasets.
+- `datasets[0].path`: JSONL dataset path to run.
+- `datasets[0].multi_turn.turn_timeout_s`: per-turn deadline. A timeout aborts
+  the remaining turns in that conversation.
+- `datasets[0].multi_turn.enable_salt`: appends a deterministic cache salt to
+  each conversation system prompt.
+- `datasets[0].multi_turn.inject_tool_delay`: honors positive `delay_seconds`
```
arekay-nv: `datasets[0]` should be replaced with the `first datasets section entry in the yaml`.

arekay-nv (changes requested):
Can you refactor to reuse the existing accuracy infrastructure. This will enable us to reuse the same setup across models/datasets.

viraatc (commented):

Reviewed by: **Claude** | Depth: **thorough**

Codex review failed (CLI/cloud-requirements compatibility issue after recent CLI upgrade, then internal bwrap hangs on Ubuntu 24 AppArmor restriction). Falling back to Claude-only.

7 issues posted inline after deduping against the 44 existing comments and re-verifying each finding against current HEAD. Critical Pandas dict-assignment bug verified by repro. Schema gap on `delay_seconds` verified empirically.

src/inference_endpoint/openai/openai_msgspec_adapter.py
```diff
@@ -79,6 +80,7 @@ def dataset_transforms(cls, model_params: ModelParams) -> list[Transform]:
             "top_p": model_params.top_p,
             "top_k": model_params.top_k,
             "repetition_penalty": model_params.repetition_penalty,
+            "chat_template_kwargs": model_params.chat_template_kwargs,
```
viraatc, line 85: [Claude] **critical** (bug): `chat_template_kwargs` is added to the `metadata` dict that is later passed to `AddStaticColumns(metadata)` at line 108. When `chat_template_kwargs` is a dict (e.g. `{thinking: true, preserve_thinking: true}` from `kimi_agentic_benchmark.yaml`), Pandas interprets `df['chat_template_kwargs'] = {'thinking': True, ...}` as a column-name→value mapping, not a broadcast scalar — the column silently becomes all NaN. Verified empirically: `df['c'] = {'k': True}` produces `[nan, nan, ...]`. `to_endpoint_request` then reads `query.data.get('chat_template_kwargs')` as NaN and omits the field, so the YAML setting never reaches the server. Unit tests miss this because they inject `chat_template_kwargs` directly into `Query.data` and bypass the transform. Fix: assign as a list (`df[key] = [value] * len(df)`) or wrap dict values before passing to `AddStaticColumns`, and add a regression test that calls the transform's `__call__` on a real DataFrame.

src/inference_endpoint/load_generator/session.py
```diff
@@ -228,7 +229,8 @@ def issue(
         data = self._dataset.load_sample(sample_index)
         if data_override is not None:
             data = {**data, **data_override}
-        query = Query(id=query_id, data=data)
+        headers = {_SESSION_ID_HEADER: conversation_id} if conversation_id else {}
```
viraatc, line 239: [Claude] **medium** (bug): `conversation_id` flows raw into the HTTP `X-Session-ID` header here, but `_validate_conversation_grouping` only checks non-emptiness — it doesn't reject CR/LF or other control characters. The HTTP request builder encodes headers via `f'{k}: {v}\r\n'`, so a dataset row with `conversation_id` containing `\n` will produce a malformed request (or, with `\r\n`, inject extra headers). Not exploitable by an outside attacker (datasets are user-supplied), but a benign typo or generator bug produces a confusing wire-protocol failure instead of a clean validation error. Validate `conversation_id` for printable ASCII / no CR-LF in `_validate_conversation_grouping`, or sanitize at the header-construction site.

scripts/multi_turn_dataset_schema.json
```diff
@@ -423,6 +423,11 @@
         "type": "array",
         "description": "Tool execution results (required for all 'tool' role rows). Each element has {tool_call_id, content}. Single results have one entry; parallel results have multiple entries."
       },
+      {
+        "field": "delay_seconds",
```
viraatc, line 432: [Claude] **medium** (data-integrity): `delay_seconds` is only described in this human-facing `optionalFields` documentation block — no actual schema property exists in `basicMessageTypes` (line 8), `toolCallMessage` (line 70), `toolMessage` (line 131), or `generationParameters` (line 173), and all four set `additionalProperties: true`. Empirically verified: feeding `{... 'delay_seconds': 'not-a-number'}` and `{... 'delay_seconds': -5}` to `validate_jsonl_schema.py` produces zero errors. At runtime, `multi_turn_dataset.py:487-490` does `try: float(delay_val); except: 0.0`, so malformed values silently degrade to "no delay" with no diagnostic. Add `delay_seconds: {"type": "number", "minimum": 0}` to the real schema definitions (or a shared `$ref`) so the validator catches malformed values up-front.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+        return None
+    return _multiset_iou(g, extract_exes_from_turn(model))
+
+
+# ---------------------------------------------------------------------------
+# Workflow scorer - extract `intent: IXXX` code, exact match
+# ---------------------------------------------------------------------------
+
+_INTENT_RE = re.compile(r"\bintent:\s*(I\d{3})\b", re.IGNORECASE)
+_BARE_INTENT_RE = re.compile(r"\bI(\d{3})\b")
```
viraatc: [Claude] **medium** (bug): `_INTENT_RE` (line 194) is compiled with `re.IGNORECASE`, but `_BARE_INTENT_RE` here is not. When the explicit `intent:` prefix is absent, the bare fallback matches only uppercase `I\d{3}`; a model that reasons in lowercase (`"the answer is i057 (delay)"`) falls through both branches and `_extract_intent_code` returns `None`, scoring 0.0 even when the intent is correct. Add `re.IGNORECASE` here and normalize the captured digits to upper-case at the return site (line 217) to match the explicit-form behavior already in place at line 199.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+        n_zero            - turns scoring exactly 0.0
+        skipped_unscorable_gt - gt had no intent (workflow) or no exes (coding)
+        missing_in_model  - gt scorable but model has no matching turn
+        per_turn          - list of {conv, turn, score}
+    """
+    if domain not in ("coding", "workflow"):
+        raise ValueError("domain must be 'coding' or 'workflow'")
+    scorer = score_coding if domain == "coding" else score_workflow
+
+    gt_by_key = {_key(t): t for t in _iter_assistant_turns(gt_jsonl)}
```
viraatc: [Claude] **medium** (data-integrity): `gt_by_key = {_key(t): t for t in _iter_assistant_turns(gt_jsonl)}` (and the model dict on the next line) silently overwrites entries with duplicate `(conv_id, turn)` keys. If the gt or model JSONL ever contains a duplicate row (concatenated dataset, retry-with-append bug, off-by-one in an exporter), only the last write is scored and earlier rows vanish from `n_scorable`. The resulting `pass_rate` looks clean. Detect duplicates explicitly (raise or log a warning, or accumulate `list[dict]` and reject) so a silently-truncated input doesn't masquerade as a successful score.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
viraatc: [Claude] **medium** (design): `pass_rate = total / n` where `n = len(results)` excludes turns the model never produced (the `missing_in_model` branch at line 311 `continue`s without appending to `results`), but *includes* turns the model returned with 0.0 score. A backend that drops 90% of requests can post a high `pass_rate` by scoring well on the 10% that succeed, while a backend that always responds gets penalized for low-quality answers. Either count `missing_in_model` as 0.0 in the denominator (conservative reading: a missing answer is a failed answer), or document the asymmetry in the function docstring so users don't compare numbers across runs with different drop rates.

examples/09_MultiTurn/accuracy/score_inline_accuracy.py
```diff
@@ -0,0 +1,525 @@
...
+_ENVKV_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=")
+_PATH_LEAF = re.compile(r"[^/]+$")
+_PYVER_RE = re.compile(r"\.\d+(\.\d+)?$")
+
+
+def _canonicalize_stage(stage: str) -> str | None:
+    """Return the whitelisted canonical exe of a single pipeline stage, or None."""
+    tokens = stage.split()
+    i = 0
+    while i < len(tokens) and (_ENVKV_RE.match(tokens[i]) or tokens[i] in _WRAPPERS):
```
viraatc: [Claude] **low** (bug): `tokens[i] in _WRAPPERS` is case-sensitive, but the leaf comparison on line 132 lowercases. So `"SUDO ls"` doesn't strip the `SUDO` wrapper (it isn't in `{env, time, nice, sudo, exec, command}`), then `_PATH_LEAF.search('SUDO').group(0).lower()` returns `'sudo'`, which is not in `EXE_MAP` — the real `ls` is silently missed. Same for `"Env ls"`. Lowercase `tokens[i]` for the wrapper / envkv membership check (or apply `.lower()` to all tokens up-front) so canonicalization is consistent end-to-end.

viraatc:
## Review Council — Summary

Reviewed by: **Claude** (Codex unavailable: post-upgrade cloud-requirements compatibility + Ubuntu 24 AppArmor `bwrap` lockup) | Depth: **thorough**

Found **7** issues across **4** files. All findings re-verified against HEAD (`56db1b21`) and deduped against the 48 existing inline comments.

| File | Line | Category | Summary |
|------|-----:|----------|---------|
| `openai/openai_msgspec_adapter.py` | 83 | bug | `chat_template_kwargs` dict silently becomes NaN through `AddStaticColumns` (Pandas dict-assignment); YAML setting never reaches the server. Repro'd. |

### 🟡 Should Fix (medium)

| File | Line | Category | Summary |
|------|-----:|----------|---------|
| `load_generator/session.py` | 232 | bug | `conversation_id` flows raw into `X-Session-ID` header; no CR/LF validation. |
| `scripts/multi_turn_dataset_schema.json` | 427 | data-integrity | `delay_seconds` only documented, not in any schema definition; validator accepts `"not-a-number"` and `-5`. Verified. |
| `accuracy/score_inline_accuracy.py` | 195 | bug | `_BARE_INTENT_RE` missing `re.IGNORECASE` (sibling `_INTENT_RE` has it). |
| `accuracy/score_inline_accuracy.py` | 298 | data-integrity | `gt_by_key`/`model_by_key` silently overwrite duplicate `(conv,turn)` keys. |
| `accuracy/score_inline_accuracy.py` | 328 | design | `pass_rate` excludes `missing_in_model` from denominator → high-drop backends score artificially well. |

### 🔵 Consider (low)

| File | Line | Category | Summary |
|------|-----:|----------|---------|
| `accuracy/score_inline_accuracy.py` | 128 | bug | `_WRAPPERS` case-sensitive but leaf is `.lower()`-ed → `"SUDO ls"` silently misses `ls`. |

Inline thread: https://github.com/mlcommons/endpoints/pull/314#pullrequestreview-4332438939

src/inference_endpoint/dataset_manager/multi_turn_dataset.py
hvagadia: This is intentional for Kimi replay. The dataset has `reasoning_content` because SGLang/TRT-LLM with the `kimi_k2` reasoning parser returns thinking separately from final `content`.

For replay, Kimi needs prior thinking in history; with `chat_template_kwargs.preserve_thinking: true`, the chat template renders `reasoning_content` back into the prompt. We’ve verified this path with SGLang and TRT-LLM.

hvagadia:
I have removed the accuracy script for now. It is still under discussion in TF, so doesnt make sense to integrate it yet. I will put up a separate draft PR for folks to experiment.

examples/09_MultiTurn/README.md
hvagadia: Removed that term, was an artifact of earlier experiments.

examples/09_MultiTurn/README.md
hvagadia: Clarified better in the readme.

examples/09_MultiTurn/kimi_agentic_benchmark.yaml
hvagadia: Let me get back on the error. I got errors while back, need to test on latest code.

tianmu-li: To add, I see connection error that requires setting warmup_connections to 0 when agents try to run the benchmark, but I don't see the error when I run it myself.

viraatc: ive seen the sandbox in claudecode/codex can interfere with server/client side applications - iiuc its because agents tend to run a lot of commands, end up spamming the system

warmup-connection feature assumes exclusive access to system at runtime,
its an acceptable solution to disable warmup_connections by setting 0 for agents / parallel runs.

src/inference_endpoint/core/types.py
```diff
@@ -78,6 +78,37 @@ class QueryStatus(Enum):
 OUTPUT_ELEM_TYPE = str | tuple[str, ...]
 """Type for a single output or reasoning value: string (non-streaming) or tuple of strings (streaming)."""
 
+TOOL_CALL_TYPE = dict[str, Any]
```
viraatc, line 81: we dont want to export these 🤔 

feel free to revert-inline, or use `_` prefix.
also 3.12 has adds new `type` keyword for typedefs if u want to keep it:
```
type _TOOL_CALL_ELEMENT_T = tuple[Any, ...]
type _TOOL_CALL_CHUNKS_TYPE = tuple[TOOL_CALLS_TYPE, ...]
```

src/inference_endpoint/dataset_manager/multi_turn_dataset.py
tianmu-li: Can you also add "reasoning" field and use the same content as "reasoning_content"? vllm expects "reasoning" instead of "reasoning_content", so the current version causes vllm to silently drop reasoning content in history.

src/inference_endpoint/config/schema.py
```diff
@@ -253,6 +257,8 @@ class MultiTurnConfig(BaseModel):
 
     turn_timeout_s: float = Field(default=300.0, gt=0)
     use_dataset_history: bool = True
+    enable_salt: bool = False
+    inject_tool_delay: bool = False
```
arekay-nv: Please add a comment on these two fields. The salt can be confusing with the warmup salt, and the tool delay could be confusing for workflow datasets.

arekay-nv (approved):
Made a quick pass, but looks good.

src/inference_endpoint/dataset_manager/multi_turn_dataset.py
hvagadia: The dataset also has field named reasoning_content, does that need to be named reasoning for vllm?

hvagadia: @tianmu-li will tackle this in separate PR if needed.

src/inference_endpoint/core/types.py
hvagadia: Done

src/inference_endpoint/config/schema.py
hvagadia: Done

PR #315, opened 2026-05-18
feat(wan22): add VBench accuracy scorer for WAN 2.2 video generation

wu6u3tw:
## Summary

- Adds `VBenchScorer` (`scorer_id="vbench"`) that scores video-generation outputs on six VBench dimensions (subject_consistency, background_consistency, motion_smoothness, dynamic_degree, appearance_style, scene) and returns the mean of the per-dimension aggregates.
- Runs VBench in an isolated `uv` subproject under `examples/09_Wan22_VideoGen_Example/accuracy/` (vbench pins `transformers==4.33.2` + `numpy<2`, incompatible with the parent env). `VBenchScorer.score()` shells out to `vbench_runner.py` via `uv run --project`, so the benchmark process never imports vbench.
- `VideoGenAdapter.decode_response` now mirrors `video_path` into `response_output` so the event log carries it to the scorer.
- Wires `ScorerMethod.VBENCH` into the config enum and adds `offline_wan22_accuracy.yaml` as a peer to the existing perf example.

## Test plan

- [x] Unit: `tests/unit/evaluation/test_scoring.py` covers registration, mean-of-6-dims with mocked `subprocess.run`, missing-subproject `FileNotFoundError`.
- [x] Unit: `tests/unit/videogen/test_adapter.py` + integration adapter test updated for the new `response_output` contract.
- [x] `examples/09_Wan22_VideoGen_Example/offline_wan22_accuracy.yaml` validates as `OfflineBenchmarkConfig`.
- [x] Subproject `uv lock` resolves (115 packages).
- [x] `pre-commit run --all-files` clean.
- [ ] End-to-end VBench run on a GPU host with `uv sync` in the accuracy subproject — not yet executed.

Draft for code review.


AGENTS.md
```diff
@@ -85,18 +85,18 @@ Dataset Manager --> Load Generator --> Endpoint Client --> External Endpoint
...
+| **VideoGen**           | `src/inference_endpoint/videogen/`                                | Adapter for video-generation endpoints (e.g. trtllm-serve `POST /v1/videos/generations`, used by MLPerf WAN2.2-T2V-A14B). Defaults to `response_format=video_path` (server saves video to shared storage and returns path) to avoid large byte payloads. Accuracy mode also runs on `video_path`: the adapter mirrors the path into `response_output` so the event log carries it to `VBenchScorer` (see `evaluation/scoring.py`), which scores videos via VBench from a sibling `uv` subproject at `examples/09_Wan22_VideoGen_Example/accuracy/` (vbench's `transformers==4.33.2` + `numpy<2` pins are incompatible with the parent env, so it runs out-of-process via `uv run --project`). Dataset is ingested via the generic JSONL loader. |
```
viraatc, line 100: should we update .gitignore as well to include the expected subproject artefact files?

viraatc (approved):
lgtm, thanks!

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,186 @@ def score(self) -> tuple[float, int]:
...
+        Indexing is per-prompt to disambiguate when the same prompt appears
+        multiple times (num_repeats > 1). Stale symlinks from a prior run
+        are removed first so re-scoring an existing report_dir is safe.
+        """
+        per_prompt_idx: dict[str, int] = defaultdict(int)
+        for video_path, prompt in zip(video_paths, prompts, strict=True):
+            idx = per_prompt_idx[prompt]
+            per_prompt_idx[prompt] += 1
+            src = Path(video_path)
+            dst = staged_dir / f"{prompt}-{idx}{src.suffix or '.mp4'}"
```
viraatc: [Claude] **critical** (security): `_stage_videos` builds `f"{prompt}-{idx}{src.suffix or '.mp4'}"` from the dataset's prompt column without sanitization. Beyond the `/` case @gemini-code-assist flagged at L951, a prompt containing `..` (e.g. `"../../etc/passwd"`) escapes `staged_dir` entirely, then `dst.unlink(missing_ok=True)` deletes whatever exists at that resolved path before `dst.symlink_to(...)` plants a symlink outside the report directory. Datasets are user-supplied JSONL, so this is exploitable by any benign typo or hostile dataset. Reject `..` segments and slashes before constructing the path — hash the prompt or use the per-row UUID as the staged filename and keep a side-map for VBench's lookup.

examples/09_Wan22_VideoGen_Example/offline_wan22_accuracy.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+model_params:
+  name: "wan22"
+  max_new_tokens: 1
+  streaming: "off"
+
+datasets:
+  - name: wan22_vbench
+    path: examples/09_Wan22_VideoGen_Example/wan22_prompts.jsonl
+    type: "accuracy"
```
viraatc, line 43: [Codex] **high** (bug): As written this YAML cannot make it past `setup_benchmark()` — `execute._load_datasets()` rejects configs with no performance dataset and raises `InputValidationError("At least one performance dataset required")` before any requests are issued. A single `type: "accuracy"` entry is rejected at validation. Either add a matching `type: "performance"` phase pointing at the same JSONL, or add explicit support for accuracy-only runs before documenting this config as the user-facing example.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,186 @@ def score(self) -> tuple[float, int]:
...
+    def score(self) -> tuple[float, int]:
+        df = self.get_outputs()
+        valid_uuids = self.sample_index_map.keys()
+        df = df[df["sample_uuid"].isin(valid_uuids)]
+        # Drop failed queries: Scorer.get_outputs() emits "" when record.data
+        # is None (workers set response_output=None on error). Passing "" to
+        # _stage_videos would Path("").resolve() → cwd and symlink the repo
+        # root as a "video", corrupting the entire VBench run.
+        n_total = len(df)
+        df = df[df["output"].astype(bool)]
```
viraatc, line 1110: [Codex+Claude] **high** (data-integrity): The filter that drops empty `video_path`s silently removes failed generations from the VBench score. Two downstream symptoms: (1) `mean_score = float(np.mean(per_dim_scores))` is computed over only the successful samples — accuracy is overstated exactly when failures matter; (2) `n_repeats = len(video_paths) // self.dataset.num_samples()` (L1029) integer-divides on the *filtered* set, so even one failure on a single-repeat run reports `n_repeats == 0`; (3) `np.mean([])` returns NaN with a RuntimeWarning (per the type annotation at L977 the return is `tuple[float, int]`, but NaN slips through and json-serializes as the bare `NaN` token). Either count failed generations as 0-score in the mean (conservative reading: a missing video is a failed video), or make scoring fail loudly. Match the base class's `tuple[float | None, int]` contract and return `(None, n_repeats)` on empty/all-failed, like `LiveCodeBenchScorer`.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,186 @@ def score(self) -> tuple[float, int]:
...
+        self._run_vbench_subprocess(staged_dir, vbench_out, run_name)
+
+        # VBench writes `{run_name}_eval_results.json` to vbench_out. Each
+        # dim entry is `[aggregate_score, [per_video_results, ...]]`.
+        results_path = vbench_out / f"{run_name}_eval_results.json"
+        with results_path.open() as f:
+            results = msgspec.json.decode(f.read())
+
+        per_dim_scores: list[float] = [
+            float(results[dim][0]) for dim in self.dimensions
```
viraatc: [Claude] **high** (bug): `[float(results[dim][0]) for dim in self.dimensions]` raises `KeyError` if VBench omits a requested dimension (and `IndexError` on malformed entries). RUNBOOK §5 already lists `KeyError: 'scene'` as a known failure mode (caused when the WAN 2.2 prompt set doesn't overlap with VBench's `scene` suite). Today this crashes the entire benchmark finalize step after a 30-minute scoring run and loses all artifacts. Either skip missing dims with a warning, fall back to a partial score, or fail fast with a clear message naming the missing dim — but don't let the natural Python exception propagate from inside an external dep's output dict.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,186 @@ def score(self) -> tuple[float, int]:
...
+            "--out-dir",
+            str(vbench_out),
+            "--name",
+            run_name,
+            "--dims",
+            ",".join(self.dimensions),
+        ]
+        if self.full_info_json_path is not None:
+            cmd += ["--full-info-json", self.full_info_json_path]
+        subprocess.run(cmd, check=True)
```
viraatc: [Claude] **high** (error-handling): `subprocess.run(cmd, check=True)` has no `timeout`, captures no output, and inherits the parent's stdin/stdout/stderr. Three problems: (1) VBench is heavyweight (~5 GB download on first run per RUNBOOK §0) and a single hung dim wedges the orchestrator indefinitely; (2) when `check=True` raises `CalledProcessError`, the only diagnostic is the return code — VBench's real error (CUDA OOM, missing HF model, decord failure) is lost to whatever stream it printed on; (3) `uv run`'s progress and HF tqdm interleave with `BenchmarkSession` status lines, and an interactive download prompt could hang on the parent's stdin. Capture stdout/stderr (or tee to `report_dir/vbench_subprocess.log`), pass `stdin=subprocess.DEVNULL`, add a configurable `timeout`, and include the captured stderr tail in the error message.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,186 @@ def score(self) -> tuple[float, int]:
...
+        """
+        per_prompt_idx: dict[str, int] = defaultdict(int)
+        for video_path, prompt in zip(video_paths, prompts, strict=True):
+            idx = per_prompt_idx[prompt]
+            per_prompt_idx[prompt] += 1
+            src = Path(video_path)
+            dst = staged_dir / f"{prompt}-{idx}{src.suffix or '.mp4'}"
+            dst.parent.mkdir(parents=True, exist_ok=True)
+            dst.unlink(missing_ok=True)
+            dst.symlink_to(src.resolve())
```
viraatc: [Claude] **high** (bug): `dst.symlink_to(src.resolve())` calls `Path.resolve()` with the default `strict=False`, so a missing or typo'd `src` silently produces a dangling symlink. VBench then fails opaquely with a decord/cv2 read error rather than at staging. Either pass `strict=True` (raises `FileNotFoundError` immediately, with the offending video path in the message) or precheck `src.is_file()`. Same symptom when `video_path` is an absolute path from a shared filesystem that's unmounted on this host — no diagnostic until VBench reads.

(Different concern from @gemini-code-assist's L951 comment about path-component sanitization.)

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,186 @@ def score(self) -> tuple[float, int]:
...
+        self, staged_dir: Path, video_paths: list[str], prompts: list[str]
+    ) -> None:
+        """Symlink each video into staged_dir as `{prompt}-{index}.mp4`.
+
+        Indexing is per-prompt to disambiguate when the same prompt appears
+        multiple times (num_repeats > 1). Stale symlinks from a prior run
+        are removed first so re-scoring an existing report_dir is safe.
+        """
+        per_prompt_idx: dict[str, int] = defaultdict(int)
+        for video_path, prompt in zip(video_paths, prompts, strict=True):
```
viraatc, line 1004: [Codex+Claude] **medium** (bug): `_stage_videos` indexes per-prompt as `per_prompt_idx[prompt] += 1`. If `score()` is called twice on the same `report_dir` (RUNBOOK explicitly supports re-scoring), the second call re-indexes from 0 and `dst.unlink(missing_ok=True)` replaces matching symlinks — fine if the video set is unchanged, but if the second run has *fewer* repeats than the first, stale `prompt-(M-1).mp4` from the first run survives in `vbench_videos/`. VBench walks the whole directory and scores those stale files. Same problem if two VBench datasets share a `report_dir`. Clear `staged_dir` entirely at the top of `_stage_videos` (or assert no leftover files match the per-prompt pattern beyond the current count) — the docstring's "re-scoring is safe" contract requires it.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,186 @@ def score(self) -> tuple[float, int]:
...
+        )
+        self.dimensions = dimensions
+        self.full_info_json_path = full_info_json_path
+        self.vbench_project_path = Path(
+            vbench_project_path
+            if vbench_project_path is not None
+            else _DEFAULT_VBENCH_PROJECT_PATH
+        )
+        self.uv_executable = uv_executable
+        runner = self.vbench_project_path / "vbench_runner.py"
```
viraatc, line 960: [Claude] **medium** (api-contract): `VBenchScorer.__init__` accepts `dimensions`, `full_info_json_path`, `vbench_project_path`, and `uv_executable`, but the only call site in `commands/benchmark/execute.py:680` passes only the standard four base args. None of these VBench-specific fields are exposed via `AccuracyConfig` in `config/schema.py`, so YAML-driven users cannot override the default project path or switch the python invocation without hard-editing the source. Either plumb these through `AccuracyConfig` (e.g. a `extra: dict[str, Any] | None` field forwarded as kwargs) or drop the parameters to avoid implying a contract that doesn't exist.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,186 @@ def score(self) -> tuple[float, int]:
...
+    bundled `vbench_full_info.json`.
+
+    The scorer reads each sample's video path from response_output (the
+    VideoGenAdapter mirrors `video_path` into `TextModelOutput.output`)
+    and the prompt from `dataset.dataframe[ground_truth_column]` — the
+    prompt is the VBench input, not a comparison target, so callers should
+    set `ground_truth_column: prompt` in `accuracy_config`.
+    """
+
+    DIMENSIONS: ClassVar[tuple[str, ...]] = _VBENCH_DIMENSIONS
```
viraatc, line 928: [Claude] **medium** (api-contract): `commands/benchmark/execute.py:243-250` enforces `acc_cfg.accuracy_config.extractor is not None` for every accuracy dataset, but `VBenchScorer.score_single_sample` is never called — the extractor is genuinely unused. The example YAML at `offline_wan22_accuracy.yaml:45` works around this by setting `extractor: "identity_extractor"`, which is a lie about the contract: the field is mandatory at the validator and ignored at the scorer. Make `extractor` conditionally required (or drop it for scorers like `vbench` and `code_bench_scorer` that bypass extraction) so the YAML doesn't have to fabricate a meaningless extractor name.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,186 @@ def score(self) -> tuple[float, int]:
...
+            "--name",
+            run_name,
+            "--dims",
+            ",".join(self.dimensions),
+        ]
+        if self.full_info_json_path is not None:
+            cmd += ["--full-info-json", self.full_info_json_path]
+        subprocess.run(cmd, check=True)
+
+    def score(self) -> tuple[float, int]:
```
viraatc: [Claude] **medium** (testing): The new tests in `tests/unit/evaluation/test_scoring.py` mock `subprocess.run` to always succeed and write a perfect 6-dim JSON. None of the realistic failure modes are covered: (1) non-zero subprocess exit, (2) malformed `_eval_results.json`, (3) missing dimension key (RUNBOOK §5 explicit failure), (4) all samples dropped → empty `video_paths`, (5) prompts containing `/`, `..`, or null bytes, (6) `dst.symlink_to` with missing `src`. Add unit tests that monkeypatch `subprocess.run` to write a partial result dict and assert the scorer returns `None` (after the L1028 fix) rather than crashing the entire finalize phase.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,186 @@ def score(self) -> tuple[float, int]:
...
+_VBENCH_DIMENSIONS: tuple[str, ...] = (
+    "subject_consistency",
+    "background_consistency",
+    "motion_smoothness",
+    "dynamic_degree",
+    "appearance_style",
+    "scene",
+)
+
+_DEFAULT_VBENCH_PROJECT_PATH = (
```
viraatc, line 862: [Codex+Claude] **low** (design): `_DEFAULT_VBENCH_PROJECT_PATH = Path(__file__).resolve().parents[3] / "examples" / "09_Wan22_VideoGen_Example" / "accuracy"` only resolves to the right place from an editable/source checkout. The build config packages `src/` (`pyproject.toml` sets `module-root = "src"`), so a normal wheel install puts this module at `site-packages/inference_endpoint/evaluation/scoring.py` and `parents[3]` is `site-packages/`, where `examples/` does not exist — `VBenchScorer` raises `FileNotFoundError: vbench_runner.py not found at /…/site-packages/examples/…` before scoring starts. Since `ScorerMethod.VBENCH` is exposed in the shipped schema/templates, either bundle the subproject as package data, resolve via an env var (`VBENCH_PROJECT_PATH`), or document that the scorer is editable-install-only.

examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
```diff
@@ -0,0 +1,78 @@
...
+            "Optional path to a VBench_full_info.json file. When omitted, "
+            "falls back to the JSON bundled with the installed vbench "
+            "package — required for vbench_standard mode (which the "
+            "MLPerf WAN 2.2 prompt set, a subset of VBench's standard "
+            "suite, runs under)."
+        ),
+    )
+    args = parser.parse_args()
+
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
viraatc, line 106: [Claude] **low** (bug): `device = torch.device("cuda" if torch.cuda.is_available() else "cpu")` silently falls back to CPU on a GPU-less host. Per RUNBOOK §0, VBench's per-dim models (CLIP/DINO/RAFT/AMT) effectively require a GPU; on CPU a single dim can take hours and may exhaust system memory. Either fail fast with a clear error when `torch.cuda.is_available()` is False, or accept a `--require-cuda` CLI flag for production runs and emit a loud warning otherwise — silent CPU fallback in a heavyweight evaluator is a usability foot-gun.

viraatc (commented):

13 issues posted inline after re-verifying each finding against current HEAD and deduping against the 3 existing comments. Findings tagged `[Codex+Claude]` were flagged independently by both reviewers (boosted confidence).

Highest-priority: a path-traversal vector in `_stage_videos` (prompt-as-filename allows `..` escape of `staged_dir`, then `unlink` + `symlink_to` mutate paths outside `report_dir`). Also: failed video generations silently dropped from the VBench denominator (overstates accuracy), and the example YAML cannot actually run because the benchmark requires a performance dataset.

examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
```diff
@@ -0,0 +1,78 @@
...
+        _pkg_files(_vbench_pkg).joinpath("VBench_full_info.json")
+    )
+    vb = VBench(device, full_info_json, args.out_dir)
+    vb.evaluate(
+        videos_path=args.videos_dir,
+        name=args.name,
+        dimension_list=dimension_list,
+    )
+    return 0
+
```
viraatc, line 125: [Claude] **low** (error-handling): `vb.evaluate(...)` has no try/except. Any VBench exception (CUDA OOM, missing checkpoint, video read failure on one sample) propagates as a non-zero exit code with no JSON output and no structured error; the parent's `subprocess.run(check=True)` then raises `CalledProcessError` with only the return code (see companion finding at scoring.py:975). Wrap `vb.evaluate` and print a structured `{"status": "error", "message": ...}` line on stderr so the parent has something to surface.

viraatc:
## Review Council — Summary

Found **13** issues across **3** files. All findings re-verified against HEAD (`8b32e3a9`) and deduped against the 3 existing inline comments. `[Codex+Claude]` tags mark issues flagged independently by both reviewers.

| File | Line | Reviewer(s) | Category | Summary |
|------|-----:|-------------|----------|---------|
| `evaluation/scoring.py` | 948 | Claude | security | Path-traversal via `..` in prompt-as-filename (related to gemini's L951 `/` finding but strictly worse) |
| `offline_wan22_accuracy.yaml` | 36 | Codex | bug | Accuracy-only YAML cannot run — `setup_benchmark()` requires a performance dataset |
| `evaluation/scoring.py` | 986 | Codex+Claude | data-integrity | Failed generations dropped from VBench denominator → overstated accuracy + `n_repeats=0` + `NaN` mean |
| `evaluation/scoring.py` | 1026 | Claude | bug | `KeyError` on missing VBench dimension crashes finalize after ~30 min scoring |
| `evaluation/scoring.py` | 975 | Claude | error-handling | `subprocess.run` no timeout, no captured output, inherits stdio → hangs / opaque failures |
| `evaluation/scoring.py` | 951 | Claude | bug | `src.resolve(strict=False)` silently produces dangling symlinks |

### 🟡 Should Fix (medium)

| File | Line | Reviewer(s) | Category | Summary |
|------|-----:|-------------|----------|---------|
| `evaluation/scoring.py` | 944 | Codex+Claude | bug | Stale symlinks in `vbench_videos/` across re-scoring with fewer repeats |
| `evaluation/scoring.py` | 922 | Claude | api-contract | VBench-specific ctor args not plumbed through `AccuracyConfig` |
| `evaluation/scoring.py` | 893 | Claude | api-contract | `extractor` required-but-unused for VBench; example YAML lies with `identity_extractor` |
| `evaluation/scoring.py` | 977 | Claude | testing | No tests for subprocess failure, malformed results, missing dim, prompts with `/`/`..` |

### 🔵 Consider (low)

| File | Line | Reviewer(s) | Category | Summary |
|------|-----:|-------------|----------|---------|
| `evaluation/scoring.py` | 859 | Codex+Claude | design | `_DEFAULT_VBENCH_PROJECT_PATH` breaks under wheel install |
| `accuracy/vbench_runner.py` | 60 | Claude | bug | Silent CPU fallback for a GPU-required evaluator |
| `accuracy/vbench_runner.py` | 75 | Claude | error-handling | No try/except wrapping `vb.evaluate` |

Inline thread: https://github.com/mlcommons/endpoints/pull/315#pullrequestreview-4332489236

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -842,3 +845,185 @@ def score(self) -> tuple[float, int]:
...
+
+        # VBench writes `{run_name}_eval_results.json` to vbench_out. Each
+        # dim entry is `[aggregate_score, [per_video_results, ...]]`.
+        results_path = vbench_out / f"{run_name}_eval_results.json"
+        with results_path.open() as f:
+            results = msgspec.json.decode(f.read())
+
+        per_dim_scores: list[float] = [
+            float(results[dim][0]) for dim in self.dimensions
+        ]
```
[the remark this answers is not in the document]
wu6u3tw: Addressed in d6dc209 — `score()` now wraps the dimension lookup in an explicit `try/except KeyError` and re-raises a named `ValueError` that lists the missing dim and points the reader at the VBench prompt suite. See [scoring.py#L1080-L1090](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/evaluation/scoring.py#L1080-L1090).

AGENTS.md
wu6u3tw: Good catch — `.gitignore` updated in d6dc209 to ignore the heavy subproject working artefacts (`pretrained/`, `evaluation_results/`). See [.gitignore](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/.gitignore).

src/inference_endpoint/evaluation/scoring.py
wu6u3tw: Fixed in d6dc209 — added `_sanitize_prompt_for_filename` which rejects `..` segments and replaces `/`/null bytes before the path is built. `_stage_videos` calls it on every prompt before constructing the symlink. See [scoring.py#L879-L894](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/evaluation/scoring.py#L879-L894) and the call site at [L1005](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/evaluation/scoring.py#L1005).

examples/09_Wan22_VideoGen_Example/offline_wan22_accuracy.yaml
wu6u3tw: Fixed in d6dc209 — example YAML now declares the required `wan22_perf` performance dataset alongside the `wan22_vbench` accuracy entry so `setup_benchmark()` validation passes. See [offline_wan22_accuracy.yaml#L37-L51](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/examples/09_Wan22_VideoGen_Example/offline_wan22_accuracy.yaml#L37-L51).

src/inference_endpoint/evaluation/scoring.py
wu6u3tw: Fixed in d6dc209 — adopted the conservative reading: `n_repeats` is now computed from the pre-filter total (`n_total // num_samples`), so a single failure no longer zeros it out; and when every generation failed, the scorer returns `(None, n_repeats)` matching `LiveCodeBenchScorer`'s contract. See [scoring.py#L1115-L1126](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/evaluation/scoring.py#L1115-L1126).

src/inference_endpoint/evaluation/scoring.py
wu6u3tw: Fixed in d6dc209 — `score()` wraps the dim lookup in an explicit `try/except` and re-raises as a `ValueError` naming the missing dim. RUNBOOK §5 cross-references this. See [scoring.py#L1080-L1090](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/evaluation/scoring.py#L1080-L1090).

src/inference_endpoint/evaluation/scoring.py
wu6u3tw: Fixed in d6dc209 — `_run_vbench_subprocess` now passes `stdin=subprocess.DEVNULL`, tees stdout+stderr to `report_dir/vbench_subprocess.log`, takes a configurable `subprocess_timeout_s`, and on failure raises with the stderr tail included. See [scoring.py#L1043-L1075](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/evaluation/scoring.py#L1043-L1075).

src/inference_endpoint/evaluation/scoring.py
wu6u3tw: Fixed in d6dc209 — `_stage_videos` now calls `src.resolve(strict=True)`, so a missing or unmounted source raises `FileNotFoundError` at staging with the offending path in the message instead of letting VBench fail opaquely on a dangling symlink. See [scoring.py#L1009-L1011](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/evaluation/scoring.py#L1009-L1011).

src/inference_endpoint/evaluation/scoring.py
wu6u3tw: Fixed in d6dc209 — `_stage_videos` now `shutil.rmtree(staged_dir)` at the top before recreating it, so stale `{prompt}-{N}.mp4` from a prior run with more repeats cannot survive into the next score. See [scoring.py#L1000-L1003](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/evaluation/scoring.py#L1000-L1003).

src/inference_endpoint/evaluation/scoring.py
wu6u3tw: Fixed in d6dc209 — added `AccuracyConfig.extras: dict[str, Any]` and `execute._load_datasets` now forwards it as `**kwargs` to the scorer ctor. VBench-specific kwargs (`dimensions`, `vbench_project_path`, `subprocess_timeout_s`, …) are now YAML-driven. See [schema.py#AccuracyConfig.extras](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/config/schema.py) and the wiring at [execute.py#L278](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/commands/benchmark/execute.py#L278) / [#L699](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/commands/benchmark/execute.py#L699).

src/inference_endpoint/evaluation/scoring.py
wu6u3tw: Fixed in d6dc209 — added `Scorer.REQUIRES_EXTRACTOR: ClassVar[bool] = True` (default) and set it to `False` on `VBenchScorer`. `execute._load_datasets` now only enforces `extractor is not None` when the scorer requires it, so the example YAML no longer needs the `identity_extractor` placeholder. See [scoring.py#L65,#L927](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/evaluation/scoring.py#L65) and [execute.py#L255](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/commands/benchmark/execute.py#L255).

src/inference_endpoint/evaluation/scoring.py
wu6u3tw: Addressed in d6dc209 (9 new failure-mode tests) + 705f087 (2 binary-fallback tests). Now covers: subprocess non-zero exit, malformed results JSON, missing dim → `ValueError`, all-failed → `(None, n_repeats)`, prompts containing `..` / `/`, missing src, and the new mp4 binary-response path. See [tests/unit/evaluation/test_scoring.py](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/tests/unit/evaluation/test_scoring.py).

src/inference_endpoint/evaluation/scoring.py
wu6u3tw: Fixed in d6dc209 — `VBenchScorer._resolve_vbench_project_path` resolves the subproject via `$VBENCH_PROJECT_PATH` env var → editable `parents[3]` fallback, and the error message names the env var when neither resolves. Wheel users now set the env var; editable installs work out of the box. See [scoring.py#L869,#L974-#L985](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/src/inference_endpoint/evaluation/scoring.py#L869).

examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
wu6u3tw: Fixed in d6dc209 (well before the merge of this comment) — `vbench_runner.py` now fails fast with a structured stderr JSON `{status:error, type:CudaUnavailable, ...}` and exit code 2 when CUDA is unavailable, unless `--allow-cpu` is explicitly passed. See [vbench_runner.py#L70-L86](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py#L70-L86).

examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
wu6u3tw: Fixed in d6dc209 — `vb.evaluate(...)` is wrapped in `try/except Exception` and prints a structured `{status:error, type, message, traceback}` line on stderr, returning exit code 1. The parent scorer surfaces the stderr tail in its raised error. See [vbench_runner.py#L96-L106](https://github.com/wu6u3tw/endpoints/blob/705f08788c30f67b8ed90c5f1bbce02add6f5f89/examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py#L96-L106).

wu6u3tw:
Modified code per review. Please check it @viraatc @arekay-nv.

arekay-nv (approved):
LGTM. Thanks!

PR #321, opened 2026-05-20
[Feat] Allow tokenizer_name override

tianmu-li:
## What does this PR do?

Allow specifying a tokenizer name that's different from model name in the yaml file.
Example: OpenRouter keeps all model names lower case, so Qwen/Qwen3.6-35B-A3B becomes qwen/qwen3.6-35b. SambaNova strips org name, so it becomes Qwen3.6-35B-A3B. tokenizer name override allows using those endpoints.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [x] Tests added/updated
- [x] All tests pass locally

## Checklist

- [x] Pre-commit hooks pass


arekay-nv:

Reviewed by: **Claude** | Depth: **quick** | Codex: unavailable (branch not on remote)

No new issues found. The one concern identified (silent discard of an invalid user-supplied `tokenizer_name` override) is already covered by an existing inline comment at `execute.py:407`.

arekay-nv:
@tianmu-li can you update this so that if a tokenizer override is incorrect, it is a hard failure. Silently using the default tokenizer can be challenging if the assumption is that a different tokenizer was expected to be used.

tianmu-li:
> @tianmu-li can you update this so that if a tokenizer override is incorrect, it is a hard failure. Silently using the default tokenizer can be challenging if the assumption is that a different tokenizer was expected to be used.

Added. Also added regression testing.

arekay-nv (commented):
## Review Council — follow-up

Reviewing resolved status of prior comments. 2 items remain unresolved (posted inline). 1 item (smaller templates missing `tokenizer_name`) cannot be posted inline since those files aren't in this diff — see the existing comment on `online_template_full.yaml` for context.

src/inference_endpoint/config/schema.py
```diff
@@ -204,6 +204,10 @@ class ModelParams(BaseModel):
         StreamingMode,
         cyclopts.Parameter(alias="--streaming", help="Streaming mode: auto/on/off"),
     ] = StreamingMode.AUTO
+    tokenizer_name: str | None = Field(
```
arekay-nv: **No CLI alias for `tokenizer_name`.**

Every other override in `ModelParams` that users commonly set at the command line has a `cyclopts.Parameter(alias=...)` (e.g. `--streaming`, `--max-new-tokens`). Without one here, `tokenizer_name` can only be set via YAML — defeating the main use-case of a quick per-run override.

```python
tokenizer_name: Annotated[
    str | None,
    cyclopts.Parameter(
        alias="--tokenizer",
        help="HF repo ID or local path for the tokenizer. Overrides model name for ISL/OSL/TPOT metrics.",
    ),
] = None
```

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -401,7 +401,17 @@ def setup_benchmark(config: BenchmarkConfig, test_mode: TestMode) -> BenchmarkCo
 
     # Tokenizer check (light API call, no download)
     model_name = config.model_params.name
-    tokenizer_name = model_name if _check_tokenizer_exists(model_name) else None
+    tokenizer_override = config.model_params.tokenizer_name
+    tokenizer_name: str | None
+    if tokenizer_override:
+        if not _check_tokenizer_exists(tokenizer_override):
+            raise SetupError(
```
arekay-nv, line 408: **Misleading error message — says "not found" for all failure modes.**

`_check_tokenizer_exists()` returns `False` both when the repo/path genuinely doesn't exist and when verification fails due to a network timeout, auth error, or missing tokenizer files. A user hitting a transient network issue will be told their tokenizer is "not found", pointing them at the wrong problem.

Consider wording more generically, e.g.:
```python
raise SetupError(
    f"Tokenizer override '{tokenizer_override}' could not be verified. "
    "Check that the HF repo ID or local path is correct, accessible, and contains tokenizer files. "
    "See logs above for details."
)
```

viraatc (approved):

arekay-nv (approved):
Thanks!

PR #327, opened 2026-05-28
Feat(Urgent): Add support for auto populating system description and run metadata 

anandhu-eng:
## What does this PR do?

This PR adds automated system info capture for MLPerf inference submissions. There are two ways to trigger it:

1. **Standalone (inference-endpoint sysinfo from-config --config benchmark_config.yaml):** collects hardware, software info and serving config from one or more remote nodes via SSH and writes a system_desc.json suitable for endpoints, independent of any benchmark run.

2. **Integrated(if system_info block is included in the benchmark_config.yml when running the benchmark):** runs automatically after a benchmark completes. In addition to collecting the same hardware info, it also extracts serving configuration (tensor/pipeline/expert parallelism, batch size, framework version) from the inference server's startup log or via HTTP probe, and patches those values into run_metadata.json.

Both paths SSH into each target node to run the [get-mlperf-multi-node-system-info](https://github.com/mlcommons/mlperf-automations/tree/dev/script/get-mlperf-multi-node-system-info) [mlcflow](https://github.com/mlcommons/mlcflow) script, merge the per-node results into a single JSON, and optionally validate the topology against a declared node_config (Eg; Prefill/Decode node groupings). Failures in the integrated path are non-blocking, results.json and run_metadata.json are always written first.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [x] Tests added/updated
- [x] All tests pass locally

## Checklist

- [x] Pre-commit hooks pass


src/inference_endpoint/sys_info/capture.py
```diff
@@ -0,0 +1,139 @@
...
+from pathlib import Path
+
+import yaml
+
+from inference_endpoint.config.schema import SysInfoCaptureConfig
+from inference_endpoint.exceptions import ExecutionError, SetupError
+
+logger = logging.getLogger(__name__)
+
+_OUT_FILE_NAME = "system_desc.json"
```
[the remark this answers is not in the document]
anandhu-eng: Updated the DESIGN.md and testa_capture.py accordinly

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -978,6 +981,150 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
     except Exception as e:
         logger.error(f"Save failed: {e}")
 
+    # Write run_metadata.json before sys_info capture so mlcflow's postprocess
+    # can read and patch it in-place with serving config values.
+    metadata_path = ctx.report_dir / "run_metadata.json"
```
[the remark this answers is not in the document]
anandhu-eng: Updated DESIGN.md

pyproject.toml
```diff
@@ -73,6 +73,8 @@ dependencies = [
     # Fix pytz-2024 import warning
     "pytz==2026.1.post1",
     "urllib3==2.7.0",
+    # MLCFlow for system info
+    "mlcflow",
```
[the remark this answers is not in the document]
anandhu-eng: I think it would be better to have the `mlc-script` dependency in the core list, as it is one of the main tasks, with the flexibility to opt out. That said, I am happy to move it to an optional dependency as well.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -987,6 +989,150 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
+            "tensor_parallel": None,
+            "pipeline_parallel": None,
+            "data_parallel": None,
+            "batch": None,
+        },
+        "config_summary_notes": None,
+        "concurrency": concurrency,
+        "system_tps": system_tps,
+        "tps_per_user": tps_per_user,
+        "ttft": _pct(ttft, "99"),
```
arekay-nv: 🔴 **[Both] high · data-integrity**

**Percentile fields silently write `null` in `run_metadata.json`.** `_pct` looks up keys `"99"`, `"95"`, `"90"`, `"50"`, but the metrics registry stores percentile keys as `str(p)` of the float values in `_DEFAULT_PERCENTILES` — i.e. `"99.0"`, `"95.0"`, `"90.0"`, `"50.0"`, `"99.9"` (see `registry.py` `str(p)` + `_DEFAULT_PERCENTILES`; `report.py` passes these float-string keys through). So this `"ttft"` field and every `_p50/_p90/_p95/_p99` field below (1110-1129) resolve to `None`; only the `"99.9"` lookups (`*_p999`) and the `min/max/avg` `_stat` rollups populate. Fix: use float-string keys (`"99.0"`, `"95.0"`, `"90.0"`, `"50.0"`) or normalize on read.

pyproject.toml
```diff
@@ -73,6 +73,8 @@ dependencies = [
     # Fix pytz-2024 import warning
     "pytz==2026.1.post1",
     "urllib3==2.7.0",
+    # mlc-scripts for system info
+    "mlc-scripts",
```
arekay-nv: 🔴 **[Claude] high · design**

**`mlc-scripts` is added with no version pin.** AGENTS.md requires every new dependency be pinned to an exact version (`==`); `[build-system] requires` and the rest of `dependencies` follow this. Pin it (e.g. `mlc-scripts==<version>`) and run `uv run pip-audit`. (The core-vs-`[project.optional-dependencies]` question is already in this file's thread — this comment is only about the missing `==` pin.)

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -987,6 +989,150 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
     except Exception as e:
         logger.error(f"Save failed: {e}")
 
+    # Write run_metadata.json before sys_info capture so mlcflow's postprocess
+    # can read and patch it in-place with serving config values.
+    metadata_path = ctx.report_dir / "run_metadata.json"
+    with metadata_path.open("w") as f:
```
arekay-nv: 🟡 **[Claude] medium · error-handling**

**`run_metadata.json` write is unguarded, unlike the adjacent `results.json` write.** The `results.json` block (963-990) is wrapped in `try/except Exception`, but this `metadata_path.open("w")` / `json.dump` is not. A write failure here (disk full, permissions) propagates out of `finalize_benchmark` and aborts the run *after* a successful benchmark — defeating the stated guarantee that benchmark output is complete regardless of capture outcome. Wrap it in the same try/except.

src/inference_endpoint/config/schema.py
```diff
@@ -609,6 +610,189 @@ def _validate_endpoint_scheme(cls, v: list[str]) -> list[str]:
...
+    @field_validator("endpoint_url", mode="after")
+    @classmethod
+    def _validate_endpoint_url(cls, v: str | None) -> str | None:
+        if v is not None and not v.startswith(("http://", "https://")):
+            raise ValueError(
+                f"endpoint_url must include scheme (http:// or https://), got: {v!r}"
+            )
+        return v
+
+    @field_validator("serving_node", mode="after")
```
arekay-nv, line 1101: 🟡 **[Codex] medium · api-contract**

**`serving_node` skips the port-range check that `ssh_ids` enforces.** `_validate_serving_node` matches `_SSH_ID_RE` but never bounds the port, whereas `_validate_ssh_ids` (736-742) rejects ports outside `1..65535`. So `user@host:99999` passes config validation and the failure is deferred to mlcflow/SSH at runtime with a less actionable error. Apply the same `1 <= port <= 65535` check here.

src/inference_endpoint/config/schema.py
```diff
@@ -609,6 +610,189 @@ def _validate_endpoint_scheme(cls, v: list[str]) -> list[str]:
...
+    """Top-level model for a standalone sysinfo YAML config file.
+
+    The file must have a ``system_info`` key. Extra top-level keys are allowed
+    so the same YAML can be shared with benchmark configs.
+
+    ``report_dir`` mirrors the same field in ``BenchmarkConfig`` and takes
+    priority over ``system_info.output_path`` when set.
+    """
+
+    model_config = ConfigDict(extra="ignore")
```
arekay-nv, line 1169: 🟡 **[Codex] medium · api-contract**

**Shared configs won't auto-populate `endpoint_url` for sysinfo.** `BenchmarkConfig._propagate_endpoint_url_to_sysinfo` copies `endpoint_config.endpoints[0]` into `system_info.endpoint_url`, but `SysInfoFileConfig` (used by `sysinfo from-config`) has no `endpoint_config` field and `extra="ignore"` drops it. The docstring says this YAML can be shared with benchmark configs, yet `sysinfo from-config` on a shared config that omits `system_info.endpoint_url` skips the HTTP serving-framework probe — so standalone sysinfo collects less metadata than the benchmark path. Propagate the endpoint here too, or document the asymmetry.

tests/unit/commands/test_benchmark_finalization.py
```diff
@@ -0,0 +1,125 @@
...
+    def test_execution_error_does_not_prevent_results_write(
+        self, tmp_path: Path
+    ) -> None:
+        """ExecutionError from capture_system_info must not prevent results.json."""
+        ctx = _make_ctx(tmp_path)
+        bench = _make_bench(tmp_path)
+
+        with (
+            patch(
+                "inference_endpoint.commands.benchmark.execute._build_run_metadata",
```
arekay-nv, line 149: 🟡 **[Claude] medium · testing**

**These non-blocking tests stub out the real `_build_run_metadata` (`return_value={}`)**, so the most logic-dense new function is never executed by any test — which is exactly why the percentile-key bug above is invisible to the suite. Add a test that runs the real `_build_run_metadata` with a populated `Report` and asserts the percentile fields are non-null and in ms, plus a test where the `run_metadata.json` write itself fails.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -987,6 +989,150 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
+            if total_tokens is not None:
+                measured_total_output_tokens = int(total_tokens)
+        if concurrency is not None and system_tps is not None:
+            tps_per_user = system_tps / concurrency
+        ttft = report.ttft or {}
+        tpot = report.tpot or {}
+        latency = report.latency or {}
+
+    metadata: dict[str, Any] = {
+        "run_date": datetime.now().isoformat(),
```
arekay-nv: 🔵 **[Codex] low · data-integrity**

**`run_date` uses naive local time.** `datetime.now().isoformat()` has no tz offset; for MLPerf submission metadata compared across nodes/timezones this is ambiguous. Prefer `datetime.now(timezone.utc).isoformat()`.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -900,6 +900,8 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
             report.display(fn=lambda s: print(s, file=f))
         logger.info(f"Report written to {report_txt}")
 
+    run_metadata = _build_run_metadata(ctx, report)
```
arekay-nv: 🔵 **[Claude] low · bug**

**`_build_run_metadata` runs before `results.json` is written.** If it raises (unexpected `report` shape, or a future `LoadPattern` lacking `target_concurrency`), finalization aborts before any benchmark artifact is saved — the opposite of the intended ordering. Move the call to just before the `run_metadata.json` write, or guard it.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -987,6 +989,150 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
+        measured_total_requests = report.n_samples_completed
+        if report.duration_ns is not None:
+            measured_run_duration = report.duration_ns / 1e9
+        osl = report.output_sequence_lengths or {}
+        if osl:
+            total_tokens = osl.get("total")
+            if total_tokens is not None:
+                measured_total_output_tokens = int(total_tokens)
+        if concurrency is not None and system_tps is not None:
+            tps_per_user = system_tps / concurrency
```
arekay-nv, line 1454: 🔵 **[Codex] low · bug**

**`system_tps / concurrency` has no zero-guard.** Safe today (schema rejects `target_concurrency <= 0`; offline yields `None`, guarded by `concurrency is not None`), but a `concurrency > 0` check would make the metadata builder robust to future load-pattern changes.

src/inference_endpoint/config/schema.py
```diff
@@ -777,6 +964,20 @@ def _resolve_and_validate(self) -> Self:
...
+    @model_validator(mode="after")
+    def _propagate_endpoint_url_to_sysinfo(self) -> Self:
+        """Copy endpoint_config.endpoints[0] into system_info.endpoint_url if unset."""
+        if (
+            self.system_info is not None
+            and self.system_info.endpoint_url is None
+            and self.endpoint_config.endpoints
+        ):
+            new_sic = self.system_info.model_copy(
+                update={"endpoint_url": self.endpoint_config.endpoints[0]}
```
arekay-nv: 🔵 **[Codex] low · api-contract**

**`model_copy(update=...)` bypasses field validators.** The injected `endpoint_url` skips `_validate_endpoint_url` (scheme check). Well-formed today because endpoints are scheme-validated elsewhere, but the invariant is now implicit — if endpoint scheme validation is relaxed later, the sysinfo `endpoint_url` contract degrades silently. Consider `model_validate` on the updated dict.

arekay-nv (commented):

Reviewed by: **Codex** (gpt-5.4) + **Claude** | Depth: **thorough**

10 inline findings posted across the sysinfo / run-metadata changes. See the summary comment for the tiered breakdown and the design/performance/deployment impact analysis.

arekay-nv:

Reviewed by: **Codex** (gpt-5.4) + **Claude** | Depth: **thorough**

Found **10** issues across **5** files. All line numbers verified against source at HEAD (`8d99b66`); two topics already raised by existing bots (mlcflow core-vs-extra, `run_metadata` JSON/YAML doc mismatch) were de-duplicated.

### 🔴 Must Fix (high)
| 1 | `commands/benchmark/execute.py` | 1100 | data-integrity | Both | `_pct` looks up `"99"/"95"/"90"/"50"` but registry keys are float-strings (`"99.0"`…); `ttft` + all `p50/p90/p95/p99` fields write **`null`**. Only `*_p999` works. |
| 2 | `pyproject.toml` | 77 | design | Claude | `mlc-scripts` added with **no `==` pin** — violates AGENTS.md dependency rule. |

### 🟡 Should Fix (medium)
| 3 | `commands/benchmark/execute.py` | 995 | error-handling | Claude | `run_metadata.json` write is **unguarded** (unlike `results.json`); a write failure aborts an otherwise-complete benchmark. |
| 4 | `config/schema.py` | 707 | api-contract | Codex | `serving_node` skips the `1..65535` port check that `ssh_ids` enforces → bad port deferred to runtime. |
| 5 | `config/schema.py` | 772 | api-contract | Codex | `SysInfoFileConfig` doesn't propagate `endpoint_url`, so shared-config `sysinfo from-config` silently skips the serving-framework probe. |
| 6 | `tests/unit/commands/test_benchmark_finalization.py` | 87 | testing | Claude | Tests stub `_build_run_metadata` → the real builder (and bug #1) is never exercised. |

### 🔵 Consider (low)
| 7 | `commands/benchmark/execute.py` | 1086 | data-integrity | Codex | `run_date` is naive local time — ambiguous for cross-node submission metadata; use UTC. |
| 8 | `commands/benchmark/execute.py` | 903 | bug | Claude | `_build_run_metadata` runs **before** `results.json` is written; a raise aborts before artifacts are saved. |
| 9 | `commands/benchmark/execute.py` | 1080 | bug | Codex | `system_tps / concurrency` has no zero-guard (safe today via validation; defensive only). |
| 10 | `config/schema.py` | 976 | api-contract | Codex | `model_copy(update=...)` bypasses the `endpoint_url` scheme validator. |

---

## Change summary & impact

**What the PR does.** Adds auto-population of MLPerf system description + run metadata. New pieces: a standalone `sysinfo from-config` command (`commands/sysinfo/`), a `sys_info/capture.py` wrapper around the external `get-mlperf-multi-node-system-info` mlcflow script, a `system_info` config block (`SysInfoCaptureConfig` / `SysInfoFileConfig` in `schema.py`), and a `run_metadata.json` emitter in benchmark finalization recording latency/throughput stats. Also includes a genuine bug fix in `dataset_manager/transforms.py` (`columns_to_keep = list(self.required_columns)` — prevents the later `+=` from mutating the shared `required_columns` list across calls).

**Design.** Additive and largely clean. `system_info` defaults to `None`; the discriminated union and existing validators are unchanged; capture is fully gated behind `system_info is not None`, and a missing mlcflow raises `SetupError` from the lazily-imported `mlc` rather than crashing. The main design debts are the percentile-key contract mismatch (#1) and the duplicated-but-divergent validation between `serving_node` and `ssh_ids` (#4).

**Performance.** No hot-path impact. All new code runs in finalization or in the standalone `sysinfo` command — nothing touches `load_generator/`, `endpoint_client/`, or the ZMQ transport. No new hot-path serialization or async suspends.

**Deployment / backward-compatibility (the explicit focus).** Existing `benchmark` (offline/online/from-config) and `probe` workflows are **preserved**:
- `system_info` is optional and defaults to `None`; regenerated `*_template_full.yaml` only append `system_info: null`.
- No new required CLI args, no schema defaults changed, no discriminated-union change → existing YAMLs and CLI invocations keep working.
- sysinfo capture is gated and failure-tolerant: `ExecutionError` **and** generic exceptions are caught and logged, and `results.json`/`run_metadata.json` are written *before* capture, so a missing/failing mlcflow never aborts a benchmark.

The one real deployment regression is **#2**: `mlc-scripts` lands in the **core** `dependencies` list unpinned, so every install now pulls the full mlcflow stack (even users who never run sysinfo) at an unpinned version. That widens the install surface and breaks reproducibility for existing consumers — worth resolving before merge. Functionally, **#1** means the headline `run_metadata.json` latency percentiles ship as `null` today, and **#6** explains why CI doesn't catch it.

docs/commands/DESIGN.md
```diff
@@ -94,8 +98,134 @@ commands/benchmark/execute.py::run_benchmark()
...
+
+| Field                    | Type                             | Default | Description                                                                                                                                                                                                                  |
+| ------------------------ | -------------------------------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ssh_ids`                | `list[str]`                      | —       | **Required.** Nodes to collect hardware info from. Format: `user@host` or `user@host:port`.                                                                                                                                  |
+| `accelerator_backend`    | `"cuda"` \| `"rocm"` \| `"none"` | —       | **Required.** GPU backend on the target nodes.                                                                                                                                                                               |
```
attafosu: Can you add "xpu" to the backends?

anandhu-eng: Hi @attafosu , thanks for pointing it out. it has been added in recent commit

anandhu-eng:
Hi @arekay-nv @nvzhihanj, would you be able to review and merge this if everything looks good?

arekay-nv (commented):

Reviewed by: **Claude** (Codex unavailable — enterprise policy block) | Depth: **thorough**

Found **9** issues across multiple files. See inline comments for details.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -996,6 +996,152 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
+        with open(metadata_path, "w") as f:
+            json.dump(run_metadata, f, indent=2)
+        logger.info("Run metadata written to %s", metadata_path)
+    except Exception as e:
+        logger.error("Failed to write run_metadata.json: %s", e)
+
+    if ctx.config.system_info is not None:
+        try:
+            # Local import: mlcflow is optional and only needed when system_info is configured.
+            from inference_endpoint.sys_info.capture import (
```
arekay-nv: **[Claude] medium · design** — Lazy import violates project rules and the `SetupError` is dead code.

```python
# Local import: mlcflow is optional and only needed when system_info is configured.
from inference_endpoint.sys_info.capture import (
    capture_system_info,
)
```

AGENTS.md only permits lazy imports for three reasons: circular-import avoidance, optional deps via a top-level `try/except`, or security sandboxing. None apply here: `mlc-scripts==1.1.0` is in the core `dependencies` array in `pyproject.toml`, so the `ImportError` guard in `capture.py` is unreachable. Move the import to the module top level.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -996,6 +996,152 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
+                capture_system_info,
+            )
+
+            output_path = capture_system_info(
+                ctx.config.system_info,
+                output_dir=ctx.report_dir,
+                run_metadata_path=ctx.report_dir / "run_metadata.json",
```
arekay-nv: **[Claude] high · data-integrity** — `run_metadata.json` path is passed to mlcflow unconditionally even if the file was never written.

```python
run_metadata_path=ctx.report_dir / "run_metadata.json",
```

If the `open()` write earlier in the `try` block raises `OSError` (disk full, permissions), the `except Exception` block logs the error and execution continues. The path handed to `capture_system_info` then points to a non-existent file. mlcflow will either error or produce silently incomplete output.

The standalone `sysinfo` CLI correctly guards with `if candidate.exists()`. Apply the same guard here:
```python
metadata_path = ctx.report_dir / "run_metadata.json"
run_metadata_path=metadata_path if metadata_path.exists() else None,
```

src/inference_endpoint/commands/sysinfo/cli.py
```diff
@@ -0,0 +1,73 @@
...
+    The config file must contain a ``system_info`` key with capture settings
+    and an optional ``node_config`` for function-based node groupings.
+
+    Example::
+
+        system_info:
+          ssh_ids:
+            - user@host
+          accelerator_backend: cuda
+          output_path: /tmp/sys_info
```
arekay-nv: **[Claude] medium · api-contract** — Docstring example uses a non-existent field `output_path`.

```yaml
system_info:
  output_path: /tmp/sys_info   # this field does not exist
```

`SysInfoCaptureConfig` has no `output_path` field — output location is controlled by the top-level `report_dir` field of `SysInfoFileConfig`. Because `SysInfoCaptureConfig` uses `extra="forbid"`, a user following this example gets a `ValidationError` at runtime with no obvious hint about the correct field name. Remove `output_path` from the example and add a top-level `report_dir` key instead.

tests/unit/sys_info/test_capture.py
```diff
@@ -0,0 +1,373 @@
...
+        cfg = _make_config(accelerator_backend="rocm", exclude_current_system=True)
+        tags = _build_tags(cfg)
+        assert tags == "get-mlperf-multi-node-system-info,_rocm,_exclude_current_node"
+
+
+def _build_tags(cfg: SysInfoCaptureConfig) -> str:
+    """Mirror the tag-building logic from capture.py."""
+    tags: list[str] = [
+        "get-mlperf-multi-node-system-info",
+        f"_{cfg.accelerator_backend}",
```
arekay-nv: **[Claude] medium · testing** — `_build_tags` test helper diverges from the real implementation for `accelerator_backend="none"`.

Helper (here):
```python
f"_{cfg.accelerator_backend}",   # always appended
```

Real code (`capture.py`):
```python
if config.accelerator_backend \!= "none":
    tags.append(f"_{config.accelerator_backend}")
```

For `accelerator_backend="none"`, the helper produces `get-mlperf-multi-node-system-info,_none` while the real function produces `get-mlperf-multi-node-system-info`. Since no existing test case uses `"none"`, this divergence is invisible. Add a test case for `accelerator_backend="none"` and fix the helper to match the implementation.

pyproject.toml
```diff
@@ -73,6 +73,8 @@ dependencies = [
     # Fix pytz-2024 import warning
     "pytz==2026.1.post1",
     "urllib3==2.7.0",
+    # mlc-scripts for system info
+    "mlc-scripts==1.1.0",
```
arekay-nv: **[Claude] medium · design** — `mlc-scripts` in core deps inflates every install and transitively pulls in `requests`.

```toml
"mlc-scripts==1.1.0",
```

This adds `requests` and `giturlparse` as transitive dependencies for every user of the benchmark tool, even those who never use `system_info`. AGENTS.md explicitly prohibits `requests` from production code. Consider moving `mlc-scripts` to an optional group (e.g. `[project.optional-dependencies] sysinfo = ["mlc-scripts==1.1.0"]`) so users opt in, which would also make the `ImportError` guard in `capture.py` live rather than dead code.

src/inference_endpoint/sys_info/capture.py
```diff
@@ -0,0 +1,147 @@
...
+    assert config.node_config is not None
+    data = {
+        "system_info": {
+            "node_config": {
+                func: [entry.model_dump() for entry in entries]
+                for func, entries in config.node_config.items()
+            }
+        }
+    }
+    fd, path = tempfile.mkstemp(suffix=".yaml", prefix="mlperf_node_cfg_")
```
arekay-nv, line 56: **[Claude] low · bug** — Temp file leak if `yaml.dump` raises after `mkstemp`.

```python
fd, path = tempfile.mkstemp(suffix=".yaml", prefix="mlperf_node_cfg_")
with os.fdopen(fd, "w") as fh:
    yaml.dump(data, fh, default_flow_style=False)
return path
```

If `yaml.dump` raises, the `with` block closes the fd but `path` is never returned, so `node_config_tmp` stays `None` in the caller and the `finally` block's `os.unlink` guard won't fire. Wrap with try/except to unlink on error:
```python
fd, path = tempfile.mkstemp(suffix=".yaml", prefix="mlperf_node_cfg_")
try:
    with os.fdopen(fd, "w") as fh:
        yaml.dump(data, fh, default_flow_style=False)
except Exception:
    os.unlink(path)
    raise
return path
```

src/inference_endpoint/config/schema.py
```diff
@@ -644,6 +645,192 @@ def _validate_endpoint_scheme(cls, v: list[str]) -> list[str]:
...
+    )
+    serving_node: str | None = Field(
+        default=None,
+        description=(
+            "SSH ID of the node running the serving framework "
+            "(e.g. 'root@host:8022'). Used together with log_path for "
+            "log-based framework detection when endpoint_url is unavailable."
+        ),
+    )
+    log_path: str | None = Field(
```
arekay-nv: **[Claude] low · api-contract** — No cross-field validation between `serving_node` and `log_path`.

```python
log_path: str | None = Field(default=None, ...)
```

A user who sets `log_path` but omits `serving_node` silently passes the path to mlcflow without a target host, likely producing confusing or incomplete output. A `model_validator` checking `log_path is None or serving_node is not None` would catch this at config-load time with a clear error message.

src/inference_endpoint/sys_info/capture.py
```diff
@@ -0,0 +1,147 @@
...
+    Returns the Path to the generated combined system info JSON file.
+    Raises SetupError if mlcflow is not installed.
+    Raises ExecutionError if the script returns a non-zero return code.
+    """
+    # Optional dependency — only imported when this function is actually called.
+    # mlc-scripts (PyPI) installs its runtime under the 'mlc' module name.
+    try:
+        import mlc  # noqa: PLC0415
+    except ImportError as exc:
+        raise SetupError(
```
arekay-nv, line 82: **[Claude] low · error-handling** — `ImportError` handler is unreachable dead code.

```python
try:
    import mlc  # noqa: PLC0415
except ImportError as exc:
    raise SetupError("mlc-scripts is required...") from exc  # never reached
```

`mlc-scripts==1.1.0` is in core `dependencies`, so `ImportError` can never be raised in a correctly installed environment. This creates a misleading signal that the feature is optional-install-gated. See also: related comment on `pyproject.toml:77` about moving `mlc-scripts` to an optional group, which would make this guard meaningful.

tests/unit/commands/test_benchmark_finalization.py
```diff
@@ -0,0 +1,286 @@
...
+        """A write error on run_metadata.json must not propagate out of finalize_benchmark."""
+        ctx = _make_ctx(tmp_path)
+        bench = _make_bench(tmp_path)
+
+        with (
+            patch(
+                "inference_endpoint.commands.benchmark.execute._build_run_metadata",
+                return_value={},
+            ),
+            patch("builtins.open", side_effect=OSError("disk full")),
```
arekay-nv, line 91: **[Claude] low · testing** — Global `builtins.open` patch blocks all writes, including `results.json`.

```python
patch("builtins.open", side_effect=OSError("disk full")),
```

The test only asserts no exception propagates, but because `results.json` is also blocked, it doesn't verify the write-ordering invariant (results.json written before run_metadata.json). The sibling test `test_results_json_written_even_if_metadata_write_fails` does this correctly with selective patching. Consider renaming this test to `test_disk_full_does_not_abort_finalize` to clarify its narrower scope.

arekay-nv:

Reviewed by: **Claude** (Codex unavailable — enterprise policy block) | Depth: **thorough**

Found **9 issues** across 7 files.

### 🔴 Must Fix (high)

| 1 | `src/inference_endpoint/commands/benchmark/execute.py` | 1020 | data-integrity | `run_metadata.json` path passed to mlcflow unconditionally even if the file write failed |

### 🟡 Should Fix (medium)

| 2 | `src/inference_endpoint/commands/benchmark/execute.py` | 1013 | design | Lazy import violates AGENTS.md rules; `ImportError` guard is dead code since `mlc-scripts` is in core deps |
| 3 | `src/inference_endpoint/commands/sysinfo/cli.py` | 50 | api-contract | Docstring example uses `output_path` which doesn't exist on `SysInfoCaptureConfig` — causes `ValidationError` at runtime |
| 4 | `tests/unit/sys_info/test_capture.py` | 148 | testing | `_build_tags` helper always appends `accelerator_backend` tag; real code skips it for `"none"`, gap is untested |
| 5 | `pyproject.toml` | 77 | design | `mlc-scripts` in core deps pulls in transitive `requests` (violates AGENTS.md) and inflates every install |

### 🔵 Consider (low)

| 6 | `src/inference_endpoint/sys_info/capture.py` | 56 | bug | Temp file leak if `yaml.dump` raises after `mkstemp` (fd closed but path never unlinked) |
| 7 | `src/inference_endpoint/config/schema.py` | 717 | api-contract | No cross-field validation: `log_path` without `serving_node` silently passes an incomplete config to mlcflow |
| 8 | `src/inference_endpoint/sys_info/capture.py` | 78 | error-handling | `ImportError` handler is unreachable dead code (related to #2 and #5) |
| 9 | `tests/unit/commands/test_benchmark_finalization.py` | 90 | testing | Global `builtins.open` patch also blocks `results.json`; test scope is narrower than its name implies |

pyproject.toml
anandhu-eng: Hi @arekay-nv, would `requests` really be a dealbreaker here? We're not using `requests` directly in any of our modified code. It comes in transitively as part of the `mlc-scripts` library itself. So moving `mlc-scripts` to optional wouldn't actually eliminate `requests` from the environment for users who need `sysinfo`. They'd still end up installing it either way.

Also, we're planning to integrate `sysinfo` directly into endpoints to make it simpler for users so it would effectively be a standard dependency regardless. Just checking if this change is really necessary given that context.

arekay-nv: @anandhu-eng  - can we make `mlc-scripts` an optional dependency so that people who are using it locally or trying it out without immediate intent to submit do not install it. The only difference would be for submission they would need to install the optional dependencies and that should be fine. 
This reduces the barrier to entry.

anandhu-eng: Hi @arekay-nv, have updated accordingly. Thanks!

anandhu-eng:
Hi @arekay-nv @nvzhihanj, just a gentle ping, could we go ahead and merge this PR if everything looks good on your end?

attafosu:
@copilot resolve the merge conflicts in this pull request

attafosu (approved):

viraatc (approved):

attafosu:
@copilot can you change the commit author to my credentials? It seems Copilot-bot is still not being allowed

attafosu:
recheck

arekay-nv:
@anandhu-eng lets discuss the approach in a post-mortem. From my experience running the system info tool standalone attempts to install cuda on the compute node which is a blocker. We need to restructure it so that the system info is not a blocker for the benchmark and lightweight to run.

arekay-nv:
@anandhu-eng - should we close this?

PR #328, opened 2026-05-28
feat: add E2E perf tests for mlperf-endpoints

viraatc:
Drive the cyclopts `inference-endpoint` app in-process against the existing `MaxThroughputServer` and `VariableResponseServer` stubs, plus a throughput matrix for the metrics-tokenizer paths. Three families, all marked `@pytest.mark.performance` (CI-skipped — run on demand):

* **Roofline** — measures peak QPS for `max_throughput`, `concurrency` sweep (1k / 4k / 16k), and binary-searches the largest 10 k-multiple `target_qps` Poisson sustains. Reports numbers rather than asserting on them.

* **Low-QPS correctness** — 5 QPS Poisson against the realistic stub for ~20 s; asserts zero failed requests. Guards keep-alive / idle-pool / slow-response regressions that only surface when connections sit idle longer than `TCP_KEEPIDLE`.

* **Tokenizer throughput matrix** — one parametrized test drives `TokenBatchQueue` end-to-end over every input kind × flush lane, so a single ~1-minute run gives the full token-throughput overview: `text` (the pre-existing batched path: live = in-process thread pool ≤1024 items/flush; drain = fan-out across every pinned shard process) and the chat-template kinds added with structured token counting (#441) — `msg` (one structured assistant *output*: content + reasoning + tool call; the OSL/TPOT path) and `prompt` (one complete structured *input*: message history + tools + generation prompt; the structured-ISL path). Drain cells spawn the full shard pool exactly as production drain does — the msg/prompt drain rows matching the 2-thread live rows despite shards is the #441 architecture (shards only accelerate `TextInput`), not a config mistake. Structured kinds render one `apply_chat_template` per item against a new `char_chat` test tokenizer (same Rust backend as `char` plus a minimal ChatML-style template; no network). Reports items/s and chars processed per cell; asserts only correctness (every enqueued item recorded, zero pending).

A shared `tests/performance/conftest.py` captures each parameterized case via a `record_result` fixture and renders a unified summary table with host / CPU / core info at end of session, so cross-machine runs are easy to compare.

Test knobs are pytest options (see the `roofline` group in `pytest --help`): `--roofline-server-workers`, `--roofline-stream-interval`, `--roofline-client-workers`, `--roofline-init-timeout`.

WARNING: full run takes ~9–14 min wall.

## How to run

```bash
uv run pytest -vs -m performance --no-cov \
    tests/performance/commands/test_e2e_perf.py \
    tests/performance/async_utils/services/metrics_aggregator/test_token_metrics_perf.py

# knobs, e.g. slow worker spawn on big-core-count machines:
uv run pytest -vs -m performance --no-cov tests/performance/commands/test_e2e_perf.py \
    --roofline-init-timeout=300 --roofline-stream-interval=20
```

## Example output — local dev box (48-core Threadripper)

```
=========================== E2E Performance Summary ============================
Host:  nova-2    Arch: x86_64    Cores: 48
CPU:   AMD Ryzen Threadripper PRO 7965WX 24-Cores

Test                       Stream  QPS        Total       Elapsed   Failed
-------------------------  ------  ---------  ----------  --------  ------
max_throughput (2M burst)  off        45,962   2,000,000    43.51s     0
concurrency c=1,000        off        61,376     736,486    12.00s     0
concurrency c=4,000        off        64,883     776,000    11.96s     0
concurrency c=16,000       off        61,752     736,000    11.92s     0
poisson max_sustained      off        30,000  —           —            0
max_throughput (2M burst)  on         35,122   2,000,000    56.94s     0
concurrency c=1,000        on         41,984     503,761    12.00s     0
concurrency c=4,000        on         41,386     497,226    12.01s     0
concurrency c=16,000       on         42,461     512,000    12.06s     0
poisson max_sustained      on         20,000  —           —            0
low_qps target=5           off          4.99         200    40.07s     0
low_qps target=5           on           4.89         200    40.87s     0
================== 12 passed, 2 warnings in 503.68s (0:08:23) ==================
```

Tokenizer throughput matrix on the same box (~1 min standalone run; `Chars` = total text payload fed to the tokenizer):

```
Test                         Stream  QPS        Total       Chars      Elapsed   Failed
---------------------------  ------  ---------  ----------  ---------  --------  ------
tok text live (2 thr)        —           8,792      32,768      16.7M     3.73s     0
tok text drain (6 shards)    —          78,990     524,288     267.2M     6.64s     0
tok msg live (2 thr)         —           2,029      16,384      13.1M     8.07s     0
tok msg drain (6 shards)     —           2,061      16,384      13.1M     7.95s     0
tok prompt live (2 thr)      —           1,083      16,384      24.4M    15.13s     0
tok prompt drain (6 shards)  —           1,101      16,384      24.4M    14.87s     0
```

Reading the matrix: text drain ≫ text live (shard fan-out vs capped thread pool); the #441 chat-template kinds are identical live vs drain (per-item, thread-pool-only, GIL-bound — 8 threads ≈ 2 threads) and ~40–70× slower than sharded text; that per-item cost is what accumulates as the structured backlog #441's own stress test drained for 15.3 s. Same-node interleaved A/B vs the pre-#441 API: text path +0.3–0.6 µs/item dispatch (visible only at high shard counts, −8.5% at 28 shards; worst case ~+0.2 s per 524k-item drain), msg path ~8–11% faster (single-shot `tokenize=True`), prompt kind is new work — per structured sample the total chat-template cost is ~2.6–4.5× the old msg-only cost.

## What does this PR do?

Adds the smallest set of E2E parameterized perf tests that exercise all three online load patterns + low-QPS keep-alive guard, plus a metrics-tokenizer throughput matrix (covering the #441 chat-template pathways), behind the existing `performance` marker so CI is unaffected.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Related issues

Tokenizer matrix is the perf-tracking follow-up to #441.

## Testing

- [x] Tests added/updated
- [x] All 18 tests pass on the local dev box (12 E2E ~8 min + 6 tokenizer cells ~1 min); also verified on a computelab-sc B200 node (224-core Xeon 8570) and a lyris GB200 Grace node (144-core aarch64; needs `--roofline-init-timeout=300`)

## Checklist

- [x] Pre-commit hooks pass


tests/performance/commands/test_e2e_perf.py
```diff
@@ -0,0 +1,278 @@
...
+
+@pytest.fixture(scope="module", params=[False, True], ids=["nonstream", "stream"])
+def max_tput_server(request):
+    """Stub server returning fixed pre-compiled responses (roofline target)."""
+    with MaxThroughputServer(
+        port=0,
+        num_workers=4,
+        stream=request.param,
+        stream_interval=10,
+        quiet=True,
```
arekay-nv, line 63: Should we make these configurable? Also, is having the `stream_interval=10` a good value for a max throughput server - we should probably document this.

arekay-nv (approved):
Nice, lets get this in so we can iterate with perf tracking!

tests/performance/commands/test_e2e_perf.py
viraatc: Done — both knobs (plus the two client-side overrides) are now pytest options: `--roofline-server-workers`, `--roofline-stream-interval`, `--roofline-client-workers`, `--roofline-init-timeout` (grouped under "roofline" in `pytest --help`). Also documented `stream_interval=10` in the fixture: the stub response is 160 chars, so 10 chars/event = 16 SSE events per response — at 1 the roofline mostly measures the client's per-chunk SSE parse cost, at 160 streaming accumulation is barely exercised; 16 events approximates a realistic completion while keeping the stub cheap.

tests/performance/async_utils/services/metrics_aggregator/test_token_metrics_perf.py
```diff
@@ -0,0 +1,291 @@
...
+    TextInput,
+    TokenizationInput,
+)
+
+_TOKENIZER_ROOT = Path(__file__).resolve().parents[4] / "assets" / "tokenizers"
+TOKENIZER = str(_TOKENIZER_ROOT / "char")
+CHAT_TOKENIZER = str(_TOKENIZER_ROOT / "char_chat")
+
+# Matches the CLI default for --tokenizer-workers (the live thread lane).
+LIVE_WORKERS = 2
```
nvzhihanj, line 79: I remember worker needs to be tuned up for actual tests during 0.7 - should we test more than workers=2?

tests/performance/async_utils/services/metrics_aggregator/test_token_metrics_perf.py
```diff
@@ -0,0 +1,291 @@
...
+    def __call__(self, count: int) -> None:
+        self.items += 1
+        self.tokens += count
+
+
+def _report(
+    record_result, label: str, n_items: int, elapsed: float, rec: _Recorder
+) -> None:
+    items_per_s = n_items / elapsed
+    record_result(label, qps=items_per_s, total=n_items, elapsed=elapsed, failed=0)
```
nvzhihanj: **medium — no perf-regression gate.** `_report` records `items/s` into the summary table, but nothing ever asserts a floor or diffs a baseline — the only numeric asserts in this file are `rec.tokens > 0`. So these tests can't fail on a tokenization slowdown; they're a benchmark/measurement harness, not a regression guard. Suggest asserting a conservative per-lane `items/s` floor (or baseline-diff with a tolerance) so CI actually catches a regression.

tests/performance/async_utils/services/metrics_aggregator/test_token_metrics_perf.py
```diff
@@ -0,0 +1,291 @@
...
+    items: list[TokenizationInput] = [TextInput(t) for t in _make_texts(N_DRAIN)]
+    loop = asyncio.get_running_loop()
+    rec = _Recorder()
+    with BatchTokenizer(TOKENIZER, live_workers=LIVE_WORKERS) as tok:
+        n_shards = len(tok._procs)
+        queue = TokenBatchQueue(tok, loop)
+        for item in items:
+            queue.enqueue(item, rec)
+        t0 = time.perf_counter()
+        pending = await queue.flush_remaining(None)
```
nvzhihanj: **low — drain throughput isn't tied to the `--drain-timeout` SLA.** The batched drain exists so end-of-run tokenization finishes within the drain budget (otherwise the run is reported incomplete: `state == complete && n_pending_tasks > 0`). This measures drain `items/s`, but never asserts the drain completes within a representative timeout for a realistic backlog — the property that actually matters in production. Consider asserting `flush_remaining(timeout)` finishes (returns 0 pending) within a budget for N_DRAIN items.

nvzhihanj (commented):
## Review — tokenization-perf coverage

Two gaps in the tokenization-perf tests: they measure throughput but don't guard it (A), and drain throughput isn't tied to the `--drain-timeout` SLA (D).

tests/performance/async_utils/services/metrics_aggregator/test_token_metrics_perf.py
nvzhihanj: I think we probably want to set a reasonable floor for when the test should fail

tests/performance/async_utils/services/metrics_aggregator/test_token_metrics_perf.py
nvzhihanj: This is similar to the drain timeout discussion. Can do in a separate PR but let's fail on "drain timeout"

nvzhihanj (approved):

tests/performance/async_utils/services/metrics_aggregator/test_token_metrics_perf.py
nv-alicheng: Right but this was before batched-tokenizer right? In 0.7, we needed to use 8/16 workers before batching was implemented.

viraatc: keeping test-suite to 2, so as not to slow-down CI/CD.
ive increased default to 4 for safety net.

> In 0.7, we needed to use 8/16 workers before batching was implemented.

yes correct.

tests/performance/async_utils/services/metrics_aggregator/test_token_metrics_perf.py
viraatc: Done — added per-lane regression floors at 50% of the slowest hardware each lane was measured on (thread-pool lanes: lyris GB200 Grace; text drain: a 6-shard dev box, since drain scales with shard count): text live 2,800 / text drain 39,000 / msg 700 / prompt 380 items/s. A cell below its floor now fails the test.

tests/performance/async_utils/services/metrics_aggregator/test_token_metrics_perf.py
viraatc: Done in this PR — drain cells now run `flush_remaining` with a budget derived from the floor (2 x n / floor, i.e. fails below 25% of the slowest measured hardware) and assert zero pending, so a drain timeout fails the test the same way production surfaces it (n_pending_tasks > 0).

tests/performance/async_utils/services/metrics_aggregator/test_token_metrics_perf.py
viraatc: Right — the batched tokenizer changed this: the end-of-run drain no longer uses this knob at all (it fans out across auto-sized shard processes, one per 8-core block), and for the chat-template paths added in #441 the thread count barely matters (GIL-bound render: measured 8 threads within 2-18% of 2 threads). We bumped the production default --tokenizer-workers to 4 as a safety net for the live lane, but pinned the perf suite to 2 so runs stay light and numbers stay comparable across machines.

PR #331, opened 2026-06-03
feat(multiturn): add agentic benchmark controls

hvagadia:
## Summary

  Adds the Kimi agentic multi-turn benchmark path, including trajectory budgeting, deterministic salting, tail handling, and inline accuracy scoring over performance replay outputs.

  ## Changes

  - Add multi-turn benchmark controls:
    - `multi_turn.num_trajectories_to_issue`
    - `multi_turn.stop_issuing_on_first_user_complete`
    - longer default per-turn timeout for agentic runs
  - Remove the live-history multi-turn path and simplify dataset replay to use prebuilt dataset messages.
  - Move trajectory salting into the multi-turn issuing strategy, with validation that salted datasets include a system prompt.
  - Support repeated trajectories with logical `__repeat_N` conversation ids so runs can issue more trajectories than the dataset contains.
  - Split the performance measurement window from the multi-turn tail by emitting `STOP_PERFORMANCE_TRACKING` when the first active user runs out of assigned trajectories.
  - Fold multi-turn inline accuracy into `evaluation/scoring.py` via `multi_turn_inline`, scoring already-issued performance events without a separate accuracy phase.
  - Add coding/workflow turn scoring, per-domain score summaries, missing-output accounting, and exclusion of turns without ground truth from the denominator.
  - Update Kimi multi-turn example config and README with fields, benchmark invariants, salting, inline accuracy, and tail-management behavior.
  - Enable `trust_remote_code=True` for tokenizer loading in token metrics.
  - Add/update unit and integration coverage for schema validation, dataset construction, salting, trajectory repeats, stop tracking, and multi-turn inline scoring.


arekay-nv (commented):

Reviewed by: **Codex** (gpt-5.4) + **Claude** · Depth: **thorough**

Posting **10** inline findings. See the pinned summary comment for the tiered breakdown.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -882,6 +892,45 @@ def _salvage_tmpfs(report_dir: Path, tmpfs_dir: Path) -> None:
...
+    ctx: BenchmarkContext,
+) -> tuple[str, dict[str, Any]] | None:
+    if not isinstance(ctx.dataloader, MultiTurnDataset):
+        return None
+
+    perf_cfg = next(
+        (ds for ds in ctx.config.datasets if ds.type == DatasetType.PERFORMANCE),
+        None,
+    )
+    if (
```
arekay-nv: [Codex + Claude] **🔴 critical · bug**: `_inline_score_multi_turn_if_enabled()` is called **unconditionally** for every `MultiTurnDataset` run from `finalize_benchmark` (line 954), but this guard `raise`s `InputValidationError` whenever inline accuracy is *not* enabled — and `multi_turn.inline_accuracy` defaults to `False`.

The function name (`..._if_enabled`) and its `return None` for non-multi-turn datasets show the intent is to **no-op when disabled**. As written, every existing multi-turn benchmark that doesn't opt into inline accuracy now **crashes at finalization, after the run has already completed** — discarding results.

```python
    if (
        perf_cfg is None
        or perf_cfg.multi_turn is None
        or not perf_cfg.multi_turn.inline_accuracy
        or perf_cfg.path is None
    ):
        raise InputValidationError(...)
```

**Fix:** `return None` (skip scoring) when inline accuracy is disabled / path missing. Only raise if inline accuracy was *explicitly requested* but a prerequisite is missing (e.g. `inline_accuracy: true` but no `path`).

src/inference_endpoint/evaluation/multi_turn_inline_accuracy.py
```diff
@@ -0,0 +1,533 @@
...
+        row["gt_actions"] = extract_exes_from_turn(gt)
+        row["model_actions"] = extract_exes_from_turn(model)
+    else:
+        gt_intents = _ground_truth_intent_set(gt) or set()
+        row["gt_intents"] = sorted(gt_intents)
+        row["model_intent"] = _extract_intent_code(model)
+    return row
+
+
+def score_report(
```
arekay-nv: [Claude] **🟡 high · testing**: This entire new ~533-line scorer module has **zero test coverage** — `grep` across `tests/` finds no reference to `score_report`, `inline_accuracy`, or `multi_turn_inline`, and there is no `tests/unit/evaluation/test_multi_turn_inline_accuracy.py`. AGENTS.md targets **>90% coverage for new code**, and you specifically asked to ensure test coverage.

The untested logic is exactly the kind that's error-prone: shell-command canonicalization, tool-call merging (streamed vs complete), multiset IoU, repeat-set validity, and the `valid`/`invalid_reason` computation. Notably, the existing CodeRabbit comments on this same file already flag two robustness crashes (`row["turn"]` KeyError; unguarded `msgspec.DecodeError` on partial `events.jsonl`) — both in code paths that tests would have caught.

**Fix:** add a `tests/unit/evaluation/test_multi_turn_inline_accuracy.py` covering exe extraction, intent extraction, repeat-id splitting, multiset scoring, and the validity/`invalid_reason` branches (with `@pytest.mark.unit`).

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -121,7 +121,7 @@ def _get_thread_tokenizer(self) -> PreTrainedTokenizerBase:
         """Return the tokenizer for the current thread, loading it if needed."""
         if getattr(self._thread_local, "tokenizer", None) is None:
             self._thread_local.tokenizer = AutoTokenizer.from_pretrained(
-                self._tokenizer_name
+                self._tokenizer_name, trust_remote_code=True
```
arekay-nv, line 124: [Codex] **🟡 medium · security**: This PR flips tokenizer loading to `trust_remote_code=True` by default at **two** sites — here (the hot-path tokenizer pool) and `commands/benchmark/execute.py:329` (`_precompute_isl_for_multi_turn`):

```python
            self._thread_local.tokenizer = AutoTokenizer.from_pretrained(
                self._tokenizer_name, trust_remote_code=True
            )
```

`trust_remote_code=True` executes arbitrary Python from the model repo on load. Previously these paths only loaded standard tokenizer assets, so this broadens the security posture for *every* benchmark run — a mistyped or untrusted HF repo id now runs remote code. It's intentional and even has a test asserting the kwarg, but it's enabled globally rather than opt-in and is undocumented.

**Suggestion:** gate remote-code execution behind an explicit config flag (default `False`), or at minimum document the new default and its implication in the README so operators pointing at untrusted repos are aware.

src/inference_endpoint/config/schema.py
```diff
@@ -255,13 +255,26 @@ class MultiTurnConfig(BaseModel):
 
     model_config = ConfigDict(extra="forbid", frozen=True)
 
-    turn_timeout_s: float = Field(default=300.0, gt=0)
-    use_dataset_history: bool = True
+    turn_timeout_s: float = Field(
+        default=86400.0,
```
arekay-nv, line 266: [Claude] **🟡 medium · documentation**: `turn_timeout_s`'s default was changed from `300.0` to `86400.0` (24 hours) and the field was **removed from both the example YAML and the README** in this PR:

```python
    turn_timeout_s: float = Field(
        default=86400.0,
        gt=0,
```

A 24-hour default effectively disables the per-turn safety deadline — if a server hangs mid-turn, the run now blocks for a day instead of 5 minutes. That's a meaningful behavioral change to hide from the example config. `use_dataset_history` (default `True`) was likewise dropped from the docs.

**Fix:** document the new `turn_timeout_s` default and its implication in `examples/09_MultiTurn/README.md`, and either surface `use_dataset_history` or note why it's intentionally omitted.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -882,6 +892,45 @@ def _salvage_tmpfs(report_dir: Path, tmpfs_dir: Path) -> None:
...
+        or perf_cfg.multi_turn is None
+        or not perf_cfg.multi_turn.inline_accuracy
+        or perf_cfg.path is None
+    ):
+        raise InputValidationError(
+            "Multi-turn inline accuracy requires a performance dataset with "
+            "multi_turn.inline_accuracy: true and a dataset path"
+        )
+
+    from inference_endpoint.evaluation.multi_turn_inline_accuracy import score_report
```
arekay-nv: [Claude] **🟡 medium · design**: Lazy import inside a function violates the project's **"No lazy imports"** rule (AGENTS.md):

```python
    from inference_endpoint.evaluation.multi_turn_inline_accuracy import score_report
```

`multi_turn_inline_accuracy.py` only imports from `core.record` / `core.types`, so there is no circular-import justification (the only sanctioned exceptions are circular-import avoidance with a comment, optional-dependency `try/except`, and sandboxing).

**Fix:** move this import to the top of the file alongside the other `inference_endpoint.evaluation` imports.

src/inference_endpoint/load_generator/multi_turn_strategy.py
```diff
@@ -215,30 +224,115 @@ async def execute(self, phase_issuer: PhaseIssuerProtocol) -> int:
...
+        logical_id: str,
+        source_id: str,
+        turns: ConversationTurns,
+        repeat_id: int,
+    ) -> None:
+        sys_content = None
+        if self._store_in_history:
+            sys_content = self._dataset_metadata.system_prompts_by_conv.get(source_id)
+            if sys_content is not None and self._enable_salt:
+                if sys_content.startswith("[salt: "):
```
arekay-nv: [Claude] **🔵 low · bug**: The store-in-history salt strip here only handles the `[salt: ` prefix:

```python
                if sys_content.startswith("[salt: "):
                    marker_end = sys_content.find("]\n\n")
```

but the sibling `_messages_with_logical_id_salt` (line ~472) strips **both** `[salt: ` and the legacy `[cache_salt: ` prefixes. If a `system_prompts_by_conv` value carries a `[cache_salt: ...]` marker, this path won't strip it and will then prepend a fresh `[salt: ...]`, producing doubled/garbled markers in the rendered system prompt.

**Fix:** make both strip routines handle the same set of prefixes (factor out a shared helper).

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -882,6 +892,45 @@ def _salvage_tmpfs(report_dir: Path, tmpfs_dir: Path) -> None:
...
+        out_path=ctx.report_dir / "scores.json",
+    )
+    result_key = "multi_turn_inline"
+    valid = bool(score_result["valid"])
+    score_entry: dict[str, Any] = {
+        "score": score_result["score"],
+        "valid": valid,
+    }
+    if not valid:
+        score_entry["invalid_reason"] = score_result["invalid_reason"]
```
arekay-nv: [Claude] **🔵 low · error-handling**: `score_entry["invalid_reason"] = score_result["invalid_reason"]` (and the mirror at line ~1039) rely on an unguarded cross-module dict contract: that `score_report` always populates `invalid_reason` when `valid` is `False`. That invariant currently holds, so this isn't a live bug — but a future change to `score_report` that returns `valid=False` without a reason would raise `KeyError` at finalization and discard an otherwise-successful run.

**Suggestion:** use `score_result.get("invalid_reason", "")` to make the contract failure-tolerant at this boundary.

src/inference_endpoint/evaluation/multi_turn_inline_accuracy.py
```diff
@@ -0,0 +1,533 @@
...
+            reasons.append("no matching completed turns found")
+        elif not complete_repeat_sets:
+            reasons.append("observed turns do not cover whole dataset repeat(s)")
+        if missing_outputs:
+            reasons.append(
+                f"{missing_outputs}/{expected_total} expected turn(s) missing output"
+            )
+        result["invalid_reason"] = "; ".join(reasons)
+
+    if out_path is not None:
```
arekay-nv: [Claude] **🔵 low · data-integrity**: `scores.json` serializes the full `per_turn` list (one row per expected-turn × repeat) via a single in-memory `json.dumps(result, indent=2)`. For the documented target (990 trajectories × many turns) this artifact can grow very large. The standalone `main()` already strips `per_turn` before printing.

**Suggestion:** gate the verbose `per_turn` dump (summary-only by default, full dump behind a flag) to bound the artifact size.

examples/09_MultiTurn/kimi_agentic_benchmark.yaml
```diff
@@ -1,44 +1,42 @@
...
 
 datasets:
-  # Select the dataset to run by updating both `name` and `path`.
-  # Use agentic_coding for coding traces or agentic_workflow for workflow traces.
-  - name: agentic_coding
-    type: performance
-    path: /path/to/agentic_dataset.jsonl
+  - name: agentic_combined
+    type: performance # do not change.
+    path: <path-to-agentic-combined-jsonl>
```
arekay-nv: [Claude] **🔵 low · documentation**: `path: <path-to-agentic-combined-jsonl>` is a literal placeholder. Combined with `inline_accuracy: true # do not change` and `num_trajectories_to_issue: 990`, a user who copies this file and only edits the endpoint will hit the inline-accuracy finalization path with an unreadable ground-truth JSONL and fail (after the run).

**Suggestion:** use a clearer placeholder convention (e.g. `/path/to/agentic_dataset.jsonl`) plus a one-line note that this must be a real local/mounted file because inline accuracy reads it at finalization.

tests/unit/config/test_schema.py
```diff
@@ -492,6 +492,16 @@ def test_multi_turn_valid_config(self):
         assert config.settings.load_pattern.type == LoadPatternType.MULTI_TURN
         assert config.settings.load_pattern.target_concurrency == 16
 
+    @pytest.mark.unit
+    def test_multi_turn_rejects_removed_stop_on_first_empty_slot_config(self):
+        with pytest.raises(ValueError, match="stop_on_first_empty_slot"):
```
arekay-nv, line 507: [Claude] **🔵 low · testing**: `test_multi_turn_rejects_removed_stop_on_first_empty_slot_config` only passes because `MultiTurnConfig` uses `extra="forbid"` — it would pass for *any* unknown key, so it doesn't actually guard the removal of this specific field. It gives false confidence.

```python
        with pytest.raises(ValueError, match="stop_on_first_empty_slot"):
```

**Suggestion:** either drop it (generic `extra="forbid"` behavior is already covered elsewhere) or, if guarding a real migration, document that intent in the test.

arekay-nv:

Reviewed by: **Codex** (gpt-5.4) + **Claude** · Depth: **thorough**

Found **10** issues across **7** files. 10 inline comments posted on the relevant lines.

Will cause incorrect behavior or data loss in production.

| 1 | `commands/benchmark/execute.py` | 905 | bug | **Both** | `_inline_score_multi_turn_if_enabled()` is called unconditionally but **raises** when `inline_accuracy` is disabled (the default) → **every multi-turn run that doesn't opt in crashes at finalization**, discarding results. Should `return None` when disabled. |
| 2 | `evaluation/multi_turn_inline_accuracy.py` | 387 | testing | Claude | The new ~533-line scorer module has **zero tests**. AGENTS.md targets >90% on new code. The two existing CodeRabbit comments flag crashes (`row["turn"]` KeyError, `msgspec.DecodeError`) that tests would have caught. |

### 🟡 Should Fix (medium)
Real issues / behavioral & posture changes that should be addressed.

| 3 | `async_utils/.../token_metrics.py` | 124 | security | Codex | `trust_remote_code=True` now the default at two sites (here + `execute.py:329`) — executes arbitrary repo code on every run. Should be opt-in and/or documented. |
| 4 | `config/schema.py` | 259 | documentation | Claude | `turn_timeout_s` default changed `300 → 86400` (24h, effectively disables the per-turn deadline) and was removed from the README + example YAML. `use_dataset_history` also undocumented. |
| 5 | `commands/benchmark/execute.py` | 916 | design | Claude | Lazy `import score_report` mid-function violates the project's "No lazy imports" rule; no circular-import justification. Move to top of file. |

### 🔵 Consider (low)
Valid improvements / follow-ups.

| 6 | `load_generator/multi_turn_strategy.py` | 295 | bug | Claude | Salt strip handles only `[salt: ` here, but the sibling routine strips both `[salt: ` and `[cache_salt: ` → potential doubled/garbled markers. |
| 7 | `commands/benchmark/execute.py` | 930 | error-handling | Claude | Unguarded `score_result["invalid_reason"]` (also ~1039) — safe today, but a fragile cross-module contract; use `.get(...)`. |
| 8 | `evaluation/multi_turn_inline_accuracy.py` | 493 | data-integrity | Claude | `scores.json` dumps the full `per_turn` list in one `json.dumps`; unbounded for 990× trajectories. Consider summary-only by default. |
| 9 | `examples/09_MultiTurn/kimi_agentic_benchmark.yaml` | 17 | documentation | Claude | `path: <path-to-agentic-combined-jsonl>` placeholder + `inline_accuracy: true` → copy-paste users fail at finalization. Clarify placeholder + add a note. |
| 10 | `tests/unit/config/test_schema.py` | 497 | testing | Claude | `test_..._rejects_removed_stop_on_first_empty_slot` passes only due to `extra="forbid"` — would pass for any unknown key; false confidence. |

---

**On the three things you asked us to weight:**
- **Test coverage** — the four touched modules with new tests are well covered (schema 47, dataset 49, strategy 30, token_metrics 12 tests), but the **scorer module has none** (#2) and one schema test is hollow (#10).
- **Documentation** — the README/YAML drifted from the schema: `turn_timeout_s` default change is hidden (#4), `use_dataset_history` is undocumented (#4), and the example path is a non-runnable placeholder (#9).
- **Conformance** — one clear convention violation (lazy import #5); the rest matches existing patterns.

The **critical finalization crash (#1)** is the blocker — it breaks existing multi-turn runs regardless of this PR's new features.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) · Review Council</sub>

examples/09_MultiTurn/README.md
```diff
@@ -51,36 +51,107 @@ clients; set `model_params.name` in the YAML to the same value.
...
+
+- `name`: human-readable run name written to reports and logs. Change this when
+  creating a distinct benchmark config.
+- `version: "1.0"`: config version label for this example. Final submission:
+  keep `"1.0"`.
+- `type: "online"`: runs through the online scheduler. Final submission: keep
+  `"online"`.
+- `model_params.name`: model name sent in each OpenAI request. Set this to the
+  model name served by the endpoint.
+- `model_params.temperature: 1.0`: sampling temperature sent to the server.
```
nvzhihanj: How about we put the invariant ones in a separate sub-section?

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
nvzhihanj: This should be fine

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -964,6 +950,7 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
 
     # Write scoring artifacts + copy event log from tmpfs to disk
     _write_scoring_artifacts(ctx, result, bench.tmpfs_dir)
+    inline_accuracy_score = _inline_score_multi_turn_if_enabled(ctx)
```
nvzhihanj: Is this separate-path inline scorer necessary? (i.e. can we fold it into the regular evaluation path)

Adding one-off logics for just the agentic seems to be breaking the modularization here

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -902,6 +950,7 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
```
nvzhihanj: [Claude] **high** · design — answering @nvzhihanj's question here: **yes, this should fold into the regular `Scorer`/`accuracy_config` path.** This bespoke pre-step runs before the modular `for eval_cfg in ctx.eval_configs:` loop (line 957), selected by a *different* config surface (`multi_turn.inline_accuracy`) and producing a *different* result shape — the modularization break you flagged.

The `Scorer` ABC already supports exactly this shape, and 4 existing scorers use it: register via `scorer_id`, override `score()` wholesale (RougeScorer / LiveCodeBenchScorer / ShopifyCategoryF1Scorer / VBenchScorer all do, stubbing `score_single_sample`), set `REQUIRES_EXTRACTOR=False`, and receive scorer-specific knobs via `accuracy_config.extras` (already splatted at `**eval_cfg.extras`, line 964). Nothing about agentic scoring needs a new framework. Note "inline" is a misnomer: this runs **post-hoc at finalize**, reading the already-flushed `events.jsonl` — identical timing/inputs to a `Scorer`, so no run-time coupling forces a separate path.

**WAR-now:** wrap the existing algorithm as `class MultiTurnInlineScorer(Scorer, scorer_id="multi_turn_inline", REQUIRES_EXTRACTOR=False)` whose `score()` delegates to `score_report(...)` and returns `(score, n_repeats)`; configure it via `accuracy_config: {eval_method: multi_turn_inline}` on the dataset; **delete** `_inline_score_multi_turn_if_enabled`, the `inline_accuracy` flag, and the special-case merge below. The one general capability to add is "score-in-place" (see the separate note) — not an agentic one-off.

src/inference_endpoint/evaluation/multi_turn_inline_accuracy.py
```diff
@@ -0,0 +1,533 @@
...
+    content, reasoning, tool_calls = data.as_message_parts()
+    return {
+        "role": "assistant",
+        "content": _text_part(content),
+        "reasoning_content": _text_part(reasoning),
+        "tool_calls": list(tool_calls) if tool_calls else None,
+    }
+
+
+def _iter_complete_records(events_path: Path):
```
nvzhihanj: [Claude] **medium** · design — concrete duplication backing the fold: `_iter_complete_records` re-implements `Scorer.get_outputs()` (scoring.py ~134-158) — same `msgspec.json.Decoder(type=EventRecord, dec_hook=EventType.decode_hook)`, same `record.event_type == SampleEventType.COMPLETE` filter, same `report_dir/events.jsonl` source. Two readers of the same log drift independently (they already differ on decode-error handling: this one catches `msgspec.DecodeError` per line and skips; `get_outputs()` lets it raise). Folding into a `Scorer` lets the scorer reuse / subclass one event-decoding implementation (e.g. extend `get_outputs()` to surface `conversation_id`/`turn`).

src/inference_endpoint/evaluation/multi_turn_inline_accuracy.py
```diff
@@ -0,0 +1,533 @@
...
+    gt_jsonl: Path,
+    report_dir: Path,
+    out_path: Path | None = None,
+) -> dict[str, Any]:
+    """Score completed multi-turn replay events and optionally write scores.json."""
+    events_path = report_dir / "events.jsonl"
+    if not events_path.exists():
+        raise FileNotFoundError(f"Events log file not found at {events_path}")
+
+    expected = _build_expected_assistants(gt_jsonl)
```
nvzhihanj: [Claude] **medium** · design — second duplication: `score_report` rebuilds ground truth by re-opening and re-parsing the perf dataset JSONL (`_build_expected_assistants(gt_jsonl)`), even though that data is already loaded as `MultiTurnDataset.dataframe`. The standard path passes the constructed `Dataset` into the scorer (`eval_cfg.scorer(..., eval_cfg.dataset, ...)`) precisely so scorers read `self.dataset.dataframe` instead of redoing I/O. Re-reading the file risks desync from any transform/filter the loaded dataset applied (e.g. the `dataset_metadata` row-drop, `samples=`, parser remaps). A folded scorer reads `self.dataset.dataframe` and can't desync; keep `score_report(gt_jsonl=...)` only as the thin offline-CLI entry.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -928,6 +977,16 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
         }
         logger.info(f"Score for {eval_cfg.dataset_name}: {score} ({n_repeats} repeats)")
 
+    if inline_accuracy_score is not None:
+        result_key, score_entry = inline_accuracy_score
+        accuracy_scores[result_key] = score_entry
```
nvzhihanj: [Claude] **medium** · api-contract — divergent results contract: this writes `accuracy_scores["multi_turn_inline"] = {score, valid[, invalid_reason]}`, while every modular scorer writes `{dataset_name, num_samples, extractor, ground_truth_column, score, n_repeats}` keyed by `dataset_name` (lines 968-977). Two shapes in one `accuracy_scores` dict forces every downstream consumer to branch on key/shape, and it also emits a side `scores.json` no other scorer produces. Folding routes the multi-turn result through the same `(score, n_repeats)` builder; map `valid: false` → `score: None` (the existing "scoring failed → None" convention), and write per-turn detail as the scorer's own side-artifact (as VBench writes `vbench_results/`).

src/inference_endpoint/config/schema.py
```diff
@@ -271,6 +284,31 @@ class MultiTurnConfig(BaseModel):
             "in dataset."
         ),
     )
+    inline_accuracy: bool = Field(
```
nvzhihanj: [Claude] **medium** · design — two config surfaces for one concept (scorer selection). Every other dataset selects its scorer via `accuracy_config.eval_method` on a `type: accuracy` dataset; this introduces `MultiTurnConfig.inline_accuracy: bool` on the `type: performance` dataset, so the kimi yaml has no `accuracy_config`/`eval_method` at all. A reader now has two unrelated places to answer "is this run scored, and by what?", and a boolean can only ever select one hard-coded scorer — the inflexibility `eval_method` exists to avoid. With a `MultiTurnInlineScorer` registered under `eval_method: multi_turn_inline`, this field is redundant and removable (deletes a Pydantic field, a yaml gotcha, and a finalize branch with no loss of capability).

nvzhihanj (commented):
## Review Council — focused design review: fold the agentic eval into the existing path

Reviewed by: **Claude** (Codex ran clean at low reasoning — no correctness regressions — but doesn't assess architecture at that effort). Scope: @nvzhihanj's question at `execute.py:953` — *is the separate inline scorer necessary, or can it fold into the regular evaluation path?*

**Verdict: yes — fold it.** The new path is a parallel re-implementation of the `Scorer` framework: a `multi_turn.inline_accuracy` flag on the *performance* dataset + a separate `_inline_score_multi_turn_if_enabled` finalize call → `score_report`, which re-reads `events.jsonl` (`_iter_complete_records` duplicates `Scorer.get_outputs()`), re-reads the GT JSONL from disk instead of the loaded `Dataset`, and emits a different results shape. The `Scorer` ABC already supports this shape (register via `scorer_id`, override `score()` like 4 existing batch scorers, `REQUIRES_EXTRACTOR=False`, knobs via `accuracy_config.extras`).

**Concrete fold:**
- *WAR-now:* add `class MultiTurnInlineScorer(Scorer, scorer_id="multi_turn_inline", REQUIRES_EXTRACTOR=False)` whose `score()` delegates to the existing `score_report`; configure via `accuracy_config.eval_method`; **remove** `_inline_score_multi_turn_if_enabled`, the `inline_accuracy` flag (schema.py:287), and the special-case merge (execute.py ~980-982).
- *The one general capability to add* (not agentic): **score the performance phase in place**. The existing accuracy path runs a *separate* `PhaseType.ACCURACY` phase that re-issues the dataset and explicitly **blocks** `MultiTurnDataset` (`execute.py:505` — pre-existing, so not annotated inline). That block is the real reason the bypass exists. Add an `accuracy_config.score_in_place` opt-in that skips the re-issue and points the configured scorer at the perf phase's `events.jsonl` + `sample_idx_map.json` (the perf phase already writes one), and lift the line-505 `raise` for in-place scorers. The only other real difference — matching by `(conversation_id, turn, repeat)` from `EventRecord` rather than one-row-per-`sample_index` — lives inside the subclass's `score()`, exactly as DeepSeekR1Scorer/LiveCodeBenchScorer already do custom matching.

**No genuine blocker:** the only piece that legitimately stays separate is `score_report` as a reusable free function (it backs an offline re-scoring CLI, `multi_turn_inline_accuracy.py:499`); the Scorer simply delegates to it. One algorithm, thin entry points, one in-benchmark path.

**Inline comments** mark the concrete evidence: duplicated event reader (`:331`), GT re-read vs loaded dataset (`:397`), divergent result shape (`execute.py:982`), and the second config surface (`schema.py:287`).

_Deduped against the prior review-council run + maintainer comments (critical-bug-now-fixed, zero tests, `trust_remote_code`, `turn_timeout_s`, lazy import, salt strip, scores.json size, yaml placeholder) — not re-reported here; the lazy-import / divergent-shape points appear only as fold evidence._

src/inference_endpoint/load_generator/multi_turn_strategy.py
```diff
@@ -85,10 +87,24 @@ def __init__(
...
         """
         self._conv_manager = conversation_manager
         self._dataset_metadata = dataset_metadata
         self._multi_turn_config = multi_turn_config
+        self._num_trajectories_to_issue = num_trajectories_to_issue
+        self._stop_issuing_on_first_user_complete = (
+            stop_issuing_on_first_user_complete
+            if stop_issuing_on_first_user_complete is not None
+            else (
+                multi_turn_config.stop_issuing_on_first_user_complete
```
viraatc: lets simplify code:
 `multi_turn_config.stop_issuing_on_first_user_complete` always instead of also support `stop_issuing_on_first_user_complete` in __init__

src/inference_endpoint/evaluation/multi_turn_inline_accuracy.py
```diff
@@ -0,0 +1,533 @@
...
+        row["model_intent"] = _extract_intent_code(model)
+    return row
+
+
+def score_report(
+    gt_jsonl: Path,
+    report_dir: Path,
+    out_path: Path | None = None,
+) -> dict[str, Any]:
+    """Score completed multi-turn replay events and optionally write scores.json."""
```
viraatc: can we improve function documentation, seems like a long pipeline

viraatc (approved):

src/inference_endpoint/config/schema.py
```diff
@@ -255,13 +255,26 @@ class MultiTurnConfig(BaseModel):
...
-    use_dataset_history: bool = True
+    turn_timeout_s: float = Field(
+        default=86400.0,
+        gt=0,
+        description=(
+            "Per-turn timeout in seconds. A timeout aborts that turn and all "
+            "remaining turns in the same conversation."
+        ),
+    )
+    use_dataset_history: bool = Field(
```
viraatc: when can submitters need to turn this off?

src/inference_endpoint/config/schema.py
hvagadia: Turn timeout is not really needed as benchmark cannot progress until the turn response is acquired. We set it to a high value to avoid accidentally skipping turns.

src/inference_endpoint/config/schema.py
hvagadia: Not as per current definition, I am removing this as an option.

src/inference_endpoint/evaluation/multi_turn_inline_accuracy.py
hvagadia: This is not a big issue - file size is ~10mb for one full pass of the dataset which is reasonable.

src/inference_endpoint/commands/benchmark/execute.py
hvagadia: I have integrated inline accuracy into scorer.py

viraatc (commented):

Reviewed by: **Codex** + **Claude (fable)** | Depth: **thorough** | Focus: simplicity & correctness (unnecessary `getattr`/`setattr`, workarounds, antipatterns)

Posting 10 inline comments (11 findings total; 1 outside diff hunks, see summary). 2 additional findings were dropped as already raised in existing threads.

src/inference_endpoint/load_generator/multi_turn_strategy.py
```diff
@@ -305,27 +321,80 @@ def _issue_turn_now(self, conv_id: str, idx: int, turn: int) -> None:
...
+                    str(repeat_id).encode("utf-8"), digest_size=2
+                ).hexdigest()
+                conv_salt = hashlib.blake2b(
+                    conversation_id.encode("utf-8"), digest_size=2
+                ).hexdigest()
+                message["content"] = (
+                    f"[salt: {repeat_salt}]\n\n" f"{content}\n\n" f"[salt: {conv_salt}]"
+                )
+                return salted_messages
+        raise InputValidationError(
```
viraatc, line 400: [Claude] **high · bug**: `_messages_with_trajectory_salt` raises `InputValidationError` mid-run when a conversation lacks a system prompt. On the delayed path — `self._loop.call_later(delay, self._issue_turn_now, conv_id, idx, turn)` (lines 280-282, taken whenever `inject_tool_delay` yields a positive delay, which the kimi config enables) — the exception escapes into a bare asyncio timer callback: `_issue_turn_now` → `_build_data_override` → this raise, with no try/except (unlike the sync path, which is caught at lines 630-636). Neither `self._error` nor `self._all_done` is set, and the conversation's `_active_iters` cursor was already advanced (line 308), so the slot leaks and `execute()` awaits `_all_done` forever — with the default `max_duration_ms=0` the benchmark hangs indefinitely.

Root-cause fix: all system prompts are already known at construction via `pre_built_messages_by_key`, so validate "every conversation has a system message" once up front (in `__init__`/`execute()` when `enable_salt` is set) instead of per-turn at issue time — and wrap the timer-invoked body of `_issue_turn_now` in the same error→`_all_done` protection the sync path has.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -871,6 +874,50 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
         }
         logger.info(f"Score for {eval_cfg.dataset_name}: {score} ({n_repeats} repeats)")
 
+    perf_cfg = next(
+        (ds for ds in config.datasets if ds.type == DatasetType.PERFORMANCE),
+        None,
+    )
+    if perf_cfg is not None and perf_cfg.accuracy_config is not None:
```
viraatc: [Claude] **high · error-handling**: This perf-dataset scoring block defers all scorer-config validation to `finalize_benchmark` — after the run has completed. `Scorer.get(...)` (KeyError on an unknown `eval_method`), the `REQUIRES_EXTRACTOR` raise (line 891), and scorer-constructor errors all fire here, aborting finalize before `results.json` is written (lines ~950-976) — destroying the output of a potentially hours-long run over a config typo. The sibling accuracy-dataset path validates exactly this at setup in `_load_datasets` (lines 249-286), so misconfiguration fails fast before any traffic is issued.

The block is also a near-copy of both the `_load_datasets` validation and the `eval_cfg` loop just above (and silently ignores `num_repeats`, unlike that loop). Suggested fix: when the performance dataset carries `accuracy_config`, validate it in `_load_datasets` and append an `AccuracyConfiguration(dataset_name="performance", ...)` to `eval_configs`, letting the existing finalize loop score it. Note this also makes a previously-ignored config meaningful: `accuracy_config` on a `type: performance` dataset used to be a no-op; now any pre-existing config that set it gets post-run scoring.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -314,6 +318,498 @@ def score(self) -> tuple[float, int]:
...
+                    {
+                        "conversation_id": conversation_id,
+                        "turn": ground_truth["_assistant_turn"],
+                        "domain": domain,
+                        "exclude_reason": "no ground truth",
+                    }
+                )
+
+        model_turns = self._completed_model_turns(set(scorable_expected))
+        observed_repeats = sorted(
```
viraatc: [Codex + Claude] **high · data-integrity**: Fully-failed repeats silently vanish from the validity check. `observed_repeats` keeps only repeats with at least one truthy model output:

```python
observed_repeats = sorted({key[1] for key, model in model_turns.items() if model})
repeats_to_score = observed_repeats or [1]
```

and `expected_outputs` (lines 569-573) is built only over `repeats_to_score`. If a run is configured with `num_trajectories_to_issue = 2 × dataset` but every repeat-2 trajectory aborts or times out (synthetic COMPLETE records carry `data=None`, so `model` is falsy), repeat 2 drops out of `expected_outputs` entirely: `missing_outputs` stays 0 and the result is reported `valid: true`, scoring only repeat 1 — a perfect score despite half the configured trajectories producing nothing. The scorer never learns the intended repeat count, so it cannot detect this. Pass the configured `num_trajectories_to_issue` (or the derived repeat count) into the scorer and validate `len(observed_repeats)` against it, or at minimum record the discrepancy in `scores.json`.

src/inference_endpoint/load_generator/multi_turn_strategy.py
```diff
@@ -336,6 +405,42 @@ def _fill_slot(self) -> None:
...
+                return
+        self._fill_slot()
+
+    def _stop_performance_tracking_once(self) -> None:
+        if self._performance_tracking_stopped:
+            return
+        self._performance_tracking_stopped = True
+        if self._phase_issuer is None:
+            return
+        stop_tracking = getattr(self._phase_issuer, "stop_performance_tracking", None)
```
viraatc: [Claude] **medium · design**: Unnecessary dynamic attribute access:

```python
stop_tracking = getattr(self._phase_issuer, "stop_performance_tracking", None)
```

This PR itself adds `stop_performance_tracking()` to `PhaseIssuerProtocol` (strategy.py:87-89) and implements it on `PhaseIssuer` (session.py:216). The only object lacking the method is the pre-existing `FakePhaseIssuer` test stub (the new `RecordingPhaseIssuer` in the same test file does implement it). Production code is carrying a defensive guard to tolerate an incomplete test fake — and the guard silently swallows the missing-method case, so a real issuer missing the protocol method would never emit STOP_PERFORMANCE_TRACKING and nobody would know. Call `self._phase_issuer.stop_performance_tracking()` directly and add the method to `FakePhaseIssuer`.

src/inference_endpoint/load_generator/multi_turn_strategy.py
```diff
@@ -305,27 +321,80 @@ def _issue_turn_now(self, conv_id: str, idx: int, turn: int) -> None:
...
+        turn: int,
+        repeat_id: int,
+    ) -> dict[str, Any] | None:
+        if not self._enable_salt:
+            return None
+
+        messages = self._dataset_metadata.pre_built_messages_by_key.get(
+            (source_id, turn)
+        )
+        if not messages:
```
viraatc, line 361: [Claude] **medium · design**: `_build_data_override` silently disables salting when pre-built messages are missing:

```python
messages = self._dataset_metadata.pre_built_messages_by_key.get((source_id, turn))
if not messages:
    return None
```

For a real `MultiTurnDataset`, `pre_built_messages_by_key` always has an entry for every client turn (multi_turn_dataset.py:181-183), so a miss means corrupted/inconsistent metadata — yet the turn is then issued completely unsalted, violating the `enable_salt: true` invariant the README declares mandatory for official runs, with no log or error. This guard papers over a condition the upstream dataset loader is supposed to make impossible. Either raise (consistent with the no-system-prompt case directly below) or validate once at `execute()` start that every `(conversation_id, turn)` has pre-built messages.

src/inference_endpoint/load_generator/session.py
```diff
@@ -609,12 +625,17 @@ def _make_stop_check(
             else 0
         )
         total_samples = settings.total_samples_to_issue()
+        stop_on_sample_count = not (
```
viraatc, line 628: [Claude] **medium · api-contract**: ```python
stop_on_sample_count = not (
    settings.load_pattern is not None
    and settings.load_pattern.type == LoadPatternType.MULTI_TURN
)
```

disables the entire `issued_count >= total_samples` stop for multi-turn — including the explicit user override: `RuntimeSettings.total_samples_to_issue()` returns `n_samples_to_issue` as priority 1 (runtime_settings.py:178), so an explicitly-set `settings.runtime.n_samples_to_issue` is now silently ignored in multi-turn runs (previously it capped issuance). Either keep honoring an explicit `n_samples_to_issue` and bypass only the dataset/duration-derived default, or reject `n_samples_to_issue` for multi-turn in the config model validator so users get an error instead of a silently dead knob.

Also untested: the multi-turn integration tests all build `RuntimeSettings` with `LoadPatternType.MAX_THROUGHPUT` (test_multi_turn.py:155, 417, 724), so the `stop_on_sample_count=False` branch is never executed by the suite.

File: src/inference_endpoint/evaluation/scoring.py [hunk 27]
```diff
@@ -314,6 +318,498 @@ def score(self) -> tuple[float, int]:
...
+    _SHELL_WRAPPERS: ClassVar[set[str]] = {
+        "env",
+        "time",
+        "nice",
+        "sudo",
+        "exec",
+        "command",
+    }
+    _REPEAT_SUFFIX_RE: ClassVar[re.Pattern[str]] = re.compile(r"__repeat_(\d+)$")
+    _WORKFLOW_CONVERSATION_RE: ClassVar[re.Pattern[str]] = re.compile(r"^sim_\d+$")
```
viraatc, line 394: [Claude] **medium · design**: Two fragile string contracts are baked into the generic evaluation module:

1. `_WORKFLOW_CONVERSATION_RE = re.compile(r"^sim_\d+$")`, the intent-code regexes, and the large `_EXECUTABLE_ALIASES` table hardcode one specific dataset's naming/labeling conventions into `evaluation/scoring.py`. Any other multi-turn dataset run with `multi_turn_inline` is silently classified "coding" and scored 0/excluded, with no configuration escape hatch — these would fit naturally in `accuracy_config.extras`, which the scorer already accepts.
2. The repeat id is round-tripped through the conversation-id string: the strategy encodes `f"{source_id}__repeat_{instance_id}"` (multi_turn_strategy.py:216-218) and the scorer regex-recovers it via `_REPEAT_SUFFIX_RE`. A dataset whose own `conversation_id` legitimately ends in `__repeat_<n>` is silently split and mis-attributed to the wrong source conversation/repeat. `EventRecord` already carries structured `conversation_id`/`turn`; encoding structured data into a parsed string suffix is a smell — at minimum the strategy should reject source conversation ids matching the reserved suffix.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -644,6 +644,9 @@ async def _run_benchmark_async(
                 dataset_metadata=ctx.dataloader.conversation_metadata,
                 multi_turn_config=mt_cfg,
                 target_concurrency=ctx.config.settings.load_pattern.target_concurrency,
+                num_trajectories_to_issue=(
```
viraatc: [Claude] **low · design**: Redundant state duplication: `num_trajectories_to_issue=(mt_cfg.num_trajectories_to_issue if mt_cfg is not None else None)` is passed alongside `multi_turn_config=mt_cfg` in the same constructor call. Inside `MultiTurnStrategy.__init__`, the sibling knobs (`stop_issuing_on_first_user_complete`, `turn_timeout_s`, `enable_salt`) are all read from `multi_turn_config`, while this one alone arrives as a separate parameter that can disagree with `multi_turn_config.num_trajectories_to_issue`. Same simplification already requested for `stop_issuing_on_first_user_complete` in an earlier review thread — drop the extra parameter and read it from the config like its siblings.

src/inference_endpoint/load_generator/multi_turn_strategy.py
```diff
@@ -16,28 +16,29 @@
...
 
 # Default turn timeout when no MultiTurnConfig is provided.
-_DEFAULT_TURN_TIMEOUT_S = 300.0
+_DEFAULT_TURN_TIMEOUT_S = 86400.0
 
 ConversationTurn = tuple[int, int]
 ConversationTurns = list[ConversationTurn]
-PendingConversation = tuple[str, ConversationTurns]
-ActiveConversationState = tuple[ConversationTurns, int]
+ActiveConversationState = tuple[str, ConversationTurns, int, int]
```
viraatc, line 40: [Claude] **low · design**: `ActiveConversationState = tuple[str, ConversationTurns, int, int]` is an anonymous positional 4-tuple, unpacked with throwaway names at five sites (`source_id, turns, cursor, repeat_id = ...` at lines 263/296, rebuilt at 308, `_source_id, turns, cursor, _repeat_id` at 546) and hand-seeded in tests (`("conv-x", [(1, 2), (2, 3)], 0, 1)`); same for `ConversationInstance` (line 41), which callers immediately 4-way unpack. A `NamedTuple` or `dataclass(slots=True)` (the repo's stated preference for frequently-instantiated classes) would name the fields and prevent positional mix-ups of the two adjacent ints (`cursor` vs `repeat_id`).

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -314,6 +318,498 @@ def score(self) -> tuple[float, int]:
...
+        assert (
+            self.dataset.dataframe is not None
+        ), f"Dataset {self.dataset} has no dataframe loaded"
+
+        rows_by_conversation: dict[str, list[dict[str, Any]]] = defaultdict(list)
+        for raw_row in self.dataset.dataframe.to_dict("records"):
+            row: dict[str, Any] = {}
+            for field, value in raw_row.items():
+                try:
+                    row[field] = None if value != value else value
```
viraatc, line 617: [Claude] **low · design**: The NaN scrub relies on the `x != x` trick plus exception side-channels:

```python
row[field] = None if value != value else value
```

wrapped in `try/except (TypeError, ValueError)` — non-scalar cells (e.g. numpy arrays, "truth value is ambiguous") raise into the except and are silently kept. The sibling module already established a cleaner idiom for this same dataframe: `isinstance(val, float) and pd.isna(val)` (multi_turn_dataset.py:166, 187-189) — use it for homogeneity and to avoid exception-driven control flow.

Relatedly, the sort key at line 642 (`int(row.get("turn") or 0)`) sits outside the `try/except (TypeError, ValueError)` that protects the same conversion in the pairing loop (lines 648-652), so a single non-numeric `turn` cell crashes `score()` at finalize instead of being skipped.

viraatc:

Reviewed by: **Codex** + **Claude (fable)** | Depth: **thorough** | Focus: simplicity & correctness (unnecessary `getattr`/`setattr`, workarounds, smelly antipatterns)

Found **11** issues across **6** files (10 posted inline; #11 falls outside the diff hunks). Two further findings (`trust_remote_code=True` in `token_metrics.py`, and the `stop_on_first_empty_slot` schema test adding no coverage) were **dropped as duplicates** of existing review threads.

### 🔴 Must Fix (high)
Issues that will cause hangs, lost results, or silently wrong benchmark scores.

| 1 | `load_generator/multi_turn_strategy.py` | 385 | bug | Claude | Salt validation raises inside a `call_later` timer callback with no error→`_all_done` protection — slot leaks, `execute()` hangs forever; validate system prompts once up front instead |
| 2 | `commands/benchmark/execute.py` | 881 | error-handling | Claude | Perf-dataset scorer config validated only at finalize — a config typo aborts before `results.json` is written, destroying an hours-long run; validate in `_load_datasets` like the sibling accuracy path |
| 3 | `evaluation/scoring.py` | 503 | data-integrity | **Both** | A fully-failed repeat (all `model is None`) vanishes from `expected_outputs` — run reports `valid: true` with a perfect score despite half the configured trajectories producing nothing |

### 🟡 Should Fix (medium)
Real issues triggering under specific conditions, or design flaws that will compound.

| 4 | `load_generator/multi_turn_strategy.py` | 424 | design | Claude | Unnecessary `getattr(..., "stop_performance_tracking", None)` guard — the method is on the protocol this PR adds; the guard only tolerates an incomplete test fake and silently swallows real omissions |
| 5 | `load_generator/multi_turn_strategy.py` | 349 | design | Claude | Missing pre-built messages silently disables salting (no log, no error), violating the `enable_salt` invariant the README declares mandatory for official runs |
| 6 | `load_generator/session.py` | 628 | api-contract | Claude | `stop_on_sample_count=False` for multi-turn silently ignores an explicitly-set `n_samples_to_issue`; branch also never exercised by the test suite |
| 7 | `evaluation/scoring.py` | 394 | design | Claude | Dataset-specific string contracts baked into the generic scorer: `sim_\d+` regex + alias table with no config escape hatch, and repeat id round-tripped through a `__repeat_N` conversation-id suffix |

### 🔵 Consider (low)

| 8 | `commands/benchmark/execute.py` | 647 | design | Claude | `num_trajectories_to_issue` passed as a separate ctor param alongside `multi_turn_config` that already carries it — same simplification viraatc requested for `stop_issuing_on_first_user_complete` |
| 9 | `load_generator/multi_turn_strategy.py` | 40 | design | Claude | `ActiveConversationState`/`ConversationInstance` are anonymous positional 4-tuples unpacked at 5+ sites; a `NamedTuple`/`dataclass(slots=True)` would name fields and prevent `cursor`/`repeat_id` mix-ups |
| 10 | `evaluation/scoring.py` | 633 | design | Claude | `x != x` NaN scrub with exception side-channels (sibling module already uses `pd.isna`); sort key `int(row.get("turn") or 0)` at line 642 is outside the try/except, so one bad cell crashes `score()` at finalize |
| 11 | `load_generator/conversation_manager.py` | 53 | documentation | Claude | *(not in diff — couldn't post inline)* Stale docstring: "All states are pre-created by MultiTurnStrategy.execute()…" — pre-creation is gone; states are now created lazily per trajectory, including mid-run `__repeat_N` instances. The no-locking conclusion still holds (single event loop) but for a different reason; update the rationale |

nvzhihanj (approved):

src/inference_endpoint/evaluation/scoring.py
hvagadia: I am gonna leave this as is for now, we can handle it if we expand the dataset in the future.

PR #329, opened 2026-06-02
feat: show progress bar during multi-turn ISL pre-compute

YibaiMeng:
## What does this PR do?

Show a `tqdm` progress bar over the multi-turn ISL pre-computation loop in `_precompute_isl_for_multi_turn` (`src/inference_endpoint/commands/benchmark/execute.py`).

*Why*: On large multi-turn datasets the serial `apply_chat_template` loop runs for minutes with no output (30 min on a 26k-turn agentic-coding dataset, with the entire benchmark taking less than 50 minutes), so it looks like a hang.  This PR adds a progress bar.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactor/cleanup

## Related issues

N/A

## Testing

- [ ] Tests added/updated
- [x] All tests pass locally
- [x] Manual testing completed

## Checklist

- [x] Code follows project style
- [x] Pre-commit hooks pass
- [x] Documentation updated (if needed)

PR #336, opened 2026-06-04
fix: parallel precompute for multi-turn datasets

tianmu-li:
## What does this PR do?

ISL pre-compute for multi-turn runs single-threaded, which can take a long time for large datasets. This PR parallelizes the pre-compute part and adds a progress bar similar to https://github.com/mlcommons/endpoints/pull/329

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [x] Tests added/updated
- [x] All tests pass locally

## Checklist

- [x] Pre-commit hooks pass


YibaiMeng:
Thanks for the quick fix Tianmu, I'll test it and report back.

I took another look and it seems single-threading wasn't the only issue?  The total work seems to be
O(T²) per conversation. `_precompute_isl_for_multi_turn` runs `apply_chat_template` on the entire `dataloader.data` array, onces per conversation. Looking the dataloader class [`MultiTurnDataset`](https://github.com/mlcommons/endpoints/blob/9333852ed0af84fbe3dc8d421872ce63806402e3/src/inference_endpoint/dataset_manager/multi_turn_dataset.py), we can see `data` is the collections of `conversation_metadata.pre_built_messages_by_key[conversation_id, turn]` (L546, L606-L612). This `pre_built_messages_by_key` holds one entry per client turn (L402-L469). So a trajectory's entries carry message-lists of length 1,3,…,2T−1, `_precompute_isl_for_multi_turn` tokenizes each in full, leading to reprocessing of tokens.

tianmu-li:
> Thanks for the quick fix Tianmu, I'll test it and report back.
> 
> I took another look and it seems single-threading wasn't the only issue? The total work seems to be O(T²) per conversation. `_precompute_isl_for_multi_turn` runs `apply_chat_template` on the entire `dataloader.data` array, onces per conversation. Looking the dataloader class [`MultiTurnDataset`](https://github.com/mlcommons/endpoints/blob/9333852ed0af84fbe3dc8d421872ce63806402e3/src/inference_endpoint/dataset_manager/multi_turn_dataset.py), we can see `data` is the collections of `conversation_metadata.pre_built_messages_by_key[conversation_id, turn]` (L546, L606-L612). This `pre_built_messages_by_key` holds one entry per client turn (L402-L469). So a trajectory's entries carry message-lists of length 1,3,…,2T−1, `_precompute_isl_for_multi_turn` tokenizes each in full, leading to reprocessing of tokens.

One concern with incremental tokenization (tokenizing only the diff between turns) is that the number of tokens depends on text boundary, so doing it incrementally can result in a different number of tokens computed versus tokenizing the entire prompt, which is what the model/serving framework actually sees.

YibaiMeng:
But the turns are separated by `<|im_start|>` and `<|im_end|>` right? These are special tokens, so tokenization of texts within boundaries shouldn't matter even if we tokenize the entire prompt at once? 

I agree that making sure the number matches what the framework actually sees is more important than perf and ATM the wasted time is acceptable. Though I wonder if it would be feasible to add a field to the benchmark datasets about the token count.

src/inference_endpoint/commands/benchmark/execute.py
nvzhihanj, line 321: Hi @tianmu-li , can you remind me why the tokenization of input is needed on the client side?
Ideally we just need to send prepped text prompt as packet to server, and server should handle the incremental tokenization.

I don't seem to understand why it's needed on client slide

nvzhihanj: (I.e. I would suggest that we remove the tokenization entirely, or just tokenize once)

tianmu-li: v1/chat/completions (and v1/responses) does not accept token input, only text input. Tokenized input would require using v1/completions or sglang api, which is mostly deprecated or backend-specific.

tianmu-li:
> But the turns are separated by `<|im_start|>` and `<|im_end|>` right? These are special tokens, so tokenization of texts within boundaries shouldn't matter even if we tokenize the entire prompt at once?
> 
> I agree that making sure the number matches what the framework actually sees is more important than perf and ATM the wasted time is acceptable. Though I wonder if it would be feasible to add a field to the benchmark datasets about the token count.

I also thought about adding token count to the dataset itself so that getting ISL is just a lookup, but that will only be accurate for one model, and a different model will require calculating ISL again.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -362,26 +373,47 @@ def _precompute_isl_for_multi_turn(
...
                 logger.exception(
                     "ISL pre-computation: apply_chat_template failed (first failure shown)"
                 )
-                first_failure_logged = True
-            skipped += 1
+            return None
+
+    n_workers = min(os.cpu_count() or 4, 16)
+    skipped = 0
+    with ThreadPoolExecutor(
```
wangshangsam, line 386: If tokenization is belived to be CPU compute-bound, would `ProcessPoolExecutor` be a better choice here?

tianmu-li: Let me try. Fast tokenizer doesn't release GIL until after chat template, so ProcessPoolExecutor might help

wangshangsam:
> But the turns are separated by `<|im_start|>` and `<|im_end|>` right? These are special tokens, so tokenization of texts within boundaries shouldn't matter even if we tokenize the entire prompt at once?

I feel that this could work. We can validate it through a full run with the following assertion

```python
# spliced (cut at every <|im_start|>/<|im_end|>, add_special_tokens=False per span)
# must equal the full pass:
assert spliced_ids == tok.apply_chat_template(messages, tokenize=True, add_generation_prompt=True)
```

Athough, this also really depends on the model (i.e., the tokenizer and the chat template), BPE + ChatML would be safe. Others (like SentencePiece), not necessarily. It probably won't look very clean, either.

src/inference_endpoint/commands/benchmark/execute.py
hvagadia: Since we are supporting only v1/chat/completions, we don't need tokenization on client right?

tianmu-li: Correct. The precompute part is just to calculate ISL. During benchmark, server still sends text+tool+reasoning directly

hvagadia: Do we need the ISL for anything specific? Seems like all the metrics are only OSL dependent. Can we just remove tokenization and not worry about doing it incrementally?

src/inference_endpoint/commands/benchmark/execute.py
tianmu-li: Benchmark pins first 5 cores by default, and with that there isn't not much difference between ThreadPool and ProcessPool. In fact, 16 workers oversubscribe cpus.

tianmu-li:
Fixed in #336

PR #337, opened 2026-06-05
fix: Tunable drain timeout

arekay-nv:
## What does this PR do?
Plugs in the drain timeout from CLI/config to the aggregator so that we can adjust how long to wait particularly for high concurrency workloads.
A value of 0 means to wait indefinitely.
## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass


src/inference_endpoint/config/templates/concurrency_template_full.yaml
```diff
@@ -78,6 +78,7 @@ settings:
     warmup_timeout_s: 240.0  # Warmup drain timeout in seconds (None = wait indefinitely)
     performance_timeout_s: 240.0  # Performance drain timeout in seconds (None = wait indefinitely)
     accuracy_timeout_s: null  # Accuracy drain timeout in seconds (None = wait indefinitely)
+    metrics_drain_timeout_s: 60.0  # Wall-clock budget (seconds) for the metrics aggregator to drain in-flight tokenize tasks after ENDED (default: 60.0; 0 = unlimited).
```
nvzhihanj, line 81: From design principle I am leaning against adding even more timeout flags, which becomes increasingly user unfriendly and silently hide wrong metrics.

I am thinking we should either accelerate the metrics aggregator / tokenizer such that it will not be the bottleneck (either MP or rust tokenizer), or make default timeout null (but add a hanging error out). A silent but wrong metrics will harm more than hanging indefinitely IMO.

arekay-nv: I agree with that, but for now i think making the drain tunable is the best option since the default is already set. Post submission we can remove these for a more unified approach - possibly with a single global timeout. The PR will enable us to support longer drains while keeping the default the same. 
If we can do without this PR - that would be great.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -673,6 +673,15 @@ async def _run_benchmark_async(
             aggregator_args.append("--streaming")
         if ctx.tokenizer_name is not None:
             aggregator_args.extend(["--tokenizer", ctx.tokenizer_name])
+        aggregator_args.extend(
+            ["--drain-timeout", str(config.settings.drain.metrics_drain_timeout_s)]
+        )
```
nv-alicheng, line 678: these 2 .extends can be merged can't they

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
```diff
@@ -137,8 +137,8 @@ async def main() -> None:
             "Wall-clock budget (seconds) to wait for in-flight async tokenize "
             "tasks to finish after ENDED before the aggregator cancels them "
             "and emits the final snapshot with n_pending_tasks > 0 "
-            "(default: 60.0). Increase for long-context / low-worker-count "
-            "tokenize workloads."
+            "(default: 60.0; 0 = wait indefinitely). Increase for long-context "
```
nv-alicheng, line 140: suggest bumping the values - 300s, 8 workers maybe?

viraatc (approved):

src/inference_endpoint/config/templates/concurrency_template_full.yaml
nvzhihanj: Sure, please file an issue, and document it in the post-mortem so it's not lost

nvzhihanj (approved):

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
arekay-nv: Keeping the defaults as-is for now, will consolidate via #341 where we can discuss better defaults as well.

PR #335, opened 2026-06-04
feat: Create and push LCB container

arekay-nv:
## What does this PR do?

Adds script to build and create LCB container which can be used by the accuracy implementations to get LCB score. The implementation uses the `src/inference_endpoint/evaluation/livecodebench/lcb_serve.dockerfile` to build the container and push to the specified repository. 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally
- [ ] Manual testing completed

## Checklist

- [ ] Code follows project style
- [ ] Pre-commit hooks pass
- [ ] Documentation updated (if needed)

PR #340, opened 2026-06-05
feat : deepseek r1 accuracy support

arekay-nv:
## What does this PR do?
Adds accuracy support for Deepseek-r1 using an ensemble evaluator that will produce a combined score. 
Needs #335 to push the container image.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass


src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1157,3 +1216,373 @@ def score(self) -> tuple[float | None, int]:
...
+            text_correct,
+            text_n,
+            lcb_passed,
+            lcb_total,
+            complete,
+        )
+
+        if combined is None:
+            return None, n_repeats
+        return float(combined), n_repeats
```
arekay-nv, line 2089: [HIGH] **Partial score returned without caller-visible flag when LCB is unreachable.**

`score()` returns `float(combined)` here even when the LCB container was unreachable and only text subsets were graded. The caller in `execute.py:913` stores this as `"score": score` in `results.json` with no indication of partial coverage. The `complete=False` flag lives only in the separate `deepseek_eval/*_results.json` artifact — consumers of `results.json` cannot distinguish a full 5-subset score from a text-only 4-subset score.

Consider returning `None` when `not lcb_ok` (forcing callers to treat it as unscored), or adding a `complete` field to the `accuracy_scores` dict in `execute.py` so the incompleteness is surfaced in the top-level report.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1157,3 +1216,373 @@ def score(self) -> tuple[float | None, int]:
...
+        text_complete = True
+        for sub, d in per_dataset.items():
+            if sub == self.lcb_subset:
+                continue
+            em = d.get("exact_match")
+            n = int(d.get("num_samples", 0))
+            if em is None:  # subset failed to grade (status != "ok")
+                text_complete = False
+                continue
+            text_correct += round(em / 100.0 * n)
```
arekay-nv, line 2020: [MEDIUM] **Floating-point round-trip may reconstruct incorrect sample count.**

`text_correct += round(em / 100.0 * n)` recovers a per-subset correct count by inverting the runner's `em = 100.0 * k / n` formula. This round-trips through JSON serialization in the subprocess, where float precision may cause `round()` to yield `k±1` for certain `(k, n)` pairs. A more reliable fix is to have the runner also emit `"correct_samples": k` in each `per_dataset` entry so the aggregation is exact integer arithmetic.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1157,3 +1216,373 @@ def score(self) -> tuple[float | None, int]:
...
+            return None
+        total_samples = int(result.get("total_samples", 0))
+        per_problem = result.get("results", {})
+        if not per_problem and total_samples:
+            logger.error(
+                "lcb-service evaluated %d samples but returned an empty summary",
+                total_samples,
+            )
+            return None
+        passed = sum(sum(code_passed) for code_passed in per_problem.values())
```
arekay-nv, line 1934: [MEDIUM] **Unvalidated WebSocket response structure — malformed response raises `TypeError` instead of returning `None`.**

`sum(sum(code_passed) for code_passed in per_problem.values())` assumes every value in `per_problem` is an iterable of booleans. If the lcb-service returns a non-list for any question key (e.g. a scalar or dict), `sum(code_passed)` raises `TypeError` that propagates unhandled out of `_score_lcb_via_container`, causing `score()` to raise rather than fall back gracefully via `None`. Wrap this block in a `try/except (TypeError, ValueError)` or validate the schema of `per_problem` before iterating.

examples/10_DeepSeekR1_Example/accuracy/deepseek_eval_runner.py
```diff
@@ -0,0 +1,188 @@
...
+logger = logging.getLogger("deepseek_eval_runner")
+
+
+def _load_evaluator(eval_dir: Path):
+    """Import the fetched MLCommons ``eval_accuracy`` module."""
+    if not (eval_dir / "eval_accuracy.py").exists():
+        raise FileNotFoundError(
+            f"eval_accuracy.py not found at {eval_dir}. Run setup_eval.sh first."
+        )
+    sys.path.insert(0, str(eval_dir))
```
arekay-nv, line 61: [MEDIUM] **`sys.path.insert(0, ...)` with a user-supplied `--eval-dir` allows arbitrary code execution.**

`sys.path.insert(0, str(eval_dir))` followed by `import eval_accuracy` executes whichever `eval_accuracy.py` is found first on `sys.path`. Because `--eval-dir` is an exposed CLI argument, pointing it at a directory containing a malicious `eval_accuracy.py` leads to arbitrary code execution in the runner subprocess. Inserting at index 0 also permanently pollutes `sys.path` for the process lifetime.

Replace with `importlib.util.spec_from_file_location` loading from the explicit absolute path `eval_dir / 'eval_accuracy.py'`, which avoids both the path-injection risk and the `sys.path` side-effect.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1157,3 +1216,373 @@ def score(self) -> tuple[float | None, int]:
...
+    0-100 scale as the MLPerf golden accuracy (81.3582), or ``(None, n)`` if
+    no successful output was produced or the subprocess fails to yield a
+    usable number - matching the ``Scorer.score()`` contract.
+
+    Reads the subset id from ``dataset.dataframe[subset_column]`` (default
+    column ``dataset``) and the per-sample question from ``question_column``
+    (default ``question``); both are passed through ``accuracy_config.extras``.
+    """
+
+    REQUIRES_EXTRACTOR: ClassVar[bool] = False
```
arekay-nv, line 1766: [LOW] **`REQUIRES_EXTRACTOR = False` but LCB extraction is hardcoded inside `_score_lcb_via_container`.**

Setting `REQUIRES_EXTRACTOR = False` prevents the scorer framework from injecting an extractor, but `_score_lcb_via_container` hardcodes `PythonCodeExtractor.extract` directly. This means `accuracy_config.extractor` cannot override the LCB code-extraction strategy, inconsistent with how `LiveCodeBenchScorer` handles the same concern via its injected extractor. At minimum, add a comment explaining why the extractor is bypassed here; ideally respect `self.extractor` when set.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1157,3 +1216,373 @@ def score(self) -> tuple[float | None, int]:
...
+        )
+        results = msgspec.json.decode(out_json.read_bytes())
+        per_dataset = results.get("per_dataset", {})
+
+        # Aggregate every text subset the runner graded. Track whether any
+        # failed so a partial run is never reported as a complete score.
+        text_correct = 0
+        text_n = 0
+        text_complete = True
+        for sub, d in per_dataset.items():
```
arekay-nv, line 2012: [LOW] **Customised `lcb_subset` key may not match the runner's emitted key, causing text-subset mis-classification.**

`if sub == self.lcb_subset: continue` relies on `self.lcb_subset` matching the key the runner writes to `per_dataset`. If `lcb_subset` is set to a non-default value (e.g. `"livecodebench_v3"`) while the runner still emits `"livecodebench"`, the runner's entry passes the guard and is treated as a text subset. Its `exact_match=None` sets `text_complete = False`, marking the combined score incomplete even when all text subsets graded successfully. Consider making the skip-condition match on the actual runner entry key, or validate at construction that `lcb_subset` matches the runner's expected output.

arekay-nv (commented):
## Review Council — Claude Code Review

Found **6 issues** (Codex review unavailable — workspace auth error; Claude-only).

arekay-nv:

Reviewed by: **Claude** | Depth: **thorough** | Codex: unavailable (workspace auth error)

Found **6 issues** across **2 files**: 1 high · 3 medium · 2 low

### 🔴 Must Fix (high)

| 1 | `src/inference_endpoint/evaluation/scoring.py` | 1588 | data-integrity | Partial text-only score returned as `float` when LCB container unreachable — callers cannot distinguish it from a full 5-subset score |

### 🟡 Should Fix (medium)

| 2 | `src/inference_endpoint/evaluation/scoring.py` | 1519 | bug | Float round-trip `round(em / 100.0 * n)` may reconstruct k±1 correct count after JSON serialization |
| 3 | `src/inference_endpoint/evaluation/scoring.py` | 1433 | bug | `sum(sum(code_passed))` raises `TypeError` on malformed WebSocket response instead of returning `None` |
| 4 | `examples/10_DeepSeekR1_Example/accuracy/deepseek_eval_runner.py` | 61 | security | `sys.path.insert(0, eval_dir)` with user-controlled `--eval-dir` enables arbitrary code execution |

### 🔵 Consider (low)

| 5 | `src/inference_endpoint/evaluation/scoring.py` | 1265 | design | `REQUIRES_EXTRACTOR=False` but LCB extraction hardcodes `PythonCodeExtractor` — `accuracy_config.extractor` cannot override it |
| 6 | `src/inference_endpoint/evaluation/scoring.py` | 1511 | bug | Customised `lcb_subset` key mismatch with runner output marks combined score incomplete when all text subsets succeed |

nv-alicheng (commented):
## Review Council — Multi-AI Code Review (thorough)

Reviewed by **Codex** + **Claude**. Posting 6 inline findings that are **not already covered** by the existing automated review on this PR (the headline partial-score issue at `scoring.py:1588`, the rounding round-trip at `1519`, the `sys.path` RCE at `deepseek_eval_runner.py:61`, and the temp-file leak at `run_client.sh:43` are already flagged — see the synthesis comment for how they fit together).

examples/10_DeepSeekR1_Example/offline_deepseek_r1_accuracy_smoke.yaml
```diff
@@ -0,0 +1,56 @@
...
+# deepseek_r1 scorer) before committing to the full 4388-sample run. Not an
+# official accuracy number (truncated sample set + shorter max_new_tokens).
+
+name: "offline-deepseek-r1-fp4-accuracy-smoke"
+version: "1.0"
+type: "offline"
+
+model_params:
+  # Served model id = the checkpoint dir basename trtllm-serve registers.
+  name: "${SERVED_MODEL_NAME:-deepseek_r1-torch-fp4-v2}"
```
nv-alicheng, line 12: [Codex] **medium · bug**: The smoke config defaults `model_params.name` to `deepseek_r1-torch-fp4-v2`, but this PR's own docs state that the `-v2` checkpoint does not load with the stock TRT-LLM builds and that `deepseek_r1-torch-fp4` is the working default. A user who runs the smoke path without exporting `SERVED_MODEL_NAME` will send requests for a model id the server can't serve, and the run fails immediately — i.e. the advertised quick-validation path is broken out of the box. Default to `deepseek_r1-torch-fp4`.

examples/10_DeepSeekR1_Example/README.md
```diff
@@ -0,0 +1,260 @@
...
+      # LiveCodeBench via the lcb_serve container (optional; defaults shown):
+      lcb_websocket_port: 13835 # ws://localhost:13835; null = in-process
+      lcb_subset: livecodebench # graded out-of-band via the container
+      lcb_timeout: 60 # per-sample code-exec timeout (s)
+```
+
+`extras` are forwarded to the scorer. **`lcb_websocket_port` defaults to 13835**,
+so if the `lcb_serve` container is reachable on the client node the livecodebench
+subset is scored **in-run** (one number, no follow-up job); if it's unreachable
+the scorer logs and falls back to grading LCB in-process. The LCB keys are
```
nv-alicheng, line 186: [Claude] **medium · documentation**: This sentence says that when the `lcb_serve` container is unreachable "the scorer logs and falls back to grading LCB in-process," but that is not what the code does for the common case (port configured, container down): `_score_lcb_via_container` returns `None` and the scorer leaves livecodebench **unscored** — see `scoring.py:1527-1543` ("Do NOT re-run the in-process executor"). In-process grading happens only when `lcb_websocket_port` is `null` (the `use_container == False` path). README lines 201-202 below correctly describe the unscored/`complete: false` behavior, so the doc contradicts both itself and the implementation. A reader following this line will expect an LCB score that never appears.

tests/unit/evaluation/test_deepseek_r1_scorer.py
```diff
@@ -0,0 +1,408 @@
...
+                        "per_dataset": per,
+                    }
+                )
+            )
+            return MagicMock(returncode=0, stdout="ok\n")
+
+        monkeypatch.setattr(scoring_mod.subprocess, "run", fake_run)
+        return calls
+
+    def _scorer(self, dataset, staged, project):
```
nv-alicheng, line 307: [Claude] **medium · testing**: The container tests assert `complete` in the sidecar `*_results.json` (`test_container_unreachable_leaves_lcb_unscored`, `test_failed_text_subset_marks_incomplete`) but never assert anything about the **returned headline `score`** — `score, n_repeats = ...score()` is captured and ignored. The actual data-integrity gap (already flagged at `scoring.py:1588`: a partial `combined` is returned and stored in `results.json` with no completeness marker) is therefore not guarded by any test. Two further branches are untested: the no-container path with LCB rows present (`scoring.py:1483` reached with `n_lcb>0` but `lcb_websocket_port=None`), and the LCB count-divergence warning (`scoring.py:1562`). `_scorer` always passes `lcb_websocket_port=13835`.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -314,6 +314,65 @@ def score(self) -> tuple[float, int]:
...
+        logger.warning(
+            "websocket-client not installed; cannot reach lcb-service. "
+            "Install with: pip install websocket-client"
+        )
+        return None
+    try:
+        ws = websocket.create_connection(
+            url, timeout=7200, ping_interval=30, ping_timeout=10
+        )
+    except (ConnectionRefusedError, OSError, Exception) as e:  # noqa: BLE001
```
nv-alicheng, line 342: [Claude] **low · error-handling**: `except (ConnectionRefusedError, OSError, Exception) as e:` — `Exception` already subsumes `ConnectionRefusedError` and `OSError`, so the tuple is misleading and, more importantly, this handler now treats *any* exception (including programming errors like `AttributeError`/`TypeError` raised inside `create_connection` usage) as a benign "connect failed" and returns `None`. Catch the specific network/protocol exceptions, or use a bare `except Exception` with the explanatory comment and let it be honest about what it swallows.

examples/10_DeepSeekR1_Example/prepare_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+    ap.add_argument(
+        "--subset",
+        type=int,
+        default=0,
+        help="If >0, also write a stratified subset of this many rows "
+        "(proportional per `dataset` subset) for a quick representative run.",
+    )
+    args = ap.parse_args()
+
+    df = pd.read_pickle(args.source)
```
nv-alicheng, line 75: [Claude] **low · security**: `df = pd.read_pickle(args.source)` deserializes an arbitrary path supplied via `--source` / `$DEEPSEEK_R1_DATASET_PKL`. `pickle` load executes arbitrary code, so pointing this at an untrusted `.pkl` is RCE. The official MLCommons dataset is the expected input and this is an operator-run script, so severity is low — but a one-line caveat in the script/docstring ("only load the official MLCommons `.pkl`; pickle deserialization executes arbitrary code") would be prudent since there is currently no such warning.

src/inference_endpoint/dataset_manager/transforms.py
```diff
@@ -176,12 +176,27 @@ def __init__(
...
+        # and the Harmonizer is never needed.
+        self._tokenizer_name = tokenizer_name
+        self._encoding_name = encoding_name
+        self._reasoning_effort = reasoning_effort
+        self._conversation_start_date = conversation_start_date
+        self._harmonizer: Harmonizer | None = None
+
+    @property
+    def harmonizer(self) -> Harmonizer:
+        if self._harmonizer is None:
```
nv-alicheng, line 192: [Claude] **low · design**: The `harmonizer` lazy property is not thread-safe: `if self._harmonizer is None: self._harmonizer = Harmonizer(...)` lets two threads both observe `None` and construct two `Harmonizer` instances. This is benign today because dataset transforms run single-threaded, but the lazy-init pattern silently bakes in that assumption (the previous eager construction made the attribute set-once). If a transform is ever shared across a thread pool (e.g. `df.apply` parallelism), this becomes a real double-construction race. Flagging so the single-threaded assumption is intentional, not accidental.

nv-alicheng:

Reviewed by: **Codex** (gpt-5.4) + **Claude** | Depth: **thorough**

Posted **6** inline findings. This PR already carries an automated review with several solid catches, so the council **deliberately did not re-post** issues that bot already covers — instead the synthesis below shows how everything fits together. 1 additional latent issue is out of diff scope and noted at the bottom.

The single most important issue is **already flagged** by the existing review and is not re-posted:

| File | Line | Category | Reviewer(s) | Summary |
|---|---|---|---|---|
| `evaluation/scoring.py` | 1588 | data-integrity | Both (already posted) | `score()` returns the partial `combined` even when `complete=False`; `finalize_benchmark` (`execute.py:976-985`) stores only the `score` float in `results.json` with **no completeness marker**, so a run that left LiveCodeBench unscored (container down) or had a failed text subset publishes a partial number as the headline 5-subset `exact_match`. The `Scorer.score()` contract has no channel for `complete`. Fix: return `None` when incomplete, **or** propagate `complete` into `accuracy_scores`. |

### 🟡 Should Fix (medium)

|---|---|---|---|---|---|
| 1 | `examples/.../offline_deepseek_r1_accuracy_smoke.yaml` | 12 | bug | Codex | Default model `deepseek_r1-torch-fp4-v2` is documented in this same PR as **not loadable** with stock TRT-LLM; smoke path fails out of the box without `SERVED_MODEL_NAME`. Default to `deepseek_r1-torch-fp4`. |
| 2 | `examples/10_DeepSeekR1_Example/README.md` | 186 | documentation | Claude | Self-contradiction: claims the scorer "falls back to grading LCB in-process" when the container is unreachable, but for the configured-port-but-down case the code leaves LCB **unscored** (`scoring.py:1527-1543`); README:201-202 itself says `complete: false`. |
| 3 | `tests/unit/evaluation/test_deepseek_r1_scorer.py` | 307 | testing | Claude | Container tests assert sidecar `complete` but never assert the returned headline `score` is gated when incomplete (the `scoring.py:1588` gap is untested); the no-container+LCB-rows branch (`scoring.py:1483`, port `None`) and the count-divergence warning (`scoring.py:1562`) are also untested. |

### 🔵 Consider (low)

|---|---|---|---|---|---|
| 4 | `evaluation/scoring.py` | 338 | error-handling | Claude | `except (ConnectionRefusedError, OSError, Exception)` — `Exception` subsumes the others and swallows programming errors as "connect failed". |
| 5 | `examples/.../prepare_dataset.py` | 75 | security | Claude | `pd.read_pickle(args.source)` on an env/CLI-supplied path is arbitrary-code-execution on load; add a "official MLCommons pkl only" caveat. |
| 6 | `dataset_manager/transforms.py` | 192 | design | Claude | Lazy `harmonizer` property has a double-construct race if ever shared across threads; benign today (single-threaded transforms) but the assumption is now implicit. |

### Already flagged by the existing bot review (not re-posted)
`scoring.py:1588` (partial score, above) · `scoring.py:1519` (float round-trip → wrong count) · `scoring.py:1518` (`NaN` em crashes `round()`) · `scoring.py:1433` (unvalidated WS response) · `deepseek_eval_runner.py:61` (`sys.path.insert` RCE) · `deepseek_eval_runner.py:109` (`trust_remote_code`) · `run_client.sh:43` (temp-file leak) · `scoring.py:1265`/`1511` (extractor / `lcb_subset` key).

### ⚠️ Out-of-diff advisory (not posted inline — unchanged line)
`evaluation/scoring.py:583` references module `inference_endpoint.dataset_manager.predefined.livecodebench.lcb_serve`, but `lcb_serve.py` actually lives at `inference_endpoint.evaluation.livecodebench.lcb_serve` (the `predefined/livecodebench/` package contains only `presets.py`/`__init__.py`). The subprocess fallback would fail with `ModuleNotFoundError`. Pre-existing, so out of this PR's scope, but worth a follow-up since the new LCB work documents a subprocess fallback story.

---
Commit hygiene: 6 commits, 0 fixups — fine. Build check skipped: only the example subproject `pyproject.toml` changed (isolated `uv` project), root package build is unaffected.

examples/10_DeepSeekR1_Example/score_livecodebench.sh
```diff
@@ -0,0 +1,77 @@
...
+PYTHONPATH="${REPO_ROOT}/src" "${PY}" -m inference_endpoint.evaluation.livecodebench.lcb_serve \
+  "${LCBIN}" --datasets-dir "${LCBDIR}" --version-tag "${LCB_VARIANT}" > "${RESULTS}.raw" 2>&1
+grep -E '^\s*\{.*\}\s*$' "${RESULTS}.raw" | tail -1 > "${RESULTS}"
+echo "=== ${RESULTS} ==="; cat "${RESULTS}"; echo
+if ! [ -s "${RESULTS}" ]; then
+  echo "ERROR: lcb_serve produced no JSON result; tail of ${RESULTS}.raw:" >&2
+  tail -n 20 "${RESULTS}.raw" >&2
+  exit 1
+fi
+echo "Fold into the aggregate: full_exact_match = (text_subset_correct + LCB passed_samples) / 4388."
```
nvzhihanj, line 77: [Claude] **medium** · bug — The fold-in hint hardcodes the full-set denominator:
```
full_exact_match = (text_subset_correct + LCB passed_samples) / 4388.
```
But `OUTPUTS_PARQUET` is user-overridable and the README documents ~385-sample subset runs via `offline_deepseek_r1_accuracy_subset.yaml`. A user who scores LCB for a subset run and follows this formula divides by **4388** instead of their actual sample count, silently understating accuracy by ~10×. Derive the denominator from the run's actual total sample count (e.g. `len(df)` of the outputs parquet) rather than hardcoding 4388.

examples/10_DeepSeekR1_Example/launch_and_run.sh
```diff
@@ -0,0 +1,192 @@
...
+# loader checks the *pre-fusion* q_a_proj/kv_a_proj names, so it allocates an
+# FP4-packed buffer (3584) for the BF16 weight (7168) and aborts. The sibling
+# `deepseek_r1-torch-fp4` excludes the individual names and loads cleanly, so it
+# is the default here. Point MODEL_DIR at the -v2 checkpoint only with a
+# trtllm build that handles it (the model card's reference is TRT-LLM main on 8xB200).
+MODEL_DIR="${MODEL_DIR:?Set MODEL_DIR to your DeepSeek-R1 FP4 (ModelOpt) checkpoint directory}"
+# trtllm-serve registers the model under the checkpoint dir's basename; the
+# /v1/completions `model` field (model_params.name in the YAML) must match it.
+SERVED_MODEL_NAME="${SERVED_MODEL_NAME:-$(basename "${MODEL_DIR}")}"
+SOURCE_PKL="${SOURCE_PKL:?Set SOURCE_PKL to the MLPerf DeepSeek-R1 source dataset .pkl}"
```
nvzhihanj, line 67: [Codex] **medium** · bug — `SOURCE_PKL="${SOURCE_PKL:?…}"` aborts the launch unconditionally when `SOURCE_PKL` is unset. But in the documented pre-prepared-parquet / `SERVER_ONLY=1` workflow the source pickle isn't needed — the later `if [ ! -f $PARQUET ]` check is meant to skip conversion. This `:?` expansion fires first, so a run with `data/deepseek_r1_eval.parquet` already present fails even though no conversion is required. Only require (and mount) `SOURCE_PKL` when the parquet is missing.

examples/10_DeepSeekR1_Example/accuracy/RUNBOOK.md
```diff
@@ -0,0 +1,96 @@
...
+is exercised end-to-end only when its rows are present and the LCB dataset is
+reachable.)
+
+## How the scorer calls this
+
+`DeepSeekR1Scorer.score()` writes
+`<report_dir>/deepseek_eval/<name>_outputs.parquet`, then runs:
+
+```bash
+uv run --project <this dir> python deepseek_eval_runner.py \
```
nvzhihanj, line 90: [Claude] **low** · documentation — This documented invocation omits `--external-subsets`:
```
uv run --project <this dir> python deepseek_eval_runner.py --input … --output … --tokenizer …
```
In the default (container-reachable) path `DeepSeekR1Scorer._run_eval_subprocess` passes `--external-subsets livecodebench` (scoring.py:1500), which changes the runner's behavior (LCB tokenized but not graded, `status=external`). The documented command matches only the no-container fallback — note both invocation modes so readers understand when `--external-subsets` applies.

nvzhihanj (commented):
## Review Council — Multi-AI Code Review (second pass)

Reviewed by: **Codex** (`--yolo`) + **Claude** · Depth: **thorough**

This PR already has a prior council pass + author self-review covering the high/medium issues. This pass posts only **net-new** findings not already commented; see the summary for the confirmed/deduped set.

nvzhihanj:
## Review Council — Summary (second pass)

Reviewed by: **Codex** (`--yolo`) + **Claude** · Depth: **thorough**

This PR already has a prior review-council pass + the author's self-review + `gemini`/`github-code-quality` bot comments. To avoid duplicate noise, this pass posts only **net-new** findings and records the dedup explicitly.

### 🟡 Net-new — posted inline

| # | File | Line | Severity | Category | Reviewer | Summary |
|---|------|------|----------|----------|----------|---------|
| A | `examples/10_DeepSeekR1_Example/score_livecodebench.sh` | 77 | medium | bug | Claude | Fold-in hint hardcodes `/ 4388`; README supports ~385-sample subset runs → wrong denominator, understates accuracy ~10×. |
| B | `examples/10_DeepSeekR1_Example/launch_and_run.sh` | 67 | medium | bug | Codex | `SOURCE_PKL:?` aborts unconditionally, so `SERVER_ONLY=1` / pre-prepared-parquet runs fail before the `[ ! -f $PARQUET ]` skip. |
| E | `examples/10_DeepSeekR1_Example/accuracy/RUNBOOK.md` | 90 | low | documentation | Claude | Documented `deepseek_eval_runner.py` invocation omits `--external-subsets livecodebench`, which the container path actually passes. |

### 🔵 Net-new — out-of-diff (can't anchor inline; noted here)

| # | File:Line | Severity | Category | Reviewer | Summary |
|---|-----------|----------|----------|----------|---------|
| C | `src/inference_endpoint/evaluation/scoring.py:206` | low | bug | Claude | Base `Scorer.score()` does `len(scores) // self.dataset.num_samples()` with no div-by-zero guard, while the PR's new scorers (`DeepSeekR1Scorer`, `VBenchScorer`) all guard with `if num_samples else 0`. Empty dataset → `ZeroDivisionError`. Pre-existing; the inconsistency is newly visible. |
| D | `src/inference_endpoint/config/schema.py:353` | low | api-contract | Claude | `AccuracyConfig.extras` is forwarded verbatim as `**eval_cfg.extras` into the scorer ctor (`execute.py:981`). A typo'd key (e.g. `lcb_websocket_prt`) surfaces as `TypeError: unexpected keyword argument` only *after* a multi-hour run, not at config-load time. (Field is pre-existing infra; this PR is its first heavy user — `subset_column`/`question_column`/`lcb_websocket_port`.) Consider an unknown-key check. |

### ✅ Confirmed but already commented (deduped — not re-posted)

The council independently reproduced these already-raised findings (cross-validation, not new):
- **Partial DeepSeek score reported as complete** — author `[HIGH]` `scoring.py:1588` (also names the `execute.py` `complete`-field fix); `gemini` `[high]` `:1518` (the `round(NaN)` crash variant). Flagged by **both** Codex (P1) and Claude here.
- **Round-trip count fragility** `scoring.py:1519` — author `[MEDIUM]`.
- **`lcb_subset` key mismatch / subset mis-classification** `scoring.py:1511` — author `[LOW]`.
- **Test gaps** (count-divergence + no-container-with-LCB branches) — prior council `[Claude]` `test_deepseek_r1_scorer.py:307`.
- **Smoke config `-v2` model default** `offline_deepseek_r1_accuracy_smoke.yaml:12` — prior council `[Codex]`.
- **Redundant `except (…, Exception)`** `scoring.py:338` — prior council `[Claude]`.

_Codex review ran via `codex --yolo` (now working in this environment)._

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -661,7 +661,11 @@ async def _run_benchmark_async(
                     args=event_logger_args,
                 ),
             ],
-            timeout=30.0,
+            # Service startup imports off a shared/Lustre FS can exceed 30 s
+            # under heavy login-node I/O contention; allow an override.
+            timeout=float(
+                os.environ.get("INFERENCE_ENDPOINT_SERVICE_READY_TIMEOUT_S", "30")
```
viraatc: [Review Council] **critical · bug** — `os` is not imported in this file.

```python
os.environ.get("INFERENCE_ENDPOINT_SERVICE_READY_TIMEOUT_S", "30")
```

The import block lists `asyncio`, `json`, `logging`, `random`, `shutil`, `signal`, `tempfile`, `uuid` — `os` is absent. This raises `NameError: name 'os' is not defined` on every benchmark run the moment `launcher.launch()` is called. Add `import os` to the imports.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1157,3 +1216,373 @@ def score(self) -> tuple[float | None, int]:
...
+            lcb_entry["status"] = "unscored"
+            lcb_ok = False
+        else:
+            lcb_passed, lcb_total = lcb_scored
+            lcb_entry["exact_match"] = (
+                100.0 * lcb_passed / lcb_total if lcb_total else None
+            )
+            lcb_entry["num_samples"] = lcb_total
+            lcb_entry["status"] = "lcb-service"
+            lcb_ok = lcb_total > 0
```
viraatc, line 2052: [Review Council] **high · bug** — `lcb_total = 0` is silently treated the same as container failure, with no diagnostic.

```python
lcb_ok = lcb_total > 0
```

When `_score_lcb_via_container` returns `(passed=0, total=0)` — e.g. all rows filtered away server-side, or the dataset column name doesn't match — `lcb_ok` becomes `False` with no log message. The result JSON gets `livecodebench.status = "lcb-service"` and `livecodebench.exact_match = None`, while `complete = False`. This is indistinguishable from the service being unreachable and gives the caller no way to know whether the container ran or not. A `logger.warning` when `lcb_total == 0` would distinguish the two cases.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -314,6 +314,65 @@ def score(self) -> tuple[float, int]:
...
 
 
+def _lcb_ws_evaluate(
+    url: str, codes_dict: dict[str, list[str]], timeout_sec: int
+) -> dict | None:
+    """Evaluate extracted code via the lcb-service WebSocket (synchronous).
+
+    Sends ``{codes_dict, timeout_sec}`` and consumes progress frames until a
+    terminal ``completed`` (returns ``result`` = ``{total_samples, results}``)
+    or ``error``. Returns None on any failure so callers can fall back. Kept as
```
viraatc, line 328: [Review Council] **medium · design** — `_lcb_ws_evaluate()` docstring claims it is shared by both scorers, but `LiveCodeBenchScorer` never calls it.

The docstring says:
> *kept as a module function so both LiveCodeBenchScorer and DeepSeekR1Scorer...share one client*

`LiveCodeBenchScorer.score()` still calls `self._evaluate_via_websocket()` (lines 446–549), a duplicate implementation that differs on `print()` vs `logger.warning`, explicit `WebSocketTimeoutException` handling, and inner `try/finally` structure. The two will diverge over time. `LiveCodeBenchScorer._evaluate_via_websocket` should be removed and its call site changed to use `_lcb_ws_evaluate`.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1157,3 +1216,373 @@ def score(self) -> tuple[float | None, int]:
...
+        input_parquet = scratch / f"{self.dataset_name}_outputs.parquet"
+        out_json = scratch / f"{self.dataset_name}_results.json"
+        eval_df.to_parquet(input_parquet, index=False)
+
+        n_lcb = int((eval_df["dataset"].astype(str) == self.lcb_subset).sum())
+        use_container = self.lcb_websocket_url is not None and n_lcb > 0
+
+        if not use_container:
+            # No container (or no LCB rows): grade every subset in-process.
+            self._run_eval_subprocess(input_parquet, out_json)
```
viraatc, line 1986: [Review Council] **medium · api-contract** — `score()` lets `RuntimeError` from `_run_eval_subprocess` propagate uncaught, violating the `Scorer.score()` contract.

```python
self._run_eval_subprocess(input_parquet, out_json)
```

`_run_eval_subprocess` raises `RuntimeError` on subprocess timeout and on non-zero exit. The `Scorer.score()` docstring explicitly states the method returns `(None, n_repeats)` when evaluation fails. An uncaught `RuntimeError` surfaces as an unexpected crash at the CLI level rather than a graceful `(None, n)` return. Same issue applies to the second call at line ~1500.

tests/unit/evaluation/test_deepseek_r1_scorer.py
```diff
@@ -0,0 +1,408 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
viraatc, line 1: [Review Council] **medium · testing** — Four `score()` paths have no test coverage:

1. **`lcb_scored = (0, 0)`** — container returns `(passed=0, total=0)`; `lcb_ok` becomes `False` silently.
2. **`_score_lcb_via_container` empty-results path** — non-None result with empty `per_problem` and non-zero `total_samples` (logger.error + `return None`).
3. **`total_n != expected_n` divergence warning** — count mismatch between runner's LCB count and issued rows.
4. **`combined is None` on container path** — all text subsets fail AND container unreachable simultaneously.

viraatc (commented):

_Claude-only review (Codex unavailable in this environment). See inline comments for findings._

viraatc:

Reviewed by: **Claude** (Codex unavailable — bwrap sandbox restriction) | Depth: **thorough** | PR: feat: deepseek r1 accuracy support

Found **7** issues — **6** posted inline, **1** summary-only (outside diff).

| # | File | Line | Category | Finding |
| 1 | `commands/benchmark/execute.py` | 667 | bug | `import os` missing — `NameError` on every benchmark run |
| 2 | `evaluation/scoring.py` | 1551 | bug | `lcb_total = 0` silently sets `lcb_ok = False` with no diagnostic — indistinguishable from service failure |

### 🟡 Should Fix (medium)

| # | File | Line | Category | Finding |
| 3 | `commands/benchmark/execute.py` | 666 | error-handling | No validation before `float()` on env var — non-numeric value gives cryptic `ValueError` |
| 4 | `evaluation/scoring.py` | 324 | design | `_lcb_ws_evaluate()` docstring claims it is shared by both scorers; `LiveCodeBenchScorer` was never refactored to use it — two diverging WebSocket client implementations |
| 5 | `evaluation/scoring.py` | 1485 | api-contract | `RuntimeError` from `_run_eval_subprocess` escapes `score()` uncaught — violates `Scorer.score()` contract of returning `(None, n)` on failure |
| 6 | `evaluation/scoring.py` | 65 *(outside diff)* | design | `REQUIRES_EXTRACTOR` declared on base class but never enforced in `Scorer.__init__` — silent `None` extractor defers crash to `DataFrame.apply()` inside `score()` |
| 7 | `tests/unit/evaluation/test_deepseek_r1_scorer.py` | 1 | testing | Four untested `score()` paths: `(0,0)` from container, empty `per_problem`, count-divergence warning, and `combined=None` when all text subsets fail + container unreachable |

> ℹ️ Findings already covered by prior review comments (skipped): `scoring.py:1519` (round-trip float), `scoring.py:338` (exception handler), `scoring.py:1519` (exact match reconstruction).

src/inference_endpoint/commands/benchmark/execute.py
viraatc: should we document this somehwere?
can we add/use a config default  instead for consistentcy?

arekay-nv: Tracking in [issue](https://github.com/mlcommons/endpoints/issues/341)

examples/10_DeepSeekR1_Example/offline_deepseek_r1_accuracy_smoke.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+datasets:
+  - name: deepseek_r1_perf
+    type: "performance"
+    path: examples/10_DeepSeekR1_Example/data/deepseek_r1_eval.parquet
+    samples: 24
+  - name: deepseek_r1_accuracy
+    type: "accuracy"
+    path: examples/10_DeepSeekR1_Example/data/deepseek_r1_eval.parquet
+    samples: 24
```
arekay-nv, line 26: [Review Council] **high · bug** — `samples: 24` on the accuracy dataset is a silent no-op, so this "smoke" config runs the **full** DeepSeek set, not 24 rows.

`DataLoaderFactory.create_loader()` (`dataset_manager/factory.py`) never reads `config.samples`, and `_build_phases` issues `acc_ds.num_samples() * acc_ds.repeats` (`commands/benchmark/execute.py:480`) for each accuracy dataset. Since `path` points at the full `deepseek_r1_eval.parquet`, `num_samples()` is the full 4388 — the accuracy phase issues and scores all of them. The perf phase only stays bounded because the YAML also sets `n_samples_to_issue: 24` under `settings.runtime`; the accuracy phase has no such bound. A user running this as the documented "fast smoke" step kicks off the full multi-hour run instead.

Fix: point the smoke config at a small pre-generated parquet (~24 rows), or implement honoring `DatasetConfig.samples` in the loader (note `samples` is advertised in the CLI `--dataset ...,samples=N` help but is currently consumed nowhere in `src/`).

examples/10_DeepSeekR1_Example/score_livecodebench.sh
```diff
@@ -0,0 +1,77 @@
...
+echo "=== generate LCB ${LCB_VARIANT} test cases ==="
+"${PY}" src/inference_endpoint/evaluation/livecodebench/generate.py \
+  --datasets-dir "${LCBDIR}" --variant "${LCB_VARIANT}"
+
+echo "=== extract code + question_id from saved outputs ==="
+"${PY}" - "${OUTPUTS_PARQUET}" "${LCBIN}" <<'PY'
+import pandas as pd, re, sys
+df = pd.read_parquet(sys.argv[1])
+lcb = df[df["dataset"] == "livecodebench"].copy()
+extract = lambda t: (re.findall(r"```python(.*?)```", str(t), re.DOTALL) or [""])[-1].strip()
```
arekay-nv, line 57: [Review Council] **medium · data-integrity** — this extractor only matches ```` ```python ```` fences and silently drops valid plain-fenced code.

`re.findall(r"```python(.*?)```", ...)` writes an empty `extracted_code` for any answer using a bare ```` ``` ```` block. The in-run `PythonCodeExtractor` (`evaluation/extractor.py:254`) deliberately accepts **both** ```` ```python ```` and plain ```` ``` ```` blocks (priority 2: "last plain ``` block"). Since this script is the fallback that produces the **official** LCB number when the in-run container was unreachable, plain-fence answers are graded as failures here — underreporting accuracy versus the in-run path.

Fix: reuse `PythonCodeExtractor` (or add the same plain-fence fallback) so the rescore matches in-run extraction.

examples/10_DeepSeekR1_Example/launch_and_run.sh
```diff
@@ -0,0 +1,192 @@
...
+
+cd "${REPO_ROOT}"
+
+# ---------------------------------------------------------------------------
+# 0. Prepare the benchmark dataset (pkl -> parquet) if missing.
+# ---------------------------------------------------------------------------
+PARQUET="${SCRIPT_DIR}/data/deepseek_r1_eval.parquet"
+if [ ! -f "${PARQUET}" ]; then
+    echo "==> Preparing dataset parquet"
+    python "${SCRIPT_DIR}/prepare_dataset.py" --source "${SOURCE_PKL}" --output "${PARQUET}"
```
arekay-nv, line 100: [Review Council] **medium · bug** — bare `python` runs the system interpreter, which usually lacks `pandas`/`pyarrow`.

`python "${SCRIPT_DIR}/prepare_dataset.py" ...` runs on the batch/login node (before the trtllm container srun), where the documented flow only does `uv sync` for the accuracy subproject and never activates `.venv`. On most clusters system `python` has neither `pandas` nor `pyarrow`, so the end-to-end script fails at dataset prep — before the server starts — whenever the parquet is missing (i.e. the first run). Use `uv run python` (or the accuracy subproject's interpreter) like the other client commands.

arekay-nv (commented):

Reviewed by **Codex** + **Claude** | Depth: **thorough**

Posting **3** new findings. The PR already carries 26 inline comments (incl. two prior Review Council runs), so ~10 of our findings were dropped as duplicates — see the summary comment for the full accounting.

arekay-nv:

Reviewed by **Codex** + **Claude** | Depth: **thorough** (auto-selected, diff ≈4.9k lines)

This PR is already heavily reviewed (**26** existing inline comments across 5+ reviewers, including two prior Review Council runs). After deduplicating against those, **3 new findings** were posted inline. The rest of what we found was already covered — listed below for completeness.

### 🔴 Must Fix
| 1 | `examples/10_DeepSeekR1_Example/offline_deepseek_r1_accuracy_smoke.yaml` | 26 | bug | Codex (verified) | `samples: 24` is a silent no-op for the accuracy phase — the "smoke" config runs the **full 4388-sample** set. `create_loader()` never reads `DatasetConfig.samples`; `_build_phases` issues `acc_ds.num_samples()*repeats` (`execute.py:480`). Perf phase is only bounded because the YAML also sets `n_samples_to_issue: 24`. |

### 🟡 Should Fix
| 2 | `examples/10_DeepSeekR1_Example/score_livecodebench.sh` | 57 | data-integrity | Codex (verified) | Rescore extractor only matches ` ```python ` fences; the in-run `PythonCodeExtractor` (`extractor.py:254`) also accepts plain ` ``` ` fences. As the fallback that produces the **official** LCB number, this underreports accuracy for plain-fenced answers. |
| 3 | `examples/10_DeepSeekR1_Example/launch_and_run.sh` | 100 | bug | Codex (verified) | Bare `python prepare_dataset.py` uses the system interpreter (no `pandas`/`pyarrow` in the documented `uv sync`-only flow) → end-to-end script fails at dataset prep on first run. Use `uv run python`. |

### Already covered by existing comments (not re-posted)
- **scoring.py:1588 / 1554** — partial LCB score returned without a caller-visible `complete` flag (the flag only reaches `deepseek_eval/*_results.json`, not the `score` in `results.json`). *Confirmed independently; matches arekay-nv `[HIGH]`.*
- **scoring.py:1485 / execute.py:892** — `score()` shells out and can raise (`RuntimeError`/`FileNotFoundError`/`DecodeError`); the `for eval_cfg` loop is unguarded, so **one dataset's failure aborts scoring of all remaining datasets**. *The loop-blast-radius is the new angle on viraatc's `[Review Council]` propagation comment.*
- **scoring.py:1519** — `round(em/100*n)` integer round-trip depends on the runner emitting a full-precision mean (`deepseek_eval_runner.py:138`); brittle if the runner ever rounds. *Matches arekay-nv `[MEDIUM]`.*
- **README.md:186** — claims an in-process LCB fallback that the code deliberately does NOT do (`scoring.py:1527-1543`: "Do NOT re-run the in-process executor"). *Matches nv-alicheng `[Claude]`.*
- **scoring.py:338** — `except (ConnectionRefusedError, OSError, Exception)` ≡ `except Exception`. *Matches nv-alicheng `[Claude]`.*
- **runner:109** — `trust_remote_code=True` unnecessary for token-length counting. *Matches gemini-code-assist.*
- **transforms.py:191-192** — lazy `harmonizer` property (untested + not thread-safe). *Matches nv-alicheng `[Claude]`.*
- **Testing gaps** — LCB count-divergence branch (`scoring.py:1562`) and `_service_ready_timeout()` env-parse branch (`execute.py:105`) are untested. *Subsumed by viraatc's "Four `score()` paths have no coverage".*

**Dropped as low-value noise** (narrow/cosmetic on an already-exhaustively-reviewed PR): WebSocket leak in the `tqdm`/`sum` window before the `try` (`scoring.py:341`); `run_client.sh:63` relative `REPORT_DIR` + `2>/dev/null || true` swallowing a missing-`results.json` error.

> Note: most of the actionable findings live in the new `examples/10_DeepSeekR1_Example/` artifacts. The core `src/` accuracy plumbing (`DeepSeekR1Scorer`, the `openai_completions` pre-tokenized path, the `complete`-flag bookkeeping) is sound; the recurring theme is that the **LCB-unreachable fallback path is under-surfaced and the example scripts/docs drift from it**.

examples/10_DeepSeekR1_Example/offline_deepseek_r1_accuracy.yaml
```diff
@@ -0,0 +1,87 @@
...
+type: "offline"
+
+model_params:
+  # MUST match trtllm-serve's registered model id, which is the checkpoint dir's
+  # basename. With launch_and_run.sh defaulting to the loadable
+  # `deepseek_r1-torch-fp4` checkpoint, that id is `deepseek_r1-torch-fp4`.
+  name: "deepseek_r1-torch-fp4"
+  temperature: 0.0
+  top_p: 1.0
+  max_new_tokens: 20480 # MLPerf DeepSeek-R1 max output sequence length
```
nvzhihanj, line 31: Accuracy should be 32k?

nvzhihanj: NVM, this is DS-R1

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -314,6 +314,65 @@ def score(self) -> tuple[float, int]:
         return result, 1
 
 
+def _lcb_ws_evaluate(
```
viraatc, line 321: nit: any reason for free-hanging? can we put this in LCB Scorer classmethod?

viraatc (approved):
lgtm, thanks!

examples/10_DeepSeekR1_Example/trtllm_serve_config_tp8.yaml
```diff
@@ -0,0 +1,39 @@
+# trtllm-serve --extra_llm_api_options for DeepSeek-R1 FP4 on 8 GPUs
```
nvzhihanj, line 1: @arekay-nv please remove server config from the repo and clean up the git commit for this

examples/10_DeepSeekR1_Example/trtllm_serve_config.yaml
```diff
@@ -0,0 +1,36 @@
+# trtllm-serve --extra_llm_api_options for DeepSeek-R1 FP4 on a single GB200x4
```
nvzhihanj, line 1: This too

examples/10_DeepSeekR1_Example/score_livecodebench.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
```
nvzhihanj, line 1: I think we need to properly clean this up and move the accuracy artifacts into the evaluation repo, instead of the example folder

examples/10_DeepSeekR1_Example/README.md
```diff
@@ -0,0 +1,260 @@
...
+## Accuracy config (the single YAML)
+
+One config drives everything: a tiny `performance` dataset for the perf phase and
+the full 4388-sample `accuracy` dataset scored by the `deepseek_r1` scorer. The
+accuracy entry:
+
+```yaml
+- name: deepseek_r1_accuracy
+  type: "accuracy"
+  path: examples/10_DeepSeekR1_Example/data/deepseek_r1_eval.parquet
```
nvzhihanj, line 168: We can push this parquet file to the repo, it's only ~2MB in size

examples/10_DeepSeekR1_Example/README.md
```diff
@@ -0,0 +1,260 @@
...
+9. **LiveCodeBench runs in a sandbox** (see "LiveCodeBench scoring"). It executes
+   untrusted model code, so the `deepseek_r1` scorer grades it via the `lcb_serve`
+   container over a WebSocket (in-run, port 13835) instead of in-process. If no
+   container is available, score it afterward with `score_livecodebench.sh` - the
+   MLCommons in-process executor can't kill runaway code (hangs, `N of M futures
+unfinished`) and needs a ~21 GB dataset load that OOMs the login cgroup.
+
+## Prerequisites
+
+- Single GB200x4 node via SLURM (`-A <your-account> -p <your-partition>`).
```
nvzhihanj, line 73: Internal data?

examples/10_DeepSeekR1_Example/offline_deepseek_r1_accuracy_smoke.yaml
```diff
@@ -0,0 +1,56 @@
+# SMOKE variant of offline_deepseek_r1_accuracy.yaml - a small, fast subset to
```
nvzhihanj, line 1: need clean up

examples/10_DeepSeekR1_Example/offline_deepseek_r1_accuracy_subset.yaml
```diff
@@ -0,0 +1,59 @@
+# Representative-subset variant of offline_deepseek_r1_accuracy.yaml - a
```
nvzhihanj, line 1: need cleanup

examples/10_DeepSeekR1_Example/accuracy/pyproject.toml
```diff
@@ -0,0 +1,55 @@
+# Isolated uv project for the MLPerf DeepSeek-R1 accuracy evaluator.
```
nvzhihanj, line 1: This is a separate eval docker, good to refactor into the eval folder. example folder should just be the recipe

nvzhihanj (changes requested):

arekay-nv:
closed in favor of #378

PR #344, opened 2026-06-06
feat: Enable per dataset max-osl

arekay-nv:
## What does this PR do?
Enables the max-osl to be specified per-dataset. This enables running perf/accuracy runs with different OSL in the same benchmark.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass


src/inference_endpoint/openai/completions_adapter.py
```diff
@@ -62,6 +62,7 @@ def dataset_transforms(cls, model_params: ModelParams) -> list[Transform]:
             "repetition_penalty": model_params.repetition_penalty,
             "presence_penalty": model_params.presence_penalty,
             "frequency_penalty": model_params.frequency_penalty,
+            "skip_special_tokens": False,
```
arekay-nv: **[Claude] medium · bug**

`"skip_special_tokens": False` is added to the `metadata` dict (and therefore injected into every `query.data` via `AddStaticColumns`), but `TextCompletionRequest` has no `skip_special_tokens` field and `encode_query()` never reads `query.data.get("skip_special_tokens")`. The value is silently dropped on every request.

If the intent is to pass this parameter to the server (vLLM's `/v1/completions` does accept `skip_special_tokens`), both `TextCompletionRequest` and `encode_query()` need to be updated. If it's only a client-side concern (Harmony already handles special tokens), this entry in `metadata` is dead code and should be removed to avoid confusion.

arekay-nv (commented):
## Review Council — Claude-only review (Codex failed: cloud policy auth error)

src/inference_endpoint/openai/completions_adapter.py
arekay-nv: **[Claude] medium · bug**

`"skip_special_tokens": False` is added to the `metadata` dict (and injected into every `query.data` via `AddStaticColumns`), but `TextCompletionRequest` has no `skip_special_tokens` field and `encode_query()` never reads it. The value is silently dropped on every request.

If the intent is to pass this to the server (vLLM's `/v1/completions` accepts `skip_special_tokens`), both `TextCompletionRequest` and `encode_query()` need updating. If it's only a client-side concern, this entry is dead code and should be removed.

arekay-nv:
## Review Council — Claude-only Code Review

Reviewed by: **Claude** (Codex failed — cloud workspace policy auth error) | Depth: **standard** (240 lines)

Found **2 issues** across **2 files**:
- 1 medium
- 1 low

| # | File | Line | Severity | Category | Summary |
| 1 | `src/inference_endpoint/openai/completions_adapter.py` | 65 | medium | bug | `skip_special_tokens: False` in metadata dict never wired into `TextCompletionRequest` — silently dropped |
| 2 | `src/inference_endpoint/commands/benchmark/execute.py` | 320 | low | bug | FileNotFoundError handler uses `performance_cfgs[0].path` instead of the `perf_cfg` variable extracted on line 305 (outside diff hunk — not posted inline) |

---

**Issue 1** (inline): `completions_adapter.py:65` — `"skip_special_tokens": False` is added to `metadata` and flows into `query.data`, but `TextCompletionRequest` has no such field and `encode_query()` never reads it. Either wire it through to the request or remove the entry.

**Issue 2** (summary-only, line outside diff): `execute.py:320` — the `FileNotFoundError` handler body still uses `performance_cfgs[0].path` instead of `perf_cfg.path`. This PR introduced `perf_cfg = performance_cfgs[0]` at line 305 to avoid repeated indexing, but the error message at line 320 was not updated. Minor consistency nit.

src/inference_endpoint/config/templates/concurrency_template_full.yaml
```diff
@@ -26,6 +26,7 @@ datasets:  # Dataset configs
     prompt: text_input
   accuracy_config: null  # Accuracy evaluation settings
   multi_turn: null  # Multi-turn conversation configuration
+  model_params_override: null  # Per-dataset overrides for the top-level model_params (sparse — only the fields you want to override). Merged on top of BenchmarkConfig.model_params at dataset-load time. Useful for MLPerf-style runs where accuracy and performance use different output budgets in the same fleet, e.g. model_params_override: {max_new_tokens: 32768, streaming: 'on'}.
```
nvzhihanj: The name doesn't look right. Convention is we call the parameters of the model as model parameters, e.g. the one from: https://huggingface.co/moonshotai/Kimi-K2.6/blob/main/config.json. These are invariants unless we are debugging the model.

And the max OSL and streaming are generation config: https://huggingface.co/moonshotai/Kimi-K2.6/blob/main/generation_config.json (and also https://github.com/mlcommons/inference_results_v6.0/blob/main/closed/NVIDIA/code/deepseek-r1/tensorrt/generation_config.json which has more top_p, top_k). These are controlled by the client through parameters in the query

Right now it seems like we are hard-coding the generation config in the dataset (through transform) - ideally we should create generation config for each "benchmark" (because it's a per-run settings, not per-dataset). I would recommend that you add this to post-mortem and fix the hierarchy properly later.

If you want to WAR short-term, at least name it generation_configs_override and put TODO to fix them later

src/inference_endpoint/config/schema.py
```diff
@@ -314,6 +325,34 @@ def _auto_derive_name(self) -> Self:
...
+            bad = sorted(set(self.model_params_override) - valid)
+            if bad:
+                raise ValueError(
+                    f"Dataset '{self.name}': unknown keys in "
+                    f"model_params_override: {bad}. "
+                    f"Valid keys: {sorted(valid)}"
+                )
+        return self
+
+    def effective_model_params(self, base: ModelParams) -> ModelParams:
```
nvzhihanj: Similar here, calling the function model_param will be confusing (it's not really model params)

nvzhihanj (changes requested):

src/inference_endpoint/config/schema.py
```diff
@@ -306,6 +306,17 @@ class Dataset(BaseModel):
...
     )
+    model_params_override: dict[str, Any] | None = Field(
+        None,
+        description=(
+            "Per-dataset overrides for the top-level model_params (sparse — "
+            "only the fields you want to override). Merged on top of "
+            "BenchmarkConfig.model_params at dataset-load time. Useful for "
+            "MLPerf-style runs where accuracy and performance use different "
+            "output budgets in the same fleet, e.g. "
+            "model_params_override: {max_new_tokens: 32768, streaming: 'on'}."
```
nvzhihanj: [Claude] **high** · data-integrity — the documented example advertises a key that silently half-works.

The example here is `model_params_override: {max_new_tokens: 32768, streaming: 'on'}`, but a per-dataset `streaming` override does **not** take full effect. `effective_model_params` merges it into the dataset's params and the adapter emits `stream: true` into that dataset's request rows — yet `ctx.enable_streaming` (`execute.py:185`) reads the **global** `config.model_params.streaming`, and that single value gates the one metrics-aggregator's `--streaming` flag for the whole run (`execute.py:638`). So if the global resolves OFF (offline default) while an accuracy dataset overrides `streaming: 'on'`, the server streams that phase but the aggregator was launched without `--streaming`, so TTFT/TPOT triggers are never registered — a silent half-applied override. Per-dataset streaming can't work under the single-aggregator architecture; it's per-run. **WAR fix:** drop `streaming` from the example and reject it as an override key (see the surface-narrowing comment); restrict the override to keys that actually take effect per-dataset (`max_new_tokens` + sampling params).

nvzhihanj (commented):
## Review Council — first-principles WAR design review

Reviewed by: **Claude** (Codex review timed out — `xhigh` reasoning latency exceeds the 600s cap even on this small diff) · Depth: **thorough**

Scope: is the per-dataset max-OSL **WAR** correctly designed for a clean later refactor, given the long-term goal of per-phase generation config? 8 net-new findings (none duplicate the existing maintainer / gemini comments); verdict + long-term plan in the summary comment. The HIGH streaming data-integrity bug was verified against the source.

src/inference_endpoint/config/schema.py
```diff
@@ -306,6 +306,17 @@ class Dataset(BaseModel):
     multi_turn: MultiTurnConfig | None = Field(
         None, description="Multi-turn conversation configuration"
     )
+    model_params_override: dict[str, Any] | None = Field(
```
nvzhihanj: [Claude] **high** · design — the override is bound to *dataset identity*, which is a dead-end for the stated long-term goal ("different generation config for the **same** dataset across phases").

Phases are derived *from* the dataset list (`_build_phases`: warmup / performance / accuracy), and duplicate `(name, type)` dataset entries are rejected (`schema.py:800-807`). So to run the same dataset file at two OSLs you must duplicate the entry under different names — and even then it can only ever be perf-vs-accuracy, never two perf phases. The override can vary generation config only along the "which dataset entry" axis, which is the wrong axis. (Builds on the maintainer's "this is per-run, not per-dataset" comment.)

**Why re-design:** generation config is per-run/per-phase; keying it to the dataset row blocks the long-term goal. **Suggested shape — WAR:** keep it on `Dataset` for now but rename to `generation_config_override` and add `# TODO(post-mortem): generation config is per-phase, not per-dataset`. **Long-term:** a first-class `GenerationConfig` carried on `PhaseConfig` (a phase references a dataset *and* a generation config), decoupled from the dataset entry; the adapter's `dataset_transforms` reads the phase's `GenerationConfig`.

src/inference_endpoint/config/schema.py
```diff
@@ -314,6 +325,34 @@ def _auto_derive_name(self) -> Self:
             object.__setattr__(self, "name", Path(self.path).stem)
         return self
 
+    @model_validator(mode="after")
+    def _validate_model_params_override(self) -> Self:
```
nvzhihanj: [Claude] **high** · design — the override surface is too wide: it accepts **every** `ModelParams` field, including ones that can't meaningfully vary per dataset and silently misbehave.

`valid = set(ModelParams.model_fields)` (line 334) accepts any key, so `model_params_override: {name: ...}` is permitted. But `name` is **model identity**: the tokenizer probe (`execute.py:403-415`) and the aggregator `--tokenizer` arg (`execute.py:641`) use the **global** `config.model_params.name`, so a per-dataset `name` override flows into the request `model` field while ISL/OSL are tokenized with the wrong tokenizer. Together with `streaming` (aggregator-ignored, separate comment) and `osl_distribution` (no consumer, separate comment), the override advertises three keys that don't actually work per-dataset. **Why re-design:** identity and per-run knobs share one `ModelParams`, so the override surface can't distinguish them. **Suggested shape — WAR:** restrict the accepted keys to a per-dataset-meaningful allowlist (`max_new_tokens`, `temperature`, `top_p`, `top_k`, penalties); reject `name`/`streaming`/`osl_distribution`. **Long-term:** split `ModelParams` into `ModelIdentity` (per-run) + `GenerationConfig` (per-phase) so only generation fields are overridable.

src/inference_endpoint/config/schema.py
```diff
@@ -314,6 +325,34 @@ def _auto_derive_name(self) -> Self:
...
+    def effective_model_params(self, base: ModelParams) -> ModelParams:
+        """Return base merged with this dataset's overrides.
+
+        Re-validates the merged dict through ``ModelParams.model_validate``
+        so that cross-field constraints (e.g. value ranges) catch bad
+        overrides before downstream code touches them.
+        """
+        if not self.model_params_override:
+            return base
+        merged = {**base.model_dump(), **self.model_params_override}
```
nvzhihanj: [Claude] **medium** · bug — *(extends gemini's shallow-merge note at :354 with the concrete worst case.)*

`merged = {**base.model_dump(), **self.model_params_override}` is a shallow merge. The only nested field in `ModelParams` is `osl_distribution`, and a sparse nested override **silently resets** its siblings rather than erroring: `model_params_override: {osl_distribution: {max: 512}}` discards the base `type`/`mean`/`std`/`min`, and because every `OSLDistribution` field has a default, `ModelParams.model_validate(merged)` does **not** raise — so the docstring's claim (lines 347-349) that re-validation "catches bad overrides" is false for the one case where it matters. Net-new beyond gemini: `osl_distribution` has **no consumer** anywhere in `src/` (it is read only inside `schema.py`), so overriding it has zero runtime effect today regardless. **Fix:** drop `osl_distribution` from the accepted override keys (per the surface-narrowing comment); if nested overrides are ever needed, deep-merge as gemini suggested.

tests/unit/commands/test_benchmark.py
```diff
@@ -1137,3 +1139,120 @@ def __str__(self):
 
         panel = _error_formatter(FakeError())
         assert "something went wrong" in panel.renderable
+
+
+class TestLoadDatasetsModelParamsOverride:
```
nvzhihanj: [Claude] **medium** · testing — the added tests cover happy paths but miss the branches that expose this design's sharp edges.

Not covered: (1) **nested/deep-merge** of `osl_distribution` (the silent-reset case) — no test asserts what `{osl_distribution: {max: 512}}` produces; (2) **streaming override end-to-end** — no test asserts that `{streaming: 'on'}` (the field's own documented example) changes `ctx.enable_streaming` / the aggregator `--streaming` arg (it doesn't — a test here would pin the limitation); (3) the **multi-turn** loader path (`MultiTurnDataset.load` takes `model_params` and has its own `max_new_tokens` mapping) — unexercised with an override; (4) the tests assert `max_completion_tokens` (the **chat** adapter key), but the PR's headline target is the `openai_completions` adapter, which emits `max_tokens` — the very path is untested.

src/inference_endpoint/config/templates/offline_template_full.yaml
```diff
@@ -26,6 +26,7 @@ datasets:  # Dataset configs
     prompt: text_input
   accuracy_config: null  # Accuracy evaluation settings
   multi_turn: null  # Multi-turn conversation configuration
+  model_params_override: null  # Per-dataset overrides for the top-level model_params (sparse — only the fields you want to override). Merged on top of BenchmarkConfig.model_params at dataset-load time. Useful for MLPerf-style runs where accuracy and performance use different output budgets in the same fleet, e.g. model_params_override: {max_new_tokens: 32768, streaming: 'on'}.
```
nvzhihanj: [Claude] **medium** · design — the WAR is being baked into all three full templates (`offline`/`online`/`concurrency` `_template_full.yaml`, two stanzas each), regenerated by the `regenerate-templates` pre-commit hook and parsed by the template-validation test. That makes a short-term WAR a documented, load-bearing config surface across six stanzas + the hook output, so renaming/removing it later is churny — exactly the "hard to undo" risk. The comment also advertises `streaming: 'on'`, which doesn't work per-dataset. **Suggested shape — WAR:** if it must ship in templates, rename to `generation_config_override` **now** (so the eventual rename isn't breaking template churn) and trim the example to keys that take effect (`max_new_tokens` + sampling), dropping `streaming`/`osl_distribution`. **Long-term:** when per-phase `GenerationConfig` lands, the per-dataset key is deleted and templates regenerate cleanly.

src/inference_endpoint/config/schema.py
nvzhihanj: [Claude] **medium** · error-handling — value validation of the override is deferred to dataset-load time, after setup has already produced side effects.

`_validate_model_params_override` (a `Dataset` model-validator) checks only the override **keys** at construction; **values** are validated here in `effective_model_params` via `ModelParams.model_validate`, which runs inside `_load_datasets` (`execute.py:292/307`) — after `setup_benchmark` has created the report dir, written `config.yaml`, and probed the tokenizer. So a value-invalid override (e.g. `temperature: "hot"`) aborts mid-setup with artifacts already on disk, unlike every other field error which fails at config construction. The base `model_params` is available on `BenchmarkConfig`, so value validation could move into `BenchmarkConfig._resolve_and_validate` to fail atomically at parse time. Acceptable for a WAR, but worth a TODO; the docstring's "before downstream code touches them" is true only relative to the adapter, not to setup side-effects.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -288,20 +288,31 @@ def _load_datasets(
                 acc_cfg.accuracy_config.extras or {},
             )
         )
-        ds.load(
-            api_type=config.endpoint_config.api_type, model_params=config.model_params
-        )
+        try:
+            ds_model_params = acc_cfg.effective_model_params(config.model_params)
+        except Exception as e:
```
nvzhihanj: [Claude] **low** · error-handling — `except Exception` here (and at the perf branch ~306-311) is wider than needed. `effective_model_params` does only a dict merge + `ModelParams.model_validate`, whose sole expected failure is `ValidationError` (a `ValueError`). Catching bare `Exception` would also relabel a future programming error (e.g. an `AttributeError` from a refactor) as "invalid model_params_override," misleading the user. Narrow to `(ValidationError, ValueError)` — consistent with the specific `except FileNotFoundError` used a few lines down.

nvzhihanj:
## Review Council — first-principles WAR design review

Reviewed by: **Claude** · Depth: **thorough**
_Codex review timed out (its `xhigh` reasoning latency exceeds the 600s cap even on this 381-line diff); Claude-led. The HIGH streaming bug was verified against the source._

This is a focused assessment of whether the per-dataset max-OSL **WAR** is correctly designed for a clean later refactor, given the long-term goal of per-phase generation config. **8 net-new findings** posted inline — none duplicate the existing maintainer / gemini comments.

### Verdict on the WAR

It achieves the immediate need (accuracy vs performance differ, since they're separate dataset entries) but is **not yet "correct" as a WAR** — three structural issues:

1. **Wrong axis.** It's keyed to *dataset identity*, which cannot express the stated long-term goal ("different generation config for the **same** dataset across phases") — phases derive from datasets and duplicate `(name,type)` entries are rejected, so you'd have to duplicate the dataset entry.
2. **Over-broad surface.** It accepts *all* `ModelParams` keys, including three that silently misbehave per-dataset: `name` (→ ISL/OSL tokenized with the wrong tokenizer), `streaming` (→ aggregator ignores it; TTFT/TPOT unrecorded), `osl_distribution` (no consumer + shallow-merge degrades). **The field's own documented example advertises `streaming: 'on'`, which doesn't work.**
3. **Naming** (the maintainer's point): `model_params_override` mislabels generation config as model parameters.

**Minimal clean WAR:** rename → `generation_config_override`; **restrict accepted keys** to the per-dataset-meaningful set (`max_new_tokens` + sampling params), rejecting `name`/`streaming`/`osl_distribution`; add `# TODO(post-mortem): generation config is per-phase, not per-dataset`; trim the template/docstring example accordingly.

**Long-term:** split `ModelParams` → `ModelIdentity {name, tokenizer}` (per-run invariant) + `GenerationConfig {max_new_tokens, temperature, top_p, top_k, penalties, streaming, osl_distribution}` (per-phase). Carry `GenerationConfig` on `PhaseConfig`; adapters' `dataset_transforms` read the phase's `GenerationConfig`, and the dataset becomes generation-agnostic. "Same dataset, different gen-config per phase" then falls out naturally (per-phase streaming additionally needs per-phase aggregator handling).

### 🔴 Must-fix (for the WAR to be correct)

| # | File | Line | Cat | Summary |
|---|------|------|-----|---------|
| A | `config/schema.py` | 317 | data-integrity | Per-dataset `streaming` override reaches request rows but the single aggregator reads the **global** `streaming` (execute.py:185) → TTFT/TPOT silently unrecorded. The documented example advertises this broken key. |
| B | `config/schema.py` | 309 | design | Override bound to dataset identity is a dead-end for "same dataset, different gen-config per phase." WAR: rename + TODO. Long-term: per-phase `GenerationConfig`. |
| C | `config/schema.py` | 329 | design | Accepts all `ModelParams` keys incl. `name` → per-dataset `name` mismatches the global tokenizer/aggregator. Restrict to a per-dataset allowlist. |

### 🟡 Should-fix

| # | File | Line | Cat | Summary |
|---|------|------|-----|---------|
| D | `config/schema.py` | 353 | bug | *(extends gemini)* Shallow-merge of `osl_distribution` silently resets siblings **and passes validation** (all defaults), so the docstring's "catches bad overrides" is false; also `osl_distribution` has no consumer → overriding it is a no-op. |
| E | `tests/unit/commands/test_benchmark.py` | 1144 | testing | Gaps: nested/deep-merge, streaming end-to-end, multi-turn loader, and the **completions adapter `max_tokens`** key (tests only assert the chat `max_completion_tokens`, but the PR targets the completions path). |
| F | `config/templates/offline_template_full.yaml` (×3) | 29 | design | WAR baked into all three generated templates + the template-parse test → load-bearing / hard to undo; also advertises broken `streaming:'on'`. Rename now; trim the example. |
| G | `config/schema.py` | 344 | error-handling | Override **value** validation deferred to load-time (after setup writes report dir / config.yaml / probes tokenizer); move into `BenchmarkConfig._resolve_and_validate` to fail atomically at construction. |

### 🔵 Consider

| # | File | Line | Cat | Summary |
|---|------|------|-----|---------|
| H | `commands/benchmark/execute.py` | 293 | error-handling | Narrow `except Exception` → `(ValidationError, ValueError)` so a future bug isn't mislabeled "invalid model_params_override." |

**Through-line:** A, C, D, F are all symptoms of the same root cause — generation config is modeled as part of `ModelParams` and attached to the dataset, so the override surface is both too wide (accepts identity/per-run keys) and on the wrong axis. Renaming + key-allowlisting makes the WAR honest now; the `ModelIdentity` / `GenerationConfig` split on `PhaseConfig` is the long-term fix.

_Dedup: builds on — does not duplicate — the maintainer's naming/per-run comment and gemini's shallow-merge note (D explicitly extends it). The `skip_special_tokens` dead-code was already flagged by a prior review and is omitted here._

src/inference_endpoint/config/schema.py
nvzhihanj: This can be done later

src/inference_endpoint/config/schema.py
nvzhihanj: This is valid, basically we are implementing kwargs that can overwrite anything

src/inference_endpoint/config/templates/concurrency_template_full.yaml
arekay-nv: Agree. The name is coming from the source field which show up as the model parameters, which isn't quite right. Since it is a generation config, and so should be a property of the dataset. The reason i say it is a property of the dataset is that we can have a situation where we want different osl for each dataset (like we have repeats for different accuracy datasets). Tying it with the dataset will also enable us to generate dynamic values (distributions) over a dataset independenty.

viraatc (approved):

src/inference_endpoint/config/schema.py
```diff
@@ -313,6 +330,38 @@ class Dataset(BaseModel):
     multi_turn: MultiTurnConfig | None = Field(
         None, description="Multi-turn conversation configuration"
     )
+    # TODO(post-mortem): generation config is per-phase (perf vs. accuracy),
```
viraatc: 👍

arekay-nv:

Reviewed by: **Claude** _(Codex did not complete — see note)_ | Depth: **standard** | Head: `7fe333a`

**No blocking code-level issues found** in the per-dataset `generation_config_override` change.

Verified during review:
- **Layered validation is correct** — unknown top-level keys are rejected in `_validate_generation_config_override`; nested / cross-field invariants (`min_new_tokens > max_new_tokens`, `OSLDistribution` `extra="forbid"`) are rejected at merge time via `ModelParams.model_validate`, caught in `execute.py` and surfaced as `InputValidationError`.
- **`_deep_merge`** preserves nested siblings, handles `base.osl_distribution is None`, and enum values round-trip losslessly through `model_dump()` / `model_validate`.
- **Templates** regenerated and in sync (0 drift); the `multi_turn` → `agentic_inference` rename is consistent across all PR artifacts.
- **Tests** — 27 new tests pass, covering both chat (`max_completion_tokens`) and completions (`max_tokens`) adapter branches, no-override, override propagation, and invalid-value paths.
- **Documented caveat** (per-dataset `name` / `streaming` not honored by the single-aggregator path) matches the code — `execute.py` launches the aggregator/tokenizer from the global `config.model_params`.

🔵 **One optional follow-up (non-blocking):** the perf-path invalid-override branch (`execute.py:329-334`) is structurally identical to the tested accuracy branch but has no direct test — `test_invalid_override_value_raises_input_validation_error` only exercises the accuracy override. Low risk; a one-line parametrization would close it.

---
_Notes:_
- _**Codex** review did not complete: under this host's enterprise-managed Codex profile, `--yolo` was overridden to a read-only sandbox that broke its git / `/tmp` access, and it hit the 10-min timeout with no findings emitted. This was a Claude-only pass._
- _The council reviews the diff for new issues only — it does not resolve the existing **CHANGES_REQUESTED** (field-naming thread) or the PR's **draft** status._

src/inference_endpoint/config/schema.py
```diff
@@ -379,6 +428,40 @@ def _auto_derive_name(self) -> Self:
...
+        ``ModelParams.model_validate`` so type-invalid scalar overrides (e.g.
+        ``temperature: 'hot'``) are rejected. Note that this only catches
+        scalar invalidity — a sparse nested override whose merged result
+        passes default-validation will not raise (callers that need stricter
+        nested validation should set ``base`` to an explicit instance).
+        """
+        if not self.generation_config_override:
+            return base
+        merged = _deep_merge(base.model_dump(), self.generation_config_override)
+        return ModelParams.model_validate(merged)
```
arekay-nv, line 555: [Codex] **medium** · api-contract — per-dataset overrides bypass the `openai_completions`-only guard that top-level `model_params` enforces.

`effective_generation_config` re-validates the merged override **only** as a `ModelParams` sub-model:

```python
merged = _deep_merge(base.model_dump(), self.generation_config_override)
return ModelParams.model_validate(merged)
```

It never re-runs `BenchmarkConfig`'s cross-field guard (`schema.py:1015-1031`), which rejects `min_new_tokens != 1` / `skip_special_tokens` unless `endpoint_config.api_type == openai_completions`. So a dataset override like `generation_config_override: {min_new_tokens: 5}` is **accepted** under `api_type: openai` or `sglang` and then silently ignored by those adapters — whereas the identical setting on top-level `model_params` correctly fails validation. The two config surfaces disagree for exactly the completion-only controls the guard exists to protect.

Suggest applying the same completion-control gate to the merged per-dataset params (or, per the surface-narrowing thread, restrict the override allowlist to API-agnostic keys).

arekay-nv (commented):

Reviewed by: **Codex** + **Claude** | Depth: **standard**

1 net-new issue (most council findings were already covered by existing review comments — see summary).

arekay-nv:

Reviewed by: **Codex** + **Claude** | Depth: **standard** | Head: `da55aff`

Ran two independent reviews (Codex in an isolated worktree + Claude on the diff), verified every cited line against source at HEAD, and deduped against the PR's existing review comments.

**Posted: 1 net-new issue.** Most council findings were already raised in earlier review threads and were suppressed to avoid duplicate noise.

| # | File | Line | Severity | Category | Reviewer(s) | Status |
|---|------|------|----------|----------|-------------|--------|
| 1 | `src/inference_endpoint/config/schema.py` | 463 | medium | api-contract | Codex | ✅ posted inline |

### Suppressed as duplicates of existing comments
| Finding | Reviewer | Already covered by |
|---|---|---|
| Override surface too wide — accepts metric-desyncing keys (`name`/`streaming`/`tokenizer_name`) | Claude | existing "override surface is too wide" + "documented example advertises a key that silently half-works" |
| Override *value* validation deferred to load-time → `inference-endpoint validate` gives false confidence | Claude | existing "value validation of the override is deferred to dataset-load time" |
| Perf-dataset override error branch untested | Claude | existing test-coverage + `except Exception` (perf branch) comments |

### The net-new finding
Codex spotted that `effective_generation_config` re-validates the merged per-dataset override only as a `ModelParams` sub-model, so `BenchmarkConfig`'s cross-field guard (`schema.py:1015-1031`) — which restricts `min_new_tokens`/`skip_special_tokens` to `api_type: openai_completions` — is **bypassed**. A per-dataset override of those fields is accepted under `openai`/`sglang` and silently ignored, while the same setting at top level is correctly rejected. Verified against HEAD.

> ✅ Commit hygiene: single squashed commit, no fixups.

src/inference_endpoint/config/templates/concurrency_template_full.yaml
```diff
@@ -31,6 +31,7 @@ datasets:  # Dataset configs
   generate_params: null  # Dataset-specific parameters passed to the generate() method
   accuracy_config: null  # Accuracy evaluation settings
   agentic_inference: null  # Agentic inference conversation configuration
+  generation_config_override: null  # Per-dataset overrides for the top-level model_params (sparse — only the fields you want to override). Merged on top of BenchmarkConfig.model_params at dataset-load time. Useful for MLPerf-style runs where accuracy and performance use different output budgets in the same fleet, e.g. generation_config_override: {max_new_tokens: 32768, temperature: 0.0}. NOTE: per-dataset `streaming` and `name` are accepted (kwargs-style) but not honored by the single-aggregator metrics path — set those on top-level model_params.
```
nvzhihanj: @arekay-nv can we add a gpt-oss example (per-phase OSL) so it's clear how to use it? And please try a few runs with GPT-OSS to make sure this is same as what we achieved on release/v0.5

src/inference_endpoint/config/templates/concurrency_template_full.yaml
nvzhihanj, line 6: I see the issue of naming now. As a future post-mortem, please break the model_params into model_params and generation_config

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -105,6 +105,28 @@
...
 logger = logging.getLogger(__name__)
 
+# generation_config_override keys that flow into request payloads but are NOT
+# honored by the single global tokenizer / metrics aggregator, so a per-dataset
+# value silently desyncs ISL/OSL/TTFT/TPOT accounting. The override is
+# intentionally permissive (see schema.py), so warn rather than reject.
+_METRICS_DECOUPLED_OVERRIDE_KEYS = frozenset({"name", "streaming", "tokenizer_name"})
+
+
+def _warn_ignored_override_keys(name: str, override: dict[str, Any] | None) -> None:
```
nvzhihanj: If overriding such keys will result in inconsistent behavior, I would suggest just erroring out instead of warning (they should not be overridden anyway)

src/inference_endpoint/config/schema.py
```diff
@@ -374,6 +406,38 @@ class Dataset(BaseModel):
     agentic_inference: AgenticInferenceConfig | None = Field(
         None, description="Agentic inference conversation configuration"
     )
+    # TODO(post-mortem): generation config is per-phase (perf vs. accuracy),
+    # not per-dataset — phases are derived from datasets and the override is
+    # keyed to dataset identity. This lives on Dataset as a short-term WAR
+    # so MLPerf-style accuracy + perf can share one fleet. The proper fix is
+    # a first-class GenerationConfig carried on PhaseConfig, decoupled from
```
nvzhihanj: It seems to me that per-dataset generation config might actually be a valid assumption - given we might want different maxOSL or even topP for different accuracy dataset in the future (what @zihaok found in the DS-V4). We can reword some of the TODOs here.

src/inference_endpoint/config/schema.py
```diff
@@ -59,6 +59,38 @@ class SystemDefaults(BaseModel):
...
+    ``min_new_tokens``/``skip_special_tokens`` are only honored by the
+    ``openai_completions`` adapter; ``BenchmarkConfig`` rejects them for other
+    ``api_type``s. Shared by the top-level and per-dataset-override checks so
+    both config surfaces validate identically.
+    """
+    checks = {
+        "min_new_tokens": mp.min_new_tokens != 1,
+        "skip_special_tokens": not mp.skip_special_tokens,
+    }
+    return [name for name, non_default in checks.items() if non_default]
```
nvzhihanj, line 98: I would argue that separating this to-be-refactored chunk out makes the original code even harder to read/refactor (it took me a while to understand why this function is needed). Recommend making this a in-line function or refactor it

src/inference_endpoint/config/schema.py
```diff
@@ -932,30 +1030,44 @@ def _resolve_and_validate(self) -> Self:
...
+                names = " and ".join(f"{prefix}.{name}" for name in controls)
+                verb = "requires" if len(controls) == 1 else "require"
+                raise ValueError(
+                    f"{names} {verb} endpoint_config.api_type=openai_completions"
+                )
+        for dataset in self.datasets:
+            if dataset.agentic_inference is None:
+                continue
+            effective = (
+                dataset.effective_generation_config(self.model_params)
```
nvzhihanj: Any reason we call this function twice instead of once (and complete the merge) in the beginning?

And this chunk is only checking agentic inference (merging is not needed again)

nvzhihanj (changes requested):

src/inference_endpoint/config/templates/concurrency_template_full.yaml
arekay-nv: Added example.

src/inference_endpoint/config/schema.py
arekay-nv: There are two ways to handle this - one is in the accuracy dataset class (like we do for `[reasoning_effort](https://github.com/mlcommons/endpoints/blob/85a9375df7113a9aa75381367858c806804f0e29/src/inference_endpoint/dataset_manager/transforms.py#L158)` gpt-oss) or as generation configs from the yaml that are configurable at runtime. I think we should move to the generation config approach where everything is either default or comes from the config yaml. That makes it easy to tune and replicate. 
Will update todos.

nvzhihanj (approved):

PR #349, opened 2026-06-09
fix: remove multiturn precompute

tianmu-li:
## What does this PR do?

Move ISL pre-compute to a separate file since ISL is not consumed during run. Also address the O(T^2) concerns in https://github.com/mlcommons/endpoints/pull/336

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [x] Tests added/updated
- [x] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass


scripts/agentic_inference_isl_precompute.py
```diff
@@ -0,0 +1,174 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nvzhihanj, line 1: Recommend changing the file name to agentic_benchmark_isl_precompute.py to be more explicit (and rename future files to agentic to stay consistent)

scripts/agentic_inference_isl_precompute.py
```diff
@@ -0,0 +1,174 @@
...
+    parser.add_argument(
+        "--tokenizer", required=True, help="HuggingFace repo ID or local path."
+    )
+    args = parser.parse_args()
+
+    ds = MultiTurnDataset(pd.read_json(args.dataset, lines=True))
+    ds.load()
+    _precompute_isl(ds, args.tokenizer)
+
+    stats = _isl_distribution(ds)
```
nvzhihanj, line 163: Just to understand, the precompute ISL is strictly for stats purposes, but not for actual runs?

tianmu-li: Correct. This will not be executed during benchmark.

scripts/agentic_inference_isl_precompute.py
tianmu-li: Done

src/inference_endpoint/dataset_manager/multi_turn_dataset.py
```diff
@@ -113,6 +113,127 @@ def _expand_tool_results(row: dict) -> list[dict]:
...
+                msg["content"] = None
+            row_msgs = [msg] if msg.get("role") else []
+
+        if role in ("user", "tool"):
+            t_n = int(row["turn"])
+            current_turn_msgs: list[dict] = row_msgs
+            # Snapshot: history holds everything before this turn; create new lists
+            # so stored snapshots are not mutated by later history extensions.
+            pre_built_messages_by_key[(str_conv_id, t_n)] = (
+                list(history) + current_turn_msgs
```
nvzhihanj, line 202: [Claude] **medium** · design — shared-dict aliasing: the O(T) speedup relies on an undocumented immutability invariant.

Each row's message dict is now formatted **once** and the same dict object is shared across `history`, **every** later `pre_built_messages_by_key` prefix snapshot, and `current_turn_messages_by_key`. (The old per-turn rebuild deep-copied each message into every snapshot, so snapshots were independent.) The comment here guards only the **list** aliasing (`list(history)` makes a fresh list); it does **not** guard the **dict** aliasing. It's safe today — transforms run on the dataframe *before* `_build_metadata`, salt only touches `system_content`, and the runtime paths only read/append — but a future per-turn mutation (a transform that edits a message in place, per-turn salt, an adapter mutating `query.data['messages'][i]`) would silently corrupt **every** turn that shares that dict. Suggest: document the "messages are immutable after build" invariant, and add a regression test that mutates one turn's message and asserts the other turns' prefixes are unchanged.

src/inference_endpoint/dataset_manager/multi_turn_dataset.py
```diff
@@ -113,6 +113,127 @@ def _expand_tool_results(row: dict) -> list[dict]:
...
+            ):
+                msg["content"] = None
+            row_msgs = [msg] if msg.get("role") else []
+
+        if role in ("user", "tool"):
+            t_n = int(row["turn"])
+            current_turn_msgs: list[dict] = row_msgs
+            # Snapshot: history holds everything before this turn; create new lists
+            # so stored snapshots are not mutated by later history extensions.
+            pre_built_messages_by_key[(str_conv_id, t_n)] = (
```
nvzhihanj, line 201: [Claude] **medium** · performance — O(T²) **memory** remains, and is wasted in live/agentic mode.

This PR fixes the O(T²) *time* (history is built in one incremental pass), but `pre_built_messages_by_key` still stores **every** client turn's full prefix → O(T²) message-refs per conversation. Two points:
1. The O(T²) *memory* the change set out to address (#336) is not removed — only the time. For long agentic conversations (hundreds of turns) the stored prefixes still grow quadratically.
2. In **live-history (agentic)** mode it's pure waste: `multi_turn_strategy.py:294` overrides the query with runtime `live_messages = state.message_history.copy() + current_turn_messages`, so the `pre_built[key]` that `load()` builds and attaches to each sample (`sample["messages"] = pre_built[key]`) is **discarded** at issue. Only replay mode (`use_dataset_history=True`) consumes `pre_built`.

Suggest: build/attach `pre_built_messages_by_key` only when `use_dataset_history=True` (live mode needs just `current_turn_messages_by_key`); and note in the PR that the O(T²) memory of storing all prefixes is unchanged (build prefixes lazily at issue if it ever matters for long conversations).

nvzhihanj (commented):
## Review Council — focused review: multi-turn conversation building

Reviewed by: **Claude** (Codex cross-checked the precompute-removal direction) · scope: the `_build_conversation_metadata` rewrite in `multi_turn_dataset.py`.

**Verdict: the rewrite is correct and a genuine improvement.** I diffed it against the merge-base: it is **semantically equivalent** to the old per-turn rebuild (same field extraction for history rows — incl. the pre-existing `tool_calls`/`reasoning_content`; the current turn is always a `user`/`tool` row where those don't apply or go through `_expand_tool_results`), and it changes **O(T²) time → O(T) time**. Turn-key collisions are impossible (the validator enforces turns are exactly consecutive `1..N`); salt only touches `system_content`. Two robustness caveats are posted inline.

**Confirmations (as asked):**
- ✅ **Precompute ISL is stats-only.** `input_tokens` is consumed only by the ISL metric (`session.py:247`) — never for scheduling, max-tokens, or sample counts; for multi-turn it's set to `None` at issue, so it never even reaches the metric. Removing it loses no run-time behavior.
- ✅ **Output metrics are OSL/timing-based.** OSL = `token_count(output)`, TPOT = output-duration/output-tokens, TTFT = timing — all independent of ISL.

**Inline caveats:**
1. *(design)* Shared-dict aliasing — the O(T) speedup shares message dicts across all prefix snapshots; safe today but relies on an undocumented immutability invariant (the old code deep-copied). Document + test.
2. *(performance)* O(T²) **memory** of `pre_built_messages_by_key` is unchanged (only time was fixed), and is built-then-discarded in live/agentic mode — skip it when `use_dataset_history=False`.

_Aside (low, pre-existing, not posted inline): prior-turn assistant `reasoning_content` is replayed in the prefix; real agentic deployments usually strip prior CoT — a prompt-fidelity question, out of scope for this PR._

viraatc (approved):

PR #350, opened 2026-06-09
perf(metrics): batch tokenization with defer-to-flush drain

viraatc:
## What

ISL/OSL/TPOT need a tokenizer pass per completion. `main` dispatches one
asyncio task per event into a 2-thread pool — at high completion rates the
backlog grows unboundedly and the end-of-run drain takes ~an hour per million
samples. This PR batches: triggers enqueue O(1); a small live lane keeps live
metrics current; the end-of-run drain tokenizes everything left through a
process-sharded pool that uses the whole machine.

## How

- **`BatchTokenizer`** — the drain runs `encode_batch_fast` (Rust, rayon)
  across auto-sized worker processes, one pinned per 8-core block of the
  allowed CPU universe (probed via `expand_to_all_online_cpus()`, then the
  aggregator's inherited mask is **restored** — the service stays wherever
  the parent placed it). No silent fallbacks: a tokenizer without a fast
  backend, or a failed/over-budget warmup, is a clean startup error. macOS
  shards unpinned (rayon capped per worker) at full speed.
- **Live lane** — in-process threads (`--metrics-tokenizer-workers`, schema
  default 2, the pre-existing knob and footprint; `0` = defer everything to
  the drain), rayon-capped, slice-capped per flush. Owned by the queue
  (`start_live`); the publisher knows nothing about tokenization.
- **`TokenBatchQueue`** — buffers `(text, on_count)` per event; live
  failures/cancellations re-queue items (no sample loss), drain failures are
  terminal and stay counted in `n_pending_tasks` (incomplete-drain contract:
  `state == complete && n_pending_tasks > 0`). Drain budget `--drain-timeout`
  (default 60 s, `0` = unlimited); finalize always runs.
- **`MetricsTable`** is fully synchronous; `CORES_PER_WORKER` is a module
  constant. Defaults are single-sourced in `config/schema.py`
  (`metrics_drain_timeout_s` 60 s, `metrics_tokenizer_workers` 2); the
  service args are required and always forwarded by the benchmark.

## Validation

- Unit suite green (176 metrics-aggregator: queue contract, shard sizing,
  drain timeout/failure, live requeue, RAYON caps, wiring seams);
  `pre-commit` clean. Offline-burst e2e: `state=complete`, all series
  populated, drain to `n_pending_tasks=0`.
- **Sharding is default-on through the real launch path** (verified on a
  48-core x86 host and a 144-core GB200): the drain shards span the machine
  while the aggregator keeps its inherited mask.

### Tokenizer micro-benchmark (GB200, real DeepSeek-R1 tokenizer)

144-core Grace, corpus = MLPerf DS-R1 prompts tiled to the dataset-mean OSL
of 3877 tokens; identical token counts both sides.

| impl    | parallelism                | texts/s | tokens/s | speedup |
| ------- | -------------------------- | ------: | -------: | ------: |
| `main`  | 2 threads, per-text encode |     313 |   1.21 M |       — |
| this PR | 18 shards, batched encode  |  11,951 |   46.3 M | **38×** |

### 1M-sample end-to-end A/B vs `main`

Offline 1M samples, streaming, DS-R1 tokenizer, server-paced at 8k QPS with
~1k-token outputs. Both sides: 1,000,000/1,000,000, `state=complete`,
`n_pending_tasks=0`, identical token series.

| host       | impl    | backlog at `ENDED` | drain   | total    | speedup   |
| ---------- | ------- | -----------------: | ------: | -------: | --------: |
| GB200 144c | `main`  |          2,970,972 | 3,362 s | 58.1 min |         — |
| GB200 144c | this PR |          2,782,912 |  42.9 s |  3.2 min | **18.1×** |
| B200 192c  | `main`  |          2,994,925 | 3,286 s | 56.9 min |         — |
| B200 192c  | this PR |          2,788,032 |    61 s |  3.4 min | **16.5×** |

> Measured on the final design (in-process live lane, `--tokenizer-workers 2`,
> 300 s drain default). The live lane keeps ~7% of tokenizations current; the
> rest (~2.78M) defer to the end-of-run drain, which the sharded pool clears in
> 43-61 s. A 1M-sample run needs the 300 s budget — 60 s drops the backlog.
> `main` rows (unlimited drain budget, or they never finish) and the
> micro-benchmark are unaffected.


nvzhihanj (commented):
## Review Council — re-audit after rebase (HEAD `4633699d`)

Reviewed by: **Claude** + **Codex** (low reasoning, correctness pass) · Depth: **thorough**. Focus (as requested): is the metrics-tokenization change **modular / clean / non-intrusive** (existing benchmark behavior preserved), and are **redundant/meaningless tests** added.

**Verdict: the rebase reduced the intrusiveness but did not resolve it.** Replacing the `pre_publish` hook (full tokenizer pool fired from the publish tick) with a bounded single-shard live lane is a real improvement. But mid-run tokenization is **still on by default** (`--live-tokenizers` defaults to 1 → 0.25s live flush), the live shard runs on the **highest core block which overlaps the HTTP workers** (`compute_affinity_plan` Phase-3 spillover), and the PR **removed the only opt-out** (`metrics_tokenizer_workers`) without replacement — so the observability component can perturb the SUT during measurement and the operator can't turn it off via config. Headline recommendation: default `--live-tokenizers 0` for measurement-grade runs (defer all tokenization to the post-run drain), or confine the live shard to cores disjoint from `worker_cpu_sets`; restore a benchmark-reachable knob. (A1, A2, A3.)

**Otherwise clean / non-intrusive.** The change stays in the aggregator subprocess (only cross-module touch: importing `endpoint_client.cpu_affinity`). The consumer contract is **verified intact** — `SessionState`, the `MetricsSnapshot` schema, publisher cadence, and the `state==COMPLETE && n_pending_tasks>0` incomplete-drain signal are unchanged; `flush_remaining` is bounded by the drain budget and never raises; the live-loop's failure cannot skip `publish_final`. The "shard or exit cleanly" fallback and the unpinned-without-affinity (macOS) path are correct and tested.

**Tests: no redundant or meaningless tests.** The new branches are mostly well covered with behavior-grounded assertions (the `_setup_shards` decision matrix, no-fast-backend-exit, unpinned-without-affinity, warmup-failure-exit, `flush_remaining` timeout/failure, live-loop start/stop/survives-failure, `expand_to_all_online_cpus`). Removing the old `metrics_tokenizer_workers` tests was correct (dead). The problems are **coverage gaps, not redundancy**: the aggregator-side `start_live` wiring is untested (A5) and `TestAggregatorArgs` no longer pins the forwarded-args contract (A6). Two `_FakeProc`-injection tests are borderline-coupled to internals but still verify fan-out/reassembly; `TestEvenChunks` is trivial-but-cheap. No mock-only or duplicate tests found.

**Codex findings — not posted:** (1) a multi-turn-ISL precompute regression at `execute.py:351` — that's PR #349's change, out of scope here; (2) a `shutdown(wait=False)` worker-terminate race — `_terminate_procs` already defensively handles `_processes is None` and CPython doesn't synchronously null it, so the specific mechanism couldn't be verified → dropped. Existing `gemini`/`github-code-quality` `token_metrics.py` comments (flush-exception inflight; closed-tokenizer guards; `close()` shutdown leak; Protocol `...`→`pass`) are unaddressed but deduped here, not re-posted.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -565,12 +565,6 @@ async def _run_benchmark_async(
         aggregator_args.extend(
```
nvzhihanj, line 724: [Claude] **high** · design — mid-run tokenization is ON by default with no benchmark-reachable opt-out.

`aggregator_args` here forwards only `--tokenizer` and `--drain-timeout`; it passes neither `--live-tokenizers` nor `--tokenizer-workers`. So `--live-tokenizers` takes its argparse default of **1** → `__main__.py` sets `live_flush_interval_s = publish_interval` (0.25s) → the live tokenizer lane runs **during the measurement window**. This PR also removed the `metrics_tokenizer_workers` config field, so the documented `--live-tokenizers 0` escape ("defer everything to the end-of-run drain") is only reachable by hand-invoking the subprocess — the benchmark never does. **Why:** the metrics aggregator is an observability side-component that must not perturb the SUT, but the operator can't turn mid-run tokenization off through the supported config/CLI surface. **Suggested:** for measurement-grade runs default to deferring all tokenization to the post-run drain (`--live-tokenizers 0`), and/or re-introduce a config field forwarded from here; an operator who wants a live dashboard opts in.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -60,140 +90,306 @@ def _normalize_tool_calls_for_template(
...
+        highest core blocks, farthest from the loadgen's low cores — so live
+        token metrics never contend with the benchmark hot path. The
+        end-of-run drain uses every shard.
+        """
+        if not texts:
+            return []
+        live_n = max(1, self._live_workers)
+        if self._procs and live_n < len(self._procs):
+            return await self._fan_out(self._procs[-live_n:], texts)
+        return await self.count_texts_async(texts, loop)
```
nvzhihanj: [Claude] **high** · performance — the live shard overlaps the SUT's worker cores during measurement.

`count_texts_live_async` fans out to `self._procs[-live_n:]` — the **highest** core block(s) (each shard is pinned by `_init_worker` to `sorted(sched_getaffinity(0))[i*8:(i+1)*8]` after `expand_to_all_online_cpus()` widened the mask to the whole machine). The docstring says this is "farthest from the loadgen's low cores … so live token metrics never contend with the benchmark hot path," but that only reasons about the loadgen **main** process: `compute_affinity_plan` gives the loadgen the fastest Primary-NUMA cores (Phase 2) and hands the **HTTP workers everything else, including the highest cores** (Phase-3 spillover). So the live shard's high block lands on worker cores, and mid-run tokenization contends with the SUT every 0.25s. **Why:** directly perturbs the throughput/latency being measured. **Suggested:** confine the live shard to cores explicitly disjoint from `affinity_plan.worker_cpu_sets` (plumb that in), or default live tokenization off; and correct the docstring.

docs/async_utils/services/metrics_aggregator/DESIGN.md
```diff
@@ -0,0 +1,182 @@
...
+unpinned: blocks are sized from the online CPU count and each worker caps its
+rayon pool to the block size instead of pinning.
+
+Chat-template items (tool-call outputs) take a separate in-process thread:
+they are rare relative to the batched flush, and `apply_chat_template` is
+Python/Jinja — sharding buys nothing. A template baseline (the empty
+assistant-message frame) is computed once and subtracted so only the payload
+is counted.
+
+### CPU affinity: the tokenizer stage is post-run
```
nvzhihanj: [Claude] **medium** · documentation — this section's premise is false for the default benchmark path.

"the tokenizer stage is post-run" / "Mid-run tick flushes are small batches; the drain is where the core count pays" — but with `--live-tokenizers` defaulting to **1**, the live lane flushes every `publish_interval` (0.25s) **during** the run, and at high QPS each flush is thousands of completions tokenized on a rayon shard whose cores overlap the SUT's HTTP workers (see the `token_metrics.py` comment). This is the stated justification for resetting the inherited loadgen mask to the whole machine, so the incorrect premise is load-bearing. **Suggested:** state that mid-run live flushing is on by default at the publish cadence and overlaps worker cores, document how to disable it, and (per AGENTS.md "describe current state") drop the "post-run" framing.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -311,9 +322,16 @@ async def process(self, records: list[EventRecord]) -> None:
...
                                     self._session_state,
-                                    table.in_flight_tasks_count,
+                                    self.pending_tokens,
                                 ),
                             )
+                            if (
+                                self._token_queue is not None
+                                and self._live_flush_interval_s is not None
+                            ):
+                                self._token_queue.start_live(
```
nvzhihanj, line 344: [Claude] **medium** · testing — the aggregator-side live-lane wiring is untested.

This `self._token_queue.start_live(self._live_flush_interval_s)` branch (gated on `_token_queue is not None and _live_flush_interval_s is not None`, fired on STARTED) is the headline contention path, yet nothing asserts the aggregator actually starts the live loop — `TokenBatchQueue.start_live` is tested in isolation, but not its wiring here. **Why:** a regression that silently stops starting (or wrongly starts) the live loop would be invisible — exactly the behavior most relevant to SUT intrusion. **Suggested:** add a test that processes a STARTED event with `live_flush_interval_s=0.05` and asserts the live task is running, plus the converse with `None`.

tests/unit/commands/test_benchmark.py
```diff
@@ -548,7 +541,6 @@ def test_yaml_roundtrip(self, tmp_path):
         assert drain.performance_timeout_s == 30.0
         assert drain.accuracy_timeout_s is None
         assert drain.metrics_drain_timeout_s == 300.0
-        assert drain.metrics_tokenizer_workers == 8
 
 
 class TestAggregatorArgs:
```
nvzhihanj, line 598: [Claude] **medium** · testing — the aggregator-args contract is now under-pinned at the exact SUT-intrusion seam.

`TestAggregatorArgs` asserts only `--drain-timeout`. There is no test that `--tokenizer` is forwarded when `ctx.tokenizer_name` is set, and — because live/worker args are intentionally not forwarded — nothing pins the decision that the benchmark relies on argparse defaults for live tokenization. Removing the old `metrics_tokenizer_workers` tests was correct, but it left this contract almost entirely unasserted. **Why:** the set of forwarded args is where the non-intrusiveness behavior is decided; it should be asserted so a future change (e.g. re-adding `--live-tokenizers`) is deliberate. **Suggested:** add a `--tokenizer`-forwarded test (and a `--live-tokenizers`-forwarded test if the opt-out finding is addressed).

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
```diff
@@ -204,15 +224,34 @@ async def main() -> None:
...
+    # (coalesces to 'object' not 'AbstractContextManager[BatchTokenizer | None]')
+    tokenizer_cm: AbstractContextManager[BatchTokenizer | None]
     if args.tokenizer:
-        pool_cm = TokenizePool(args.tokenizer, n_workers=args.tokenizer_workers)
+        # Tokenization drains after the benchmark run, so the loadgen/worker
+        # affinity partition does not apply to this stage: drop the narrow
+        # mask inherited from the pinned parent so shards size to the whole
+        # machine (cgroup/Slurm CPU limits still apply).
+        try:
+            cpus = expand_to_all_online_cpus()
```
nvzhihanj: [Claude] **low** · design — `expand_to_all_online_cpus()` runs unconditionally on the tokenizer path, not gated to the drain phase.

It resets the aggregator's inherited (narrow, loadgen-derived) mask to the whole machine whenever `--tokenizer` is set. That's correct for the post-run drain (which legitimately wants every core), but it's also the mechanism by which the **live** shard reaches SUT worker cores (see the `token_metrics.py` finding). **Why:** the whole-machine expansion is run-wide rather than drain-scoped. **Suggested:** if keeping live tokenization on, expand to the full machine only at drain time and keep the live shard confined to a smaller block; otherwise document that the expansion is whole-run, not post-run.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -96,7 +96,7 @@ class MetricCounterKey(str, Enum):
 _TOKEN_HDR_LOW: Final[int] = 1
 _TOKEN_HDR_HIGH: Final[int] = 10_000_000  # 10M tokens
 
-_DEFAULT_DRAIN_TIMEOUT_S: Final[float] = 60.0
+_DEFAULT_DRAIN_TIMEOUT_S: Final[float] = 300.0
```
viraatc: use old limit 60s

docs/async_utils/services/metrics_aggregator/DESIGN.md
```diff
@@ -0,0 +1,119 @@
...
+counted in `pending`. `flush_remaining` never raises.
+
+### Sharded batch encoding (`BatchTokenizer`)
+
+The drain fans the whole buffer out across worker **processes**, one pinned
+per `CORES_PER_WORKER` (8) core block. Each worker runs the raw `tokenizers`
+backend's `encode_batch_fast` (Rust, rayon); a single BPE rayon pool
+saturates ~8 cores, so disjoint pinned blocks are how the whole machine is
+used. Workers are spawn-context, warmed in parallel at construction (bounded
+— a hung load is a startup error), and ignore SIGINT.
```
arekay-nv, line 58: If we ignore SIGINT, there is no way to terminate/interrupt?

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
```diff
@@ -132,13 +133,13 @@ async def main() -> None:
     parser.add_argument(
         "--drain-timeout",
         type=float,
-        default=60.0,
+        required=True,
```
arekay-nv: Should this be required? What about a default of 0 which is what most use-cases would expect. Enforcing a limit can unnecessarily cause runs to fail for hitting the timeout?

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
```diff
@@ -162,8 +163,14 @@ async def main() -> None:
     parser.add_argument(
         "--tokenizer-workers",
         type=int,
-        default=2,
-        help="Number of tokenizer worker threads (default: 2)",
+        required=True,
+        help=(
```
arekay-nv: Maybe a good default would be nice here (0?) - we might be overloading the user with too many required flags/options which are really tuning knobs.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -367,41 +394,47 @@ async def process(self, records: list[EventRecord]) -> None:
...
+                    budget = (
+                        f"{self._drain_timeout_s:.1f}s"
+                        if self._drain_timeout_s is not None
+                        else "unlimited"
+                    )
+                    logger.warning(
+                        "tokenizer drain incomplete (budget %s); %d tokenizations "
+                        "did not complete",
+                        budget,
+                        n_pending,
```
arekay-nv, line 429: if `drain_timeout_s` is 0, hence unlimited, this statement would be executed if we had errors in the tokenization? Do we need to make that differentiation to clarify.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -367,41 +394,47 @@ async def process(self, records: list[EventRecord]) -> None:
...
+                    )
+                logger.info(
```
arekay-nv: should this be `else`?

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -223,48 +227,32 @@ def _compute_value(
...
+        if self._queue is None:
+            return
+        message_parts = self._extract_message(ev_rec, row, pre_change)
+        if message_parts is not None:
+            self._queue.enqueue_message(
+                message_parts, self._make_recorder(ev_rec, pre_change)
+            )
+            return
         text = self._extract_text(ev_rec, row, pre_change)
         if not text:
```
arekay-nv, line 253: This would be reached if text is none and message_parts is none, should we count this as an anomaly?

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -365,32 +351,24 @@ def _extract_message(self, ev_rec, row, pre_change):
         return None
 
 
-class TpotTrigger(AsyncTokenTrigger):
-    """TPOT = (complete_ns - recv_first_ns) / token_count(text_after_first_chunk).
+class TpotTrigger(TokenTrigger):
+    """TPOT = (complete_ns - recv_first_ns) / output token count.
```
arekay-nv, line 358: Can we emphasize that this excludes TTFT. A minor but subtle difference in how the TPOT can be counted - as one can count it on the client side `(compelted_response-request_sent)/num_tokens`as well.

src/inference_endpoint/config/schema.py
```diff
@@ -578,34 +578,36 @@ class DrainConfig(BaseModel):
...
         ),
     ] = Field(
-        60.0,
+        300.0,
         ge=0,
         description=(
-            "Wall-clock budget (seconds) for the metrics aggregator to drain "
-            "in-flight tokenize tasks after ENDED (default: 60.0; 0 = unlimited)."
+            "Wall-clock budget (seconds) to finish tokenizing buffered samples "
+            "after ENDED (default: 300.0; 0 = unlimited)."
```
arekay-nv: Suggestion to change to `0` as default. 
Plus ignore above comments on the defaults there - this unifies the defaults to one place.

arekay-nv (approved):
LGTM. Thanks!

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
nv-alicheng: Well if it didn't finish tokenizing, it's invalid anyway - the token count (used in TPS / TPOT calcs) will be lower than the true value.

A more robust solution would be that if this times out, we provide a subcommand to re-calculate these offline given the events.jsonl to recompute the result_summary.json - keeps the run valid and doesn't waste the allocation.

nv-alicheng: Also this flag is never *exposed* to the user as a non-developer user would never launch this by hand. The default I assume is maintained by the defaults in the schema.

src/inference_endpoint/async_utils/services/metrics_aggregator/__main__.py
nv-alicheng: Also this flag is never *exposed* to the user as a non-developer user would never launch this by hand. The default I assume is maintained by the defaults in the schema.

src/inference_endpoint/config/schema.py
nv-alicheng: Nvm I now see you found out 😆 

IMO timeout should be on the order of something like 10 min - we have seen veryyy large backlogs in the tokenizer queue before.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -203,70 +432,255 @@ def _token_count_message_worker(
...
         """
-        if self._executor is None:
-            raise RuntimeError("TokenizePool is closed")
-        return await loop.run_in_executor(
-            self._executor, self._token_count_worker, text
-        )
+        _terminate_procs(self._procs)
+        self._procs = []
+        if self._thread is not None:
+            self._thread.shutdown(wait=True)
```
nv-alicheng: [Review Council · Codex] **medium · concurrency** — `close()` does `self._thread.shutdown(wait=True)` on the live tokenizer thread pool. If the run ends while a live flush is in `run_in_executor(...)`, cancelling the queue's live task only cancels the coroutine — the executor thread keeps tokenizing, and this `wait=True` blocks process teardown until that encode finishes. For large batches with `metrics_tokenizer_workers > 0` this can exceed the `--drain-timeout` budget the rest of the PR enforces. (Distinct from the adjacent note on the *process* pools, which correctly use `wait=False` + SIGTERM via `_terminate_procs` — the gap is the live *thread* pool.) Relatedly, the SIGTERM path (`__main__.py` `_signal_finalize`) never stops the live flush loop the way the ENDED path does via `flush_remaining`. Stop/cancel the live lane (and bound the in-flight encode) before the blocking `shutdown(wait=True)`.

src/inference_endpoint/endpoint_client/cpu_affinity.py
```diff
@@ -317,6 +317,32 @@ def pin_loadgen(
...
+    after ``pin_loadgen`` inherit the loadgen mask). The kernel intersects the
+    request with the cgroup cpuset, so container/Slurm CPU limits still apply.
+
+    Returns:
+        The effective CPU set after the reset.
+
+    Raises:
+        UnsupportedPlatformError: If not running on Linux.
+    """
+    online = _read_sysfs_cpulist(_SYSFS_CPU / "online") or set()
```
nv-alicheng, line 334: [Review Council · Codex] **medium · bug** — When `_read_sysfs_cpulist(_SYSFS_CPU / 'online')` returns empty (sysfs unavailable/filtered — common in some containers), this skips the `sched_setaffinity` widening and returns `os.sched_getaffinity(0)`, i.e. the narrow inherited loadgen mask. The metrics-aggregator sharding sizes its drain pool from this, so in those environments the shard pool stays pinned to the loadgen cores and can hit the drain timeout on large runs instead of expanding to the full allowed cpuset. Fall back to widening against the cgroup cpuset (or at least log) when `online` is unreadable, rather than silently returning the narrow mask.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -203,70 +432,255 @@ def _token_count_message_worker(
...
+                    raise
+                except Exception as exc:  # noqa: BLE001 — isolate phases.
+                    failure = exc
+                    if live:
+                        # A live hiccup must not lose samples: give the items
+                        # back so the end-of-run drain (full pool) retries.
+                        # Drain failures are terminal and stay pending-only.
+                        self._text[:0] = text_items
+                else:
+                    for (_, on_count), count in zip(text_items, counts, strict=True):
```
nv-alicheng: [Review Council · Claude] **medium · data-integrity** — In the drain (`live=False`) path both phases are detached up front. The text-phase `zip(text_items, counts, strict=True)` runs in the `else:` of the `count_texts_async` try, so a wrong-length tokenizer result raises `ValueError` *after* recording the matched prefix and *outside* the independent-phase `failure` handling — it propagates straight out of `flush()`, skipping the message phase. The detached `msg_items` are then lost (never recorded, never re-queued; `_inflight` stays elevated). A wrong-length result is a tokenizer contract violation, but it's the one spot trusting output length without isolation — catch the `ValueError` inside the phase and route it through `failure` so the message phase still runs.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -60,140 +101,328 @@ def _normalize_tool_calls_for_template(
...
+        # without the override every shard would run at the live-lane width.
+        os.environ["RAYON_NUM_THREADS"] = str(len(core_set))
+        try:
+            os.sched_setaffinity(0, set(core_set))
+        except (OSError, AttributeError):
+            # No pinning (e.g. macOS): the rayon cap above still keeps
+            # unpinned shards from oversubscribing each other.
+            logger.debug("could not pin tokenizer worker to %s", core_set)
+    transformers_logging.set_verbosity_error()
+    tok = AutoTokenizer.from_pretrained(tokenizer_name, trust_remote_code=True)
```
nv-alicheng, line 140: [Review Council · Claude] **low · security** — Each spawned shard worker loads the tokenizer with `AutoTokenizer.from_pretrained(tokenizer_name, trust_remote_code=True)`, executing arbitrary repo-hosted Python at import time. The main process already did this pre-PR, but sharding now replicates remote-code execution across N worker processes. The tokenizer name is operator-supplied (not a new trust boundary), but it broadens the blast radius. If `trust_remote_code` isn't required for the supported tokenizers, consider defaulting it off or gating behind config.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -203,70 +432,255 @@ def _token_count_message_worker(
...
+        """End-of-run flush, bounded by ``timeout`` seconds.
+
+        Stops the live flush loop, then drains through the full shard pool.
+        Returns the number of items still un-tokenized — non-zero if the budget
+        was exhausted (``timeout`` reached) or tokenization failed. ``None``
+        waits indefinitely. Never raises: a failure here must not stop the
+        aggregator from publishing the (incomplete) final snapshot.
+        """
+        if self._live_task is not None:
+            self._live_task.cancel()
```
nv-alicheng, line 696: [Review Council · Claude] **low · error-handling** — `flush_remaining` documents 'Never raises' (the aggregator depends on this to always publish a final snapshot) and guards `flush()` in `except Exception`, but the live-task teardown above it is unguarded: `self._live_task.cancel(); await asyncio.gather(self._live_task, return_exceptions=True)`. If `flush_remaining` is itself cancelled (outer timeout / shutdown cancelling `process()`), the `await` raises `CancelledError` before reaching the guarded `flush()`, breaking the absolute 'never raises' contract. Document `CancelledError` as the one exception, or shield the teardown.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -60,140 +101,328 @@ def _normalize_tool_calls_for_template(
...
+def _terminate_procs(procs: list[ProcessPoolExecutor]) -> None:
+    """Best-effort immediate stop: cancel queued work and SIGTERM workers.
+
+    ``shutdown(wait=False)`` alone leaves an in-flight encode running, and the
+    non-daemon worker would still be joined at interpreter exit — so a drain
+    timeout could stall process shutdown until the chunk finished.
+    """
+    for ex in procs:
+        ex.shutdown(wait=False, cancel_futures=True)
+        workers = getattr(ex, "_processes", None) or {}  # CPython impl detail.
```
nv-alicheng: [Review Council · Claude] **low · concurrency** — `_terminate_procs` reaches into CPython-private `_processes`: `workers = getattr(ex, '_processes', None) or {}`. The guard degrades to a silent no-op if a future CPython changes the attribute name/shape — reintroducing exactly the interpreter-exit stall this function exists to prevent, with no signal. Since it's the only thing keeping a drain timeout from stalling process exit, log once if the private attribute is missing so the regression is observable.

src/inference_endpoint/async_utils/services/metrics_aggregator/aggregator.py
```diff
@@ -117,23 +115,33 @@ def __init__(
         publish_interval_s: float,
         sig_figs: int,
         n_histogram_buckets: int,
-        tokenize_pool: TokenizePool | None = None,
+        tokenizer: BatchTokenizer | None = None,
+        live_flush_interval_s: float | None = None,
         streaming: bool = False,
         shutdown_event: asyncio.Event | None = None,
-        drain_timeout_s: float | None = _DEFAULT_DRAIN_TIMEOUT_S,
+        drain_timeout_s: float | None,
```
nv-alicheng: [Review Council · Claude] **low · api-contract** — `drain_timeout_s: float | None` is a required keyword arg declared *after* defaulted keyword-only params (`tokenizer=None`, `live_flush_interval_s=None`, `streaming=False`, `shutdown_event=None`). Legal because keyword-only, but easy to misread as optional, and every test constructing the aggregator must remember to pass it. Group required params before defaulted ones, or give it an explicit `None` default (already means 'unlimited' downstream).

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -60,140 +101,328 @@ def _normalize_tool_calls_for_template(
...
+    # -- batched text path --------------------------------------------------
+
+    def _encode_lengths_inproc(self, texts: list[str]) -> list[int]:
+        tok = self._tokenizer
+        backend = getattr(tok, "backend_tokenizer", None)
+        if backend is not None:
+            return _encode_batch_lengths(backend, texts)
+        return [len(tok.tokenize(t)) for t in texts]  # type: ignore[union-attr]
+
+    async def count_texts_async(
```
nv-alicheng, line 356: [Review Council · Claude] **low · design** — The `TokenCounter` Protocol declares `count_texts_async(self, texts, loop, /, *, live=False)` and `token_count_message_async(..., loop, /)` with positional-only markers, but the concrete `BatchTokenizer` methods (here and `token_count_message_async`) omit the `/`. It type-checks (concrete is wider) and all call sites pass positionally, so no runtime bug — but the declared contract and implementation disagree. Align the concrete signatures with the Protocol's `/`.

docs/async_utils/services/metrics_aggregator/DESIGN.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+The ENDED path runs inside a finalization boundary: whatever the drain does —
+finish, time out, or fail — `publish_final` and the shutdown signal always
+run. A tokenizer failure can degrade the snapshot (see the `n_pending_tasks`
+contract) but can never hang the subprocess. SIGTERM writes a best-effort
+partial snapshot tagged `INTERRUPTED`.
+
+## Token metrics pipeline
+
```
nv-alicheng, line 28: [Review Council · Claude] **low · documentation** — The lifecycle diagram draws `INTERRUPTED` as a branch off `DRAINING`, but the SIGTERM handler is installed before `STARTED` and `_signal_finalize` publishes `interrupted=True` regardless of state — so INTERRUPTED is reachable from any state (INITIALIZE/LIVE included), which is exactly the case the handler exists for. The adjacent prose even says 'SIGTERM writes a best-effort partial snapshot' without requiring DRAINING. Redraw INTERRUPTED as reachable from any state.

nv-alicheng (commented):

Posted **9** inline findings. Well-engineered PR — no critical/high defects; the drain/flush hot path is largely correct. Findings are condition-gated edge cases + polish. See summary comment for the tiered breakdown + one untested-path note that couldn't be posted inline (handler code unchanged in the diff).

nv-alicheng:

9 inline findings across **4** files. **No critical/high defects** — the defer-to-flush hot path is largely correct: the batch is detached before tokenization, drain failures are terminal-and-pending-only (no double-count, no silent loss on the normal paths), `publish_final`/`_finalize` run on every terminal path, idempotency is guarded, and the schema is the single source of truth for defaults (drain-timeout 300s, tokenizer-workers 2 consistent across schema, all 3 templates, AGENTS.md, DESIGN.md). aiohttp 3.14.1 bump consistent in pyproject + uv.lock. Findings below are condition-gated edge cases and polish.

### 🟡 Should Fix (medium)

| 1 | `metrics_aggregator/token_metrics.py` | 461 | concurrency | Codex | `close()` `self._thread.shutdown(wait=True)` on the live thread pool can block teardown past `--drain-timeout` while a live encode is in-flight; SIGTERM path also doesn't stop the live loop |
| 2 | `endpoint_client/cpu_affinity.py` | 334 | bug | Codex | sysfs `online` unreadable → no widening → shard pool pinned to narrow loadgen mask → drain timeout on large runs in sysfs-filtered containers |
| 3 | `metrics_aggregator/token_metrics.py` | 626 | data-integrity | Claude | drain-phase `zip(..., strict=True)` in the `else` bypasses the independent-phase failure handling: a wrong-length tokenizer result records a partial prefix then propagates, dropping the detached message-phase items |

### 🔵 Consider (low)

| 4 | `metrics_aggregator/token_metrics.py` | 135 | security | Claude | `trust_remote_code=True` now replicated across N shard worker processes; broadens blast radius |
| 5 | `metrics_aggregator/token_metrics.py` | 666 | error-handling | Claude | `flush_remaining` 'never raises' contract broken by unguarded live-task teardown vs `CancelledError` |
| 6 | `metrics_aggregator/token_metrics.py` | 170 | concurrency | Claude | `_terminate_procs` silently no-ops if CPython-private `_processes` shape changes, reintroducing the exit-stall it prevents |
| 7 | `metrics_aggregator/aggregator.py` | 122 | api-contract | Claude | required kw `drain_timeout_s` declared after defaulted kw-only params; reads as optional |
| 8 | `metrics_aggregator/token_metrics.py` | 361 | design | Claude | concrete `count_texts_async`/`token_count_message_async` omit the Protocol's positional-only `/` |
| 9 | `metrics_aggregator/DESIGN.md` | 24 | documentation | Claude | lifecycle diagram draws INTERRUPTED only off DRAINING; SIGTERM can interrupt from any state |

**Not posted inline** (handler code unchanged in the diff, so no valid inline anchor):
- `__main__.py` `_on_sigterm`/`_signal_finalize` — **testing** (Claude): the SIGTERM-during-active-drain race (handler fires while `flush_remaining` is mid-drain) is untested. Single-INTERRUPTED-snapshot / `shutdown_event`-set-once / `n_pending_tasks` correctness rests on reasoning, not a test. Recommend a test firing the handler against a slow in-flight `flush_remaining`.

> ⚠️ **Commit hygiene**: 20 commits including ~8 apparent fixups (`fix(metrics): …`, `chore(metrics): drain-timeout default back to 60s`, etc.). Consider squashing the iteration history before merge.

nv-alicheng:
## Design note — scope over architecture

Independent of the line-level findings already posted, a framing for the rework as a whole. The architecture here is forced by the constraints and the PR gets it right: tokenize is heavy CPU work that must not block the loop, must not contend with the loadgen mid-run, must keep up at 50k+ QPS, and must yield exact final numbers. GIL ⇒ real parallelism needs processes; a single BPE rayon pool saturates ~8 cores; the work is only needed for metrics ⇒ it can defer past the run. Those facts make **defer-to-flush batching** + **process-sharding pinned to disjoint core blocks** essentially the only answer, and I'd converge on the same — along with no-silent-fallback and the exact-or-flagged `n_pending_tasks` contract.

Several subtle calls are better than a first pass would make: the sharding-by-core-block thesis is **measured** (16k vs 1.5k texts/s), the warmup is a bounded startup error that races the launch budget, shards stay idle until `ENDED` so tokenization never perturbs the running benchmark, and the DESIGN.md actually documents the contracts. Good work.

The one theme worth weighing is **scope, not structure**: the live in-process tokenization lane roughly doubles the edge-case surface and accounts for most of the review findings (live-cancel, re-queue-on-failure, the live thread-pool blocking teardown, the bounded per-flush cap). The authoritative metrics are computed at the drain regardless; the live lane exists only to keep token metrics current in the TUI mid-run, and `--tokenizer-workers=0` ("defer all") is already the simpler universe.

So if live token metrics are not a hard requirement, a KISS v1 would be **drain-only**, adding the live lane later if operators ask for live OSL. That removes the `live` flag threaded through `flush`/`count_texts_async`, the re-queue paths, the live-loop lifecycle, and the second executor. If live IS required, the lane as built is a legitimate, well-engineered answer — and the gap then narrows to two refactors:

1. **Make CPU-universe discovery a pure query.** The probe currently does `getaffinity → expand_to_all_online_cpus() (mutates affinity as a side effect) → restore`. The save/probe/restore dance and its restore-failure branch only exist because the helper isn't pure. A pure "list allowed CPUs" query removes the dance and gives a clean cgroup-cpuset fallback when `/sys/.../online` is unreadable (the affinity finding, addressed at the design level rather than patched).

2. **Split `flush()` into `flush_text` + `flush_messages`.** One method juggling a shared `failure` var, two re-queue paths, and a `zip(strict=True)` in the `else` is what produced the data-integrity finding. Two small methods with independent error handling remove that hazard and read better.

None of this is a rework — the bones are right and more thoroughly justified than most perf PRs. It's about trimming optional scope out of v1 and making the two trickiest spots (affinity, the dual-phase flush) simpler.

_(Independent design review by Claude at a maintainer's request — not a re-run of the automated council above.)_

PR #356, opened 2026-06-13
feat: per-dataset max_new_tokens override

roborluo:
## What does this PR do?

When running a combined performance + accuracy benchmark in a single --mode both invocation, the two phases want opposite generation caps, but today the harness only exposes one global model_params.max_new_tokens:
 - Performance phase needs a small cap. max_new_tokens is sent to the server as the per-request max_tokens, and a disaggregated decode scheduler reserves/plans decode-KV for that declared upper bound — even though generation actually stops at EOS far sooner. A large cap (e.g. 32768) over-reserves decode KV (~3.2× vs 10240), starves admittable decode slots at high concurrency, and triggers KV-transfer-timeout storms on the context→gen path. A realistic small cap avoids this.
 - Accuracy phase needs a large cap, otherwise long reasoning outputs get truncated and scores are artificially deflated. This matches the MLPerf Inference gpt-oss-120b reference, where the performance and accuracy workloads use different token settings — see language/gpt-oss-120b → [Model and Dataset download](https://github.com/mlcommons/inference/tree/master/language/gpt-oss-120b#model-and-dataset-download). Without a per-dataset override, you cannot satisfy both in one --mode both run.

  ## Type of change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Documentation update
  - [ ] Refactor/cleanup

  ## Related issues

  
  N/A

  ## Testing

  - [x] Tests added/updated
  - [x] All tests pass locally

  

  ## Checklist

  - [x] Pre-commit hooks pass


arekay-nv:
@roborluo Can you look at #344  which addresses the same issue. We can consolidate the two here and merge this one. I think the other one has also the templates correctly populated which you are failing in CI.

PR #357, opened 2026-06-14
feat: accuracy issuer inherits perf concurrency in online mode

arekay-nv:
## What

When the performance phase runs the **CONCURRENCY** load pattern (online mode), the accuracy phase now mirrors that same fixed concurrency instead of always bursting at `MAX_THROUGHPUT`. This makes accuracy evaluation exercise the endpoint the same way the performance run does.

## Behavior

| Perf load pattern | Accuracy phase (before) | Accuracy phase (after) |
| --- | --- | --- |
| `concurrency` (online) | `max_throughput` | **`concurrency` (same `target_concurrency`)** |
| `poisson` (online) | `max_throughput` | `max_throughput` (unchanged) |
| `max_throughput` (offline) | `max_throughput` | `max_throughput` (unchanged) |

`POISSON` and offline `MAX_THROUGHPUT` deliberately keep the accuracy phase at `MAX_THROUGHPUT` — inheriting `POISSON` would silently rate-limit evaluation to the perf QPS, and there's no accuracy QPS-budgeting yet. The gate is purely `load_pattern.type == CONCURRENCY`, which the schema already constrains to online mode (`schema.py`), so no separate test-type check is needed.

The accuracy issuer also now logs its chosen load mode per accuracy dataset, e.g.:

```
Accuracy issuer 'aime' load mode: concurrency (target_concurrency=64)
```

## Scope

- Localized to `_build_phases()` in `commands/benchmark/execute.py` (runs once at setup — no hot-path impact).
- No schema/CLI/template changes (automatic behavior, no new flag).

## Tests

Added to the existing `TestBuildPhases` suite in `tests/unit/commands/test_benchmark.py`:
- `test_accuracy_phase_inherits_perf_concurrency` — perf `concurrency(7)` → accuracy `concurrency`, `target_concurrency == 7`
- `test_accuracy_phase_max_throughput_when_perf_poisson` — perf `poisson` → accuracy stays `max_throughput`
- `test_accuracy_phase_max_throughput_when_perf_offline` — offline → accuracy stays `max_throughput`
- `test_accuracy_issuer_logs_load_mode` — asserts the issuer logs its mode

Verification: `tests/unit/commands` + `tests/unit/load_generator` → **266 passed**; `pre-commit` (ruff, ruff-format, mypy, license) clean.

### DIFF
--- a/src/inference_endpoint/commands/benchmark/execute.py
+++ b/src/inference_endpoint/commands/benchmark/execute.py
@@ -465,11 +465,29 @@ def _build_phases(
                 f"Accuracy dataset '{eval_cfg.dataset_name}' is a MultiTurnDataset, "
                 "which is not yet supported for accuracy evaluation."
             )
-        # Accuracy phases run at MAX_THROUGHPUT; inheriting perf_lp (e.g. POISSON)
-        # would silently rate-limit evaluation until a multi-turn accuracy strategy
-        # and QPS-budgeting support are added.
-        acc_load_pattern: LoadPattern | None = LoadPattern(
-            type=LoadPatternType.MAX_THROUGHPUT
+        # When the perf phase runs CONCURRENCY (online), the accuracy phase
+        # mirrors that fixed concurrency so evaluation exercises the endpoint
+        # the same way. Every other pattern keeps MAX_THROUGHPUT — inheriting
+        # POISSON would silently rate-limit evaluation to the perf QPS, which
+        # has no QPS-budgeting support for accuracy yet.
+        perf_lp = ctx.rt_settings.load_pattern
+        acc_load_pattern: LoadPattern | None
+        if perf_lp is not None and perf_lp.type == LoadPatternType.CONCURRENCY:
+            acc_load_pattern = LoadPattern(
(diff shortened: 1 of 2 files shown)


src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -465,11 +465,29 @@ def _build_phases(
                 f"Accuracy dataset '{eval_cfg.dataset_name}' is a MultiTurnDataset, "
                 "which is not yet supported for accuracy evaluation."
             )
-        # Accuracy phases run at MAX_THROUGHPUT; inheriting perf_lp (e.g. POISSON)
-        # would silently rate-limit evaluation until a multi-turn accuracy strategy
-        # and QPS-budgeting support are added.
-        acc_load_pattern: LoadPattern | None = LoadPattern(
-            type=LoadPatternType.MAX_THROUGHPUT
+        # When the perf phase runs CONCURRENCY (online), the accuracy phase
```
nvzhihanj, line 468: The load generator should use the same load pattern as the performance run to be fair. I will merge it for now but this is post-mortem item that we need to fix.

nvzhihanj (approved):

PR #368, opened 2026-06-22
Revert "fix: [release/v0.5] Upgrade msgpack - GHSA-6v7p-g79w-8964"

arekay-nv:
Reverts mlcommons/endpoints#367


nvzhihanj:
Instead of adding one reverting commit, you can directly rewrite the history and remove the commit, then force push so the commit history is clean.

arekay-nv:
> Instead of adding one reverting commit, you can directly rewrite the history and remove the commit, then force push so the commit history is clean.
Done.

PR #370, opened 2026-06-22
feat(client): round-robin outbound connections across source IPs

viraatc:
## Summary

Streaming inference benchmarks hold one TCP connection per in-flight request for the request's whole duration. On a single client host, concurrency is therefore capped by the OS **ephemeral-port range** (Linux default `32768–60999`, ~28k ports). Past that ceiling new connections fail with `EADDRNOTAVAIL`, and because the client waits for a port to free, the delay is absorbed into **TTFT** rather than surfaced as an error — an easy-to-miss throughput wall at high concurrency.

TCP 4-tuple uniqueness `(src_ip, src_port, dst_ip, dst_port)` is **per source IP**, so binding outbound connections across *N* local source IPs gives *N* independent ephemeral-port spaces to the same destination — multiplying the usable connection budget by ~*N* on a single host. Hosts with multiple NICs already have several usable source IPs; on Linux the entire `127.0.0.0/8` loopback block is usable without configuration.

## Changes

- **`config.py`** — new `source_ips: list[str]` option (default empty = current OS-default behavior). The ephemeral-port budget clamp is scaled by the source-IP count, so `max_connections=-1` (auto) resolves to `available_ports × len(source_ips)`, and an explicit `--max-connections` is validated against the scaled budget. Blank entries are dropped so a stray `""` can't bind to `INADDR_ANY`.
- **`http.py`** — `ConnectionPool` round-robins `local_addr` across `source_ips` on each new connection. Empty list is normalized to `None`, so the default path keeps zero overhead and unchanged behavior.
- **`worker.py`** — threads `source_ips` from `HTTPClientConfig` into each per-worker pool.

## Usage

```
--source-ips 10.0.0.1 --source-ips 10.0.0.2 --source-ips 10.0.0.3
```

Round-robins outbound connections across the three IPs, ~tripling the single-host connection ceiling. Empty (default) preserves today's behavior exactly.

## Verification

- **Unit tests (8 new):** round-robin `local_addr` assignment + empty/None normalization (`test_http.py`); budget scaling, explicit-budget validation, and blank-IP cleaning (`test_http_client_config.py`).
- **OS-level proof:** with an emulated small ephemeral window, a single source IP established exactly *W* connections to one destination; 5 source IPs established exactly *5W* (**5.00× measured**). Confirmed the same client port coexists across source IPs, and that exhausting a single IP's window raises `EADDRNOTAVAIL` — i.e. spreading across source IPs removes the single-IP ceiling.

## Notes

- **Backward compatible:** empty `source_ips` is byte-for-byte the prior behavior (kernel autobind, single ephemeral-port space).
- IPs must be assigned to local interfaces (multiple NICs, or `127.0.0.0/8` aliases on Linux); otherwise `bind()` fails with a clear error.
- Generated full-config templates (`config/templates/*_full.yaml`) are intentionally not modified in this PR.

### DIFF
--- a/src/inference_endpoint/endpoint_client/config.py
+++ b/src/inference_endpoint/endpoint_client/config.py
@@ -107,6 +107,27 @@ class HTTPClientConfig(WithUpdatesMixin, BaseModel):
         ),
     ] = Field(-1, ge=-1)
 
+    # Client source IPs to bind outbound connections to, round-robined per new
+    # connection (see ConnectionPool._create_connection).
+    #
+    # Each distinct source IP has an independent ephemeral-port space to a given
+    # destination — TCP 4-tuple (src_ip, src_port, dst_ip, dst_port) uniqueness
+    # is per src_ip — so binding across N source IPs multiplies the usable
+    # connection budget by ~N on a single host, lifting the ~28k single-IP
+    # ephemeral-port ceiling that otherwise caps concurrency for streaming runs.
+    #
+    # IPs must be assigned to local interfaces (multiple NICs, or loopback
+    # aliases in 127.0.0.0/8 on Linux). Empty = OS default source selection
+    # (unchanged single-IP behavior).
+    source_ips: Annotated[
+        list[str],
+        cyclopts.Parameter(
+            alias="--source-ips",
+            help="Client source IPs to round-robin connections across "
+            "(multiplies ephemeral-port budget by IP count). Empty=OS default.",
+        ),
+    ] = Field(default_factory=list)
+
     # Transport configuration
     transport: AnyTransportConfig = Field(
         default_factory=TransportConfig.create_default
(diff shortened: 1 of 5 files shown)


viraatc:
Superseded by #371.

Follow-up investigation showed the ephemeral-port budget is **already** multiplied across distinct destinations by the kernel — the limit is per `(source IP, destination)` pair, so each endpoint gets its own budget (validated: 1 endpoint → 999 simultaneous connections, 5 endpoints → 4975, ~995 each). So explicit source-IP binding isn't needed for the common multi-frontend case.

The real fix is just to stop the auto `max_connections` clamp from capping total concurrency at a *single* pair's budget when multiple endpoints are configured. #371 does that with **no new config or CLI** — the multiplier comes from the `endpoint_urls` already specified. Closing this in favor of #371.

PR #377, opened 2026-06-26
feat(report): record run_config (seeds + load/warmup) in result_summary.json

nvzhihanj:
## Summary
Record the run's **load config + RNG seeds** in the benchmark report so a run is self-describing and validatable from its own `result_summary.json` (and `report.txt`).

A single `run_config` field on `Report` carries:
- `load_pattern` (type / target_qps / concurrency / `use_legacy_loadgen_qps_metrics` — needed to interpret qps/tps),
- `warmup` (incl. `warmup_random_seed`),
- `scheduler_random_seed` + `dataloader_random_seed`.

`load_pattern` / `warmup` come from `settings.model_dump(include={"load_pattern", "warmup"})`; the seeds are read off `settings.runtime`. `endpoint_config` (api_key / endpoint URLs) is a **sibling** of `settings` and never included, so no secrets leak. The full, re-runnable config still lives in `config.yaml` alongside.

**Replaces the standalone `seeds` field (#353)**: the RNG seeds now live under `run_config`.

> **Scope note:** the *resolved / effective* runtime settings (the actual issued sample count and sample ordering, which can differ **per audit phase** — e.g. MLPerf TEST04's fixed-sample output-caching phase) are intentionally **left to a follow-up PR**. Getting those right means sourcing from the live `rt_settings` per phase, which is a separate change. This PR only records the static, run-identifying config above.

## Example `result_summary.json`
From a real **DGX-B300x8 DeepSeek-R1 server (poisson)** run — `qps`/`tps` stay top-level, run config under `run_config` (metric dicts elided as `{ ... }`):
```jsonc
{
  "version": "0.1.0",
  "git_sha": "...",
  "n_samples_completed": 40000,
  "duration_ns": 3328465383747,
  "state": "complete",
  "complete": true,
  "ttft": { ... }, "tpot": { ... }, "latency": { ... }, "output_sequence_lengths": { ... },
  "legacy_loadgen_window_duration_ns": 2753346633148,
  "qps": 14.53,
  "tps": 54489.98,
  "run_config":
  {
    "load_pattern": {
      "type": "poisson",
      "target_qps": 14.75,
      "target_concurrency": null,
      "use_legacy_loadgen_qps_metrics": true
    },
    "warmup": {
      "enabled": false,
      "n_requests": null,
      "salt": true,
      "drain": false,
      "warmup_random_seed": 42
    },
    "scheduler_random_seed": 42,
    "dataloader_random_seed": 42
  }
}
```
Same, rendered in `report.txt`:
```
Run config:
  load_pattern: type=poisson, target_qps=14.75, target_concurrency=None, use_legacy_loadgen_qps_metrics=True
  warmup: enabled=False, n_requests=None, salt=True, drain=False, warmup_random_seed=42
  scheduler_random_seed: 42
  dataloader_random_seed: 42
```

## Tests
- `run_config` keyword-only passthrough; serialization + `report.txt` rendering; absent -> `null`.

Rebased on latest `main` (keeps PR #372's `use_legacy_loadgen_qps_metrics`; git_sha handled by #398).

### DIFF
--- a/src/inference_endpoint/commands/benchmark/execute.py
+++ b/src/inference_endpoint/commands/benchmark/execute.py
@@ -1110,16 +1110,28 @@ async def _run_benchmark_async(
 
                 if snap_dict is not None:
                     try:
-                        runtime = ctx.config.settings.runtime
-                        warmup = ctx.config.settings.warmup
                         load_pattern = ctx.config.settings.load_pattern
+                        runtime_cfg = ctx.config.settings.runtime
+                        # load_pattern + warmup config and the RNG seeds, so
+                        # result_summary.json is self-describing and a valid run is
+                        # identified by its settings. The full, re-runnable config
+                        # lives in config.yaml alongside. The resolved/effective
+                        # runtime settings (sample count + ordering, which can differ
+                        # per audit phase) are deferred to a follow-up. endpoint_config
+                        # (api_key/URLs) is a sibling of settings and never included,
+                        # so no secrets.
+                        run_config = ctx.config.settings.model_dump(
+                            mode="json", include={"load_pattern", "warmup"}
+                        )
+                        run_config["scheduler_random_seed"] = (
+                            runtime_cfg.scheduler_random_seed
+                        )
+                        run_config["dataloader_random_seed"] = (
+                            runtime_cfg.dataloader_random_seed
+                        )
                         report = Report.from_snapshot(
                             snap_dict,
-                            seeds={
-                                "scheduler_random_seed": runtime.scheduler_random_seed,
-                                "dataloader_random_seed": runtime.dataloader_random_seed,
-                                "warmup_random_seed": warmup.warmup_random_seed,
-                            },
+                            run_config=run_config,
                             use_legacy_loadgen_qps_metrics=(
                                 load_pattern.type == LoadPatternType.POISSON
                                 and load_pattern.use_legacy_loadgen_qps_metrics
--- a/src/inference_endpoint/metrics/report.py
+++ b/src/inference_endpoint/metrics/report.py
@@ -146,25 +146,28 @@ class Report(msgspec.Struct, frozen=True):  # type: ignore[call-arg]
     qps: float | None = None
     tps: float | None = None
 
-    # RNG seeds for this run (scheduler/dataloader/warmup, from config). Carried
-    # so result_summary.json is self-validating: a reproducible run is identified
-    # by its seeds. These are config, not a measured metric, so the from_snapshot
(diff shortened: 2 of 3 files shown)


src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1097,16 +1097,20 @@ def _on_phase_start(phase: PhaseConfig) -> None:
...
+                        # Snapshot the run's load/runtime config straight from the
+                        # Pydantic settings (single source of truth) — captures RNG
+                        # seeds, load pattern, sample count, durations, etc. without
+                        # hand-listing fields. `client` is excluded (transport
+                        # plumbing, not run-defining); endpoint_config (api_key/URLs)
+                        # is a sibling of settings, so it is not included either.
+                        run_config = ctx.config.settings.model_dump(
+                            mode="json",
+                            include={"runtime", "load_pattern", "warmup"},
+                        )
```
arekay-nv, line 1131: Should we replicate sections of the config.yaml here - if the goal is reproducibility, we will still need the config.yaml file to run unless we include the whole thing. Otherwise we might end up with inconsistency between the two.

viraatc: The saved `run_config` can differ from what actually ran.

Audit phases override the sample count and ordering in `ctx.rt_settings`. For example, TEST04 runs one phase with normal sampling and another that repeatedly issues one fixed sample. This code instead saves `ctx.config.settings`, which does not contain those overrides. Consequently, both phases produce identical `run_config` blocks and report `n_samples_to_issue: null`.

Suggested fix: record effective runtime settings from `ctx.rt_settings`, including the resolved sample count from `total_samples_to_issue()`, `sample_order.kind`, and `sample_order.fixed_index`. Consider placing these under `run_config["resolved_runtime"]` so the original user configuration remains distinct from resolved values. Please add an integration assertion that TEST04 reference and output-caching phase configurations differ.

viraatc (approved):

arekay-nv (approved):

src/inference_endpoint/commands/benchmark/execute.py
nvzhihanj: Makes sense, I see run config is a bit nasty because we can have different duration, samples to be issued in different phases. Let me just put the seed, load pattern and warmup for now, need to think about what's the best way to put runtime settings (multi-phase), whether copied from config yaml or resovled from internal class

nvzhihanj: Resolving and merging, thanks for the review

PR #351, opened 2026-06-11
feat : deepseek r1 accuracy support (release/v0.5)

arekay-nv:
## What does this PR do?
Enables DSR1 accuracy checks.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally
- [ ] Manual testing completed

## Checklist

- [ ] Code follows project style
- [ ] Pre-commit hooks pass
- [ ] Documentation updated (if needed)

PR #378, opened 2026-06-27
feat : deepseek r1 accuracy support (release/v0.5) (#351)

arekay-nv:
---------

Cherry-pick DSR1 support back from the release v0.5 branch which has been thoroughly tested.

Adds end-to-end MLPerf **DeepSeek-R1** combined-subset accuracy evaluation. DeepSeek-R1's accuracy set is an ensemble of five subsets (`math500`, `aime`, `gpqa`, `mmlu_pro`, `livecodebench`), each parsed and graded differently. This PR wires up dataset loading, the perf/accuracy phases, and a new scorer that produces a single headline `exact_match` on the same 0–100 scale as the MLPerf golden number (81.3582).

### What's included

| Area | Change |
|------|--------|
| **Scorer** | `DeepSeekR1Scorer` (`scorer_id="deepseek_r1"`) in `evaluation/scoring.py` |
| **Isolated evaluator** | New uv subproject `evaluation/deepseek_r1/` (runner, `pyproject.toml`, `uv.lock`, `setup_eval.sh`, `RUNBOOK.md`) — excluded from the parent wheel |
| **Dataset** | `DeepSeekR1` loader (`dataset_id="deepseek_r1"`) in `dataset_manager/predefined/deepseek_r1/` + registration in `dataset_manager/__init__.py` |
| **Plumbing** | `commands/benchmark/execute.py`: per-scorer `complete` flag plumbed into results; clearer error on bad `accuracy_config.extras`; configurable service-ready timeout |
| **Shared client** | `_lcb_ws_evaluate()` LiveCodeBench WebSocket client, shared by `LiveCodeBenchScorer` and `DeepSeekR1Scorer` |
| **Docs/CI/tests** | `AGENTS.md`, CI workflow tweaks, `transforms.py`, and new unit tests for the dataset + scorer |

### Design notes

- **Out-of-process evaluation.** The official MLCommons `eval_accuracy.py` pulls in heavy, conflicting pins (`transformers`, the `prm800k` math grader, the LiveCodeBench executor). Following the existing `VBenchScorer` pattern, the evaluator runs via `uv run --project` against the isolated subproject; the parent process never imports it. Path resolution is explicit arg → `$DEEPSEEK_EVAL_PROJECT_PATH` → editable-checkout fallback.
- **Pre-tokenized prompt.** The dataset emits an `input_tokens` column (MLPerf `tok_input`) so the `openai_completions` adapter's `Harmonize()` is a no-op and the server chat template is bypassed — the model sees the exact MLPerf prompt. One loader serves both the perf phase (issues `input_tokens`) and the accuracy phase (grades by `dataset`/`ground_truth`).
- **LiveCodeBench sandboxing.** LCB executes untrusted model code, which the in-process MLCommons executor can't sandbox. When a port is configured, the `livecodebench` subset is graded out-of-band against the `lcb-service` WebSocket container; text subsets go through the subprocess; results are merged into one 5-subset number.
- **Partial-result honesty.** A new `Scorer.complete` flag (default `True`) is set `False` whenever the headline can't cover every issued sample — failed text subset, diverging LCB count, or an unreachable `lcb-service` container (LCB left unscored rather than silently shrinking the denominator). The flag is surfaced in results and logs so a partial number is never mistaken for a complete one.
## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass


src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -105,6 +106,31 @@
...
 
+_SERVICE_READY_TIMEOUT_ENV = "INFERENCE_ENDPOINT_SERVICE_READY_TIMEOUT_S"
+_SERVICE_READY_TIMEOUT_DEFAULT = 30.0
+
+
+def _service_ready_timeout() -> float:
+    """Service-startup readiness timeout, overridable via env.
+
+    Service imports off a shared/Lustre FS can exceed the default under heavy
+    login-node I/O contention. A non-numeric override is ignored with a warning
```
nvzhihanj: This seems to be a hack and non-parameterized field?

src/inference_endpoint/dataset_manager/predefined/deepseek_r1/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+from pathlib import Path
+
+import pandas as pd
+
+from ...dataset import Dataset
+
+logger = getLogger(__name__)
+
+#: Env var pointing at the local MLPerf DeepSeek-R1 source dataset.
+SOURCE_ENV = "DEEPSEEK_R1_DATASET_PKL"
```
nvzhihanj: Using MACRO to point dataset is not a good design, especially using a macro to point to another macro. We can push the parquet file to the repo since it's small anyway

src/inference_endpoint/dataset_manager/predefined/deepseek_r1/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+    One loader serves both phases: the perf phase issues ``input_tokens`` and the
+    accuracy phase hands the rows to ``DeepSeekR1Scorer`` (which grades by
+    ``dataset``/``ground_truth``).
+
+    The source is not bundled (it is the official MLCommons dataset). Point
+    ``$DEEPSEEK_R1_DATASET_PKL`` at your local ``.pkl`` (or an already-prepared
+    ``.parquet``), or pass ``source=`` to :meth:`generate`.
+    """
+
+    COLUMN_NAMES = _OUTPUT_COLUMNS
```
nvzhihanj: Macro pointing to macro

src/inference_endpoint/dataset_manager/predefined/legacy_mlperf_deepseek_r1/__init__.py
```diff
@@ -0,0 +1,167 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nvzhihanj, line 1: we should probably call the dataset legacy_mlperf_deepseek_r1_dataset or something similar. Calling a dataset ds-r1 will be confusing

src/inference_endpoint/dataset_manager/predefined/legacy_mlperf_deepseek_r1/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+
+        if max_samples is not None and max_samples < len(df):
+            full_n = len(df)
+            df = cls._stratified_subset(df, max_samples, seed)
+            logger.info("Stratified subset: %d of %d rows", len(df), full_n)
+        return df
+
+    @staticmethod
+    def _build_from_source(source: str | os.PathLike | None) -> pd.DataFrame:
+        resolved = source or os.environ.get(SOURCE_ENV)
```
nvzhihanj, line 128: Basically we cannot build from source, and have to use existing one. We can probably raise not implemented here and only allow local usage of the parquet then
(Otherwise the logic here is quite confusing)

src/inference_endpoint/evaluation/deepseek_r1/pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+#   # DeepSeekR1Scorer in the parent will shell out automatically.
+
+[project]
+name = "deepseek-r1-mlperf-accuracy"
+version = "0.1.0"
+description = "Isolated MLCommons DeepSeek-R1 accuracy evaluation environment."
+requires-python = ">=3.10"
+dependencies = [
+    # eval_accuracy.py core
+    "pandas>=2.0",
```
nvzhihanj: Would recommend locking the version == instead of >= to avoid maintenance issues

src/inference_endpoint/evaluation/deepseek_r1/RUNBOOK.md
```diff
@@ -0,0 +1,106 @@
+# DeepSeek-R1 Accuracy Evaluator Subproject
```
nvzhihanj: Not the right place for documentation. Example is the better place.

src/inference_endpoint/evaluation/deepseek_r1/setup_eval.sh
```diff
@@ -0,0 +1,56 @@
...
+PRM800K_COMMIT="7ecc794703b2877f63226f2477a49b34f9b25163"
+LCB_REPO="https://github.com/LiveCodeBench/LiveCodeBench"
+LCB_COMMIT="28fef95ea8c9f7a547c8329f2cd3d32b92c1fa24"
+
+EVAL_ACCURACY_URL="https://raw.githubusercontent.com/mlcommons/inference/${MLC_INFERENCE_COMMIT}/language/deepseek-r1/eval_accuracy.py"
+
+mkdir -p "${SUBMODULES_DIR}"
+
+echo "==> Fetching eval_accuracy.py @ ${MLC_INFERENCE_COMMIT}"
+curl -fsSL "${EVAL_ACCURACY_URL}" -o "${EVAL_DIR}/eval_accuracy.py"
```
nvzhihanj: I understand we need to fetch the LCB_REPO, but if we are only fetching one file from MLCommon inference repo, can we just place it here?

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1723,374 @@ def score(self) -> tuple[float | None, int]:
         per_dim_scores = self._extract_per_dim_scores(results)
         mean_score = float(np.mean(per_dim_scores))
         return mean_score, n_repeats
+
+
+_DEFAULT_DEEPSEEK_EVAL_PROJECT_PATH = Path(__file__).resolve().parent / "deepseek_r1"
+
+_DEEPSEEK_EVAL_PROJECT_PATH_ENV = "DEEPSEEK_EVAL_PROJECT_PATH"
```
nvzhihanj: MACRO pointing to Macro (one-time)

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1723,374 @@ def score(self) -> tuple[float | None, int]:
...
+                f"Run `uv sync` and `bash setup_eval.sh` in the accuracy "
+                f"subproject, or set ${_DEEPSEEK_EVAL_PROJECT_PATH_ENV} to the "
+                f"synced subproject path."
+            )
+
+    @staticmethod
+    def _resolve_project_path(explicit: os.PathLike | None) -> Path:
+        """Resolve the DeepSeek eval subproject path.
+
+        Lookup order: explicit ctor arg -> ``$DEEPSEEK_EVAL_PROJECT_PATH`` env
```
nvzhihanj, line 1738: What would be the use case of this - and should we allow such ENV override for the eval directory?

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1723,374 @@ def score(self) -> tuple[float | None, int]:
...
+
+    def _score_lcb_via_container(self, lcb_df: pd.DataFrame) -> tuple[int, int] | None:
+        """Grade the livecodebench rows against the lcb-service WebSocket.
+
+        Extracts the python block from each ``model_output`` and keys it by
+        question id (the ``ground_truth``), then evaluates via the container.
+        Returns ``(passed, total)`` or None if the service is unreachable (so
+        score() can fall back to the in-process path).
+        """
+        assert self.lcb_websocket_url is not None
```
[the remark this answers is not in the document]
nvzhihanj: +1, raise is probably better

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1723,374 @@ def score(self) -> tuple[float | None, int]:
...
+        log_path.write_text(completed.stdout or "")
+        if completed.returncode != 0:
+            tail = "\n".join((completed.stdout or "").splitlines()[-50:])
+            raise RuntimeError(
+                f"DeepSeek eval subprocess exited with code "
+                f"{completed.returncode}; full log at {log_path}. "
+                f"Last 50 lines:\n{tail}"
+            )
+
+    def _score_lcb_via_container(self, lcb_df: pd.DataFrame) -> tuple[int, int] | None:
```
nvzhihanj, line 1814: Should this code be reused from the GPT-OSS LCB?

scripts/Dockerfile.dev
```diff
@@ -45,3 +48,16 @@ RUN uv sync --frozen --no-install-project --extra dev --extra test
...
+
+# Provision the isolated DeepSeek-R1 accuracy evaluator so the image is self-contained for
+# `inference-endpoint benchmark --mode acc|both` with NO host checkout and NO runtime network.
+# DeepSeekR1Scorer shells out via `uv run --project src/inference_endpoint/evaluation/deepseek_r1`;
+# that subproject pins heavy/conflicting deps (transformers, prm800k, LiveCodeBench code
+# executor) that must stay OUT of /opt/venv. So override UV_PROJECT_ENVIRONMENT to build the
+# subproject's OWN .venv, then setup_eval.sh fetches eval_accuracy.py + prm800k + LiveCodeBench
+# into mlperf_eval/. (At run time the launcher must likewise point UV_PROJECT_ENVIRONMENT +
+# DEEPSEEK_EVAL_PROJECT_PATH at this subproject, since the image ENV defaults to /opt/venv.)
+RUN cd src/inference_endpoint/evaluation/deepseek_r1 && \
```
arekay-nv: [Both] **high** (bug): `cd src/inference_endpoint/evaluation/deepseek_r1` targets a directory this PR renamed to `legacy_mlperf_deepseek_r1`. The `RUN` fails with `No such file or directory`, breaking the dev image build before `uv sync`/`setup_eval.sh` — so `mlperf_eval/` is never baked. Update the path. (Flagged independently by both reviewers.)

scripts/Dockerfile.dev
```diff
@@ -45,3 +48,16 @@ RUN uv sync --frozen --no-install-project --extra dev --extra test
 # Copy source and install project
 COPY --chown=${USER_ID}:${GROUP_ID} src/ ./src/
 RUN uv sync --frozen --extra dev --extra test
+
+
+# Provision the isolated DeepSeek-R1 accuracy evaluator so the image is self-contained for
+# `inference-endpoint benchmark --mode acc|both` with NO host checkout and NO runtime network.
+# DeepSeekR1Scorer shells out via `uv run --project src/inference_endpoint/evaluation/deepseek_r1`;
```
arekay-nv: [Claude] **medium** (documentation): Stale pre-rename references: this comment says `DeepSeekR1Scorer ... uv run --project .../deepseek_r1` (now `LegacyMLPerfDeepSeekR1Scorer` / `legacy_mlperf_deepseek_r1`). Line 19 (`the deepseek_r1 evaluator's setup_eval.sh`) is likewise stale and misleads operators wiring up the image.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1620,375 @@ def score(self) -> tuple[float | None, int]:
...
+        lcb_entry = dict(per_dataset.get(self.lcb_subset, {}))
+        if lcb_scored is None:
+            # Container unreachable: leave livecodebench UNSCORED. Do NOT re-run
+            # the in-process executor - it can't sandbox runaway model code and
+            # needs a ~21 GB dataset load. Score LCB separately with
+            # score_livecodebench.sh and fold it in.
+            logger.warning(
+                "LegacyMLPerfDeepSeekR1Scorer: lcb-service unreachable at %s; livecodebench "
+                "left unscored (reporting %d text samples only, run marked "
+                "incomplete). Score LCB separately via score_livecodebench.sh.",
```
arekay-nv: [Codex] **medium** (documentation): The LCB-unreachable warning tells users to `Score LCB separately via score_livecodebench.sh`, but that script is not committed in the repo (`git ls-files` finds nothing). This violates the AGENTS.md rule against referencing non-committed artifacts and leaves no actionable recovery path — point at the documented lcb-service container instead, or commit the script.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1620,375 @@ def score(self) -> tuple[float | None, int]:
...
+                else (e.stdout or b"").decode("utf-8", errors="replace")
+            )
+            log_path.write_text(partial)
+            raise RuntimeError(
+                f"DeepSeek eval subprocess timed out after "
+                f"{self.subprocess_timeout_s}s; see {log_path} for partial output."
+            ) from e
+
+        log_path.write_text(completed.stdout or "")
+        if completed.returncode != 0:
```
arekay-nv, line 1806: [Claude] **medium** (testing): `_run_eval_subprocess`'s two error branches are untested: the `TimeoutExpired` path (partial-log write + `RuntimeError`) and this `returncode != 0` path (last-50-lines tail + `RuntimeError`). The scorer tests only patch `subprocess.run` to return `returncode=0`, so a regression in either ships silently.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1620,375 @@ def score(self) -> tuple[float | None, int]:
...
+        total_n = text_n + lcb_total
+        combined = 100.0 * (text_correct + lcb_passed) / total_n if total_n else None
+
+        # The headline number is only valid if it covers every issued sample;
+        # a failed text subset or a diverging LCB count silently shrinks total_n.
+        expected_n = len(eval_df)
+        complete = bool(
+            combined is not None and text_complete and lcb_ok and total_n == expected_n
+        )
+        if combined is not None and lcb_ok and total_n != expected_n:
```
arekay-nv, line 1997: [Claude] **medium** (testing): The LCB count-divergence guard (`total_n != expected_n` → warn + mark the run incomplete) has no test — no case drives `total_n != expected_n`, so the guard that stops a partial number being reported as complete is uncovered. (An older, now-outdated comment sits at this line; this is a separate, live gap.)

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -318,6 +324,65 @@ def score(self) -> tuple[float, int]:
...
+        logger.warning(
+            "websocket-client not installed; cannot reach lcb-service. "
+            "Install with: pip install websocket-client"
+        )
+        return None
+    try:
+        ws = websocket.create_connection(
+            url, timeout=7200, ping_interval=30, ping_timeout=10
+        )
+    except (ConnectionRefusedError, OSError, Exception) as e:  # noqa: BLE001
```
arekay-nv: [Claude] **low** (error-handling): `except (ConnectionRefusedError, OSError, Exception)`: the first two are redundant with `Exception`, and catching bare `Exception` on the connect path swallows programming errors (e.g. a malformed `url` → `TypeError`) as a 'connect failed' warning, returning `None` so the caller silently falls back instead of surfacing the bug. Narrow to `(OSError, websocket.WebSocketException)`.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1620,375 @@ def score(self) -> tuple[float | None, int]:
...
+        text_complete = True
+        for sub, d in per_dataset.items():
+            if sub == self.lcb_subset:
+                continue
+            em = d.get("exact_match")
+            n = int(d.get("num_samples", 0))
+            if em is None:  # subset failed to grade (status != "ok")
+                text_complete = False
+                continue
+            text_correct += round(em / 100.0 * n)
```
arekay-nv, line 1953: [Claude] **low** (bug): `text_correct += round(em / 100.0 * n)` reconstructs the integer correct-count from a per-subset mean — exact only while per-sample accuracy is strictly binary (0/100). Nothing asserts that invariant; a future MLCommons subset with fractional/partial credit would silently miscount with no error. Add a comment pinning the assumption, or sum raw counts from the runner.

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/deepseek_eval_runner.py
```diff
@@ -0,0 +1,189 @@
...
+        all_scored = pd.concat(scored_frames, ignore_index=True)
+        overall_exact_match: float | None = float(all_scored["prompt_accuracy"].mean())
+        evaluated_samples = int(len(all_scored))
+    else:
+        overall_exact_match = None
+        evaluated_samples = 0
+
+    results = {
+        "exact_match": overall_exact_match,
+        "tokens_per_sample": float(df["tok_model_output_len"].mean()),
```
arekay-nv: [Claude] **low** (data-integrity): Run as a standalone CLI on an all-external/all-failed (or empty) frame, `tokens_per_sample = float(df['tok_model_output_len'].mean())` is `NaN`; `json.dumps(NaN)` emits invalid JSON that the scorer's `msgspec.json.decode` rejects. The scorer guards the empty case upstream, but this file is documented as a standalone entry point — guard with `0.0` on empty/NaN.

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/deepseek_eval_runner.py
```diff
@@ -0,0 +1,189 @@
...
+logger = logging.getLogger("deepseek_eval_runner")
+
+
+def _load_evaluator(eval_dir: Path):
+    """Import the fetched MLCommons ``eval_accuracy`` module."""
+    if not (eval_dir / "eval_accuracy.py").exists():
+        raise FileNotFoundError(
+            f"eval_accuracy.py not found at {eval_dir}. Run setup_eval.sh first."
+        )
+    sys.path.insert(0, str(eval_dir))
```
arekay-nv, line 62: [Claude] **low** (design): `sys.path.insert(0, str(eval_dir))` front-shadows any module named `eval_accuracy` and is never removed. Low-impact in a short-lived subprocess, but `importlib.util.spec_from_file_location` would import the vendored file by path without mutating global `sys.path`.

src/inference_endpoint/dataset_manager/transforms.py
```diff
@@ -176,12 +176,27 @@ def __init__(
...
+        # prompts sent to /v1/completions) - in that case __call__ short-circuits
+        # and the Harmonizer is never needed.
+        self._tokenizer_name = tokenizer_name
+        self._encoding_name = encoding_name
+        self._reasoning_effort = reasoning_effort
+        self._conversation_start_date = conversation_start_date
+        self._harmonizer: Harmonizer | None = None
+
+    @property
+    def harmonizer(self) -> Harmonizer:
```
arekay-nv, line 191: [Claude] **low** (design): The `harmonizer` lazy `@property` isn't safe under concurrent first access. Safe today because `RowProcessor` uses serial `df.apply`, but the comment justifies laziness only by offline-tokenization, not thread-safety — a future parallel caller could build two `Harmonizer`s (wasteful + side effects). Worth a one-line note that it relies on serial `df.apply`.

arekay-nv (commented):

**10** issues across **4** files (Codex: 2, Claude: 9, both: 1). Skipped the vendored upstream `mlperf_eval/eval_accuracy.py` and the binary parquet. Dropped 1 low finding (stratified-subset overshoot) as it overlaps an existing comment on the same function. Every line was verified against the source at `ba82e568`.

### 🔴 Must Fix
| 1 | `scripts/Dockerfile.dev` | 61 | bug | **Both** | `cd .../deepseek_r1` → renamed dir; breaks dev image build |

### 🟡 Should Fix
| 2 | `scripts/Dockerfile.dev` | 55 (+19) | docs | Claude | stale `DeepSeekR1Scorer`/`deepseek_r1` path refs |
| 3 | `scoring.py` | 1940 | docs | Codex | warns to run `score_livecodebench.sh` (not in repo) |
| 4 | `scoring.py` | 1791 | testing | Claude | `_run_eval_subprocess` error branches untested |
| 5 | `scoring.py` | 1967 | testing | Claude | LCB count-divergence guard untested |

### 🔵 Consider
| 6 | `scoring.py` | 348 | error-handling | Claude | broad `except Exception` hides programming errors |
| 7 | `scoring.py` | 1924 | bug | Claude | `round(em/100*n)` assumes binary accuracy, unasserted |
| 8 | `deepseek_eval_runner.py` | 170 | data-integrity | Claude | `NaN` `tokens_per_sample` → invalid JSON (standalone) |
| 9 | `deepseek_eval_runner.py` | 62 | design | Claude | `sys.path.insert(0)` never removed; prefer `importlib` |
| 10 | `transforms.py` | 191 | design | Claude | lazy `harmonizer` property not concurrency-safe (note) |

viraatc (commented):

Reviewed by: **Claude** (+ adversary council signals) | Depth: **standard** | Codex run timed out on an unrelated env issue (MCP-server inheritance) — being re-run separately.

5 findings, deduped against existing review comments (no overlap with the assert/`trust_remote_code`/empty-df/macro/naming/version-pin/untested-branch/redundant-except comments already on this PR). A pre-existing `ColumnFilter` mutation bug in `transforms.py` was found but is out of scope for this PR and not posted.

| # | File | Line | Severity | Category |
|---|------|------|----------|----------|
| A | `evaluation/scoring.py` | 1885 | medium | security |
| B | `.../deepseek_eval_runner.py` | 120 | medium | robustness |
| C | `commands/benchmark/execute.py` | 1118 | medium | robustness |
| D | `evaluation/scoring.py` | 1693 | low | clean-code |
| E | `evaluation/scoring.py` | 1872 | low | bloat |

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1620,380 @@ def score(self) -> tuple[float | None, int]:
...
+        scratch = self.report_dir / "deepseek_eval"
+        scratch.mkdir(parents=True, exist_ok=True)
+        input_parquet = scratch / f"{self.dataset_name}_outputs.parquet"
+        out_json = scratch / f"{self.dataset_name}_results.json"
+        eval_df.to_parquet(input_parquet, index=False)
+
+        n_lcb = int((eval_df["dataset"].astype(str) == self.lcb_subset).sum())
+        use_container = self.lcb_websocket_url is not None and n_lcb > 0
+
+        if not use_container:
```
viraatc, line 1900: [Claude] **medium** (security): The `not use_container` branch grades **every** subset in-process via `_run_eval_subprocess(input_parquet, out_json)` — including `livecodebench`, which executes model-generated Python. The dedicated `LiveCodeBenchScorer` refuses this exact path unless `ALLOW_LCB_LOCAL_EVAL=true` (scoring.py:960), and this class's own constructor comment says the in-process executor "can't sandbox" LCB — yet this branch runs it ungated whenever no container is configured (or `n_lcb==0` false-negatives aside, whenever `lcb_websocket_url is None`). Gate the in-process `livecodebench` grading behind the same `ALLOW_LCB_LOCAL_EVAL` env var, or always route it as an external subset.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1620,380 @@ def score(self) -> tuple[float | None, int]:
...
+            report_dir=report_dir,
+            extractor=extractor,
+            ground_truth_column=ground_truth_column,
+        )
+        self.subset_column = subset_column
+        self.question_column = question_column
+        self.tokenizer_path = tokenizer_path
+        self.uv_executable = uv_executable
+        # LiveCodeBench executes untrusted code, which the in-process MLCommons
+        # executor can't sandbox. When a port is set, the livecodebench subset
```
viraatc, line 1706: [Claude] **low** (clean-code): This comment says *"If the socket is unreachable, score() falls back to grading all subsets in the subprocess (prior behavior)."* — but the container path does the opposite: on an unreachable socket it leaves `livecodebench` unscored and marks the run incomplete (it does **not** re-run the in-process executor). The `(prior behavior)` phrasing is also exactly the history-narration AGENTS.md forbids. Rewrite to describe current behavior: unreachable container ⇒ LCB left unscored, run marked incomplete.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1620,380 @@ def score(self) -> tuple[float | None, int]:
         per_dim_scores = self._extract_per_dim_scores(results)
         mean_score = float(np.mean(per_dim_scores))
         return mean_score, n_repeats
+
+
+class LegacyMLPerfDeepSeekR1Scorer(Scorer, scorer_id="legacy_mlperf_deepseek_r1"):
```
viraatc, line 1638: file is growing quite large, lets note it down for refactor.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -1658,3 +1620,380 @@ def score(self) -> tuple[float | None, int]:
...
+            str(out_json),
+            "--tokenizer",
+            self.tokenizer_path,
+        ]
+        if external_subsets:
+            cmd += ["--external-subsets", external_subsets]
+        log_path = self.report_dir / "deepseek_eval_subprocess.log"
+        try:
+            completed = subprocess.run(
+                cmd,
```
viraatc, line 1784: [Codex] **high** (bug): This `subprocess.run(cmd, ...)` runs `uv run --project <subproject>` with no `env=`, so it inherits the process environment. The dev image sets `ENV UV_PROJECT_ENVIRONMENT=/opt/venv` (scripts/Dockerfile.dev:16) and never resets it for runtime, and that variable (absolute path) overrides `--project`'s default `<subproject>/.venv`. Since `cmd` has no `--frozen`/`--no-sync`, at container runtime `uv run --project legacy_mlperf_deepseek_r1` would sync the evaluator's incompatible pins (`transformers==4.33.2`, `numpy<2`) into `/opt/venv` — which is the parent app env (installed by the `uv sync` at Dockerfile.dev:46/50) — corrupting it instead of using the `.venv` provisioned at build time (Dockerfile:75). This is on the primary DSR1 accuracy path: `score()` always calls `_run_eval_subprocess` (both the container and non-container branches).

The Dockerfile comment (lines 64-66) delegates this to the launcher — *"at run time the launcher must likewise point UV_PROJECT_ENVIRONMENT + DEEPSEEK_EVAL_PROJECT_PATH at this subproject"* — but nothing in the repo does so, and the scorer reads `DEEPSEEK_EVAL_PROJECT_PATH` (line 1730) yet never sets `UV_PROJECT_ENVIRONMENT`. Rather than rely on an undocumented global env var at launch (which could also perturb the parent app's own uv usage), set it per-subprocess here: pass `env={**os.environ, "UV_PROJECT_ENVIRONMENT": str(self.project_path / ".venv")}` to this `subprocess.run` call (there is only this one in the method — the `TimeoutExpired` block is its exception handler, not a second call).

viraatc (commented):
## Review Council — Codex pass (follow-up)

The Codex reviewer (which failed earlier on an environment issue — now fixed) surfaced 2 findings not covered by existing comments or the Claude pass above:

| # | File | Line | Severity | Category |
|---|------|------|----------|----------|
| P1 | `evaluation/scoring.py` | 1770 | high | bug |
| P2 | `.../legacy_mlperf_deepseek_r1/__init__.py` | 186 | medium | bug |

P1 is the highest-severity item in the review: it breaks the primary containerized accuracy path.

src/inference_endpoint/dataset_manager/predefined/legacy_mlperf_deepseek_r1/__init__.py
```diff
@@ -0,0 +1,192 @@
...
+
+    @staticmethod
+    def _stratified_subset(
+        df: pd.DataFrame, max_samples: int, seed: int
+    ) -> pd.DataFrame:
+        if df.empty:
+            return df
+        frac = max_samples / len(df)
+        parts = [
+            group.sample(
```
viraatc, line 186: [Codex] **medium** (bug): `_stratified_subset` can return **more** than `max_samples`. Each subset contributes `min(max(1, round(len(group) * frac)), len(group))` rows; the `max(1, ...)` floor forces ≥1 row per subset and `round()` can round several groups upward, so when `max_samples < n_subsets` (or with upward rounding) the concatenated frame exceeds `max_samples` — violating the documented maximum and issuing more benchmark requests than requested. Allocate an exact budget across groups, or trim the shuffled result to the cap (e.g. `.head(max_samples)`).

viraatc (approved):
lgtm! minor comments from claude ;

src/inference_endpoint/evaluation/scoring.py
viraatc: should we raise error if not in container?

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
```diff
@@ -0,0 +1,951 @@
...
+logging.basicConfig(level=logging.INFO,
+                    format='%(asctime)s - %(levelname)s - %(message)s')
+logger = logging.getLogger(__name__)
+
+
+# =============================================================================
+# MLPerf Log Accuracy Processing
+# =============================================================================
+
+def process_mlperf_log_accuracy(mlperf_log_file: Union[str, Path],
```
viraatc, line 47: nit: would make it easier to read if this was broken into smaller functions

src/inference_endpoint/evaluation/scoring.py
arekay-nv: The eval subproject is excluded from the wheel (conflicting heavy deps), so the source-relative default only resolves in a checkout (when running from a repo). $DEEPSEEK_EVAL_PROJECT_PATH lets wheel-installed / containerized users (e.g. Dockerfile.dev, or a cluster launcher) point at a synced subproject without patching installed source — resolution order is explicit ctor arg → env → source default.

src/inference_endpoint/evaluation/scoring.py
arekay-nv: Refactored.

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
```diff
@@ -0,0 +1,951 @@
...
+
+    # Try to detect by content
+    try:
+        # Try reading as JSON first
+        with open(file_path, 'r') as f:
+            first_char = f.read(1)
+            if first_char in ['[', '{']:
+                # Likely JSON
+                return "mlperf_json"
+    except BaseException:
```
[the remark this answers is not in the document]
arekay-nv: Vendored from https://raw.githubusercontent.com/mlcommons/inference/e59ce582f544edcc1b3f69a6c6f3ebc66eecb3d7/language/deepseek-r1/eval_accuracy.py

src/inference_endpoint/evaluation/scoring.py
arekay-nv: added tests in `test_deepseek_r1_scorer.py`

src/inference_endpoint/evaluation/scoring.py
arekay-nv: `test_lcb_count_divergence_marks_incomplete` addresses this.

src/inference_endpoint/dataset_manager/predefined/legacy_mlperf_deepseek_r1/__init__.py
arekay-nv: Done in e0533a7 — `_build_from_source` now raises `NotImplementedError`. The loader is parquet-only: it reads the bundled `examples/07_DeepSeekR1_Example/data/deepseek_r1_eval.parquet` (or an override via `$LEGACY_MLPERF_DEEPSEEK_R1_DATASET`), and raises an actionable error on a git-LFS pointer stub. Resolving.

src/inference_endpoint/evaluation/scoring.py
arekay-nv: Addressed in e0533a7 — the comment at `scoring.py:1949-1952` now pins the strictly-binary (0/100) per-sample assumption and names the fallback (sum raw per-sample counts from the runner) for any future MLCommons subset that emits fractional/partial credit.

src/inference_endpoint/evaluation/scoring.py
arekay-nv: Fixed in e0533a7 — the `not use_container` path now refuses to grade any LiveCodeBench rows in-process unless `ALLOW_LCB_LOCAL=1` is explicitly set (`raise RuntimeError` otherwise). Covered by `test_no_container_with_lcb_rows_refuses` and `test_allow_lcb_local_opt_in_grades_in_process`.

src/inference_endpoint/evaluation/scoring.py
arekay-nv: Fixed in e0533a7 — both subprocess calls now pass `env=_uv_subproject_env(project)`, which sets `UV_PROJECT_ENVIRONMENT` to the subproject's own `.venv` so the eval no longer inherits the image's `/opt/venv`. Covered by `test_subprocess_pins_subproject_venv`.

src/inference_endpoint/dataset_manager/predefined/legacy_mlperf_deepseek_r1/__init__.py
arekay-nv: Fixed in e0533a7 — `_stratified_subset` now concatenates the per-subset picks, shuffles, and truncates to the cap (`pool.sample(frac=1, random_state=seed).head(max_samples)`), so it can never return more than `max_samples`. Covered by `test_stratified_subset_never_overshoots`.

src/inference_endpoint/evaluation/scoring.py
arekay-nv: Fixed in e0533a7 — the comment at `scoring.py:1705-1711` now states the actual behavior: when the LCB socket is unreachable the subset is left UNSCORED and the run is marked incomplete (it is never silently graded in-process).

src/inference_endpoint/dataset_manager/transforms.py
arekay-nv: Addressed in e0533a7 — the `harmonizer` property comment now documents the concurrency contract explicitly: it relies on serial access (`RowProcessor` drives it via single-threaded `df.apply`) and is not safe for concurrent first build. Pins the invariant a future parallel caller must respect.

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
```diff
@@ -0,0 +1,951 @@
...
+            r"Answer:\s*(\d+)(?!\.)\b", text, re.IGNORECASE | re.MULTILINE))
+        if not answer_matches:
+            return None
+        extracted_str = answer_matches[-1].group(1)
+
+    try:
+        val = int(extracted_str)
+        if 0 <= val <= 999:
+            return val
+    except ValueError:
```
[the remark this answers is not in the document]
arekay-nv: This file is vendored **verbatim** from MLCommons (`mlcommons/inference@e59ce58`, `language/deepseek-r1/eval_accuracy.py`, Apache-2.0). We keep it byte-for-byte so it stays diffable against upstream and preserves provenance, so we intentionally don't modify it (empty-except / unused-locals / structure are upstream's). Resolving as won't-fix (vendored).

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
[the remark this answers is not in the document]
arekay-nv: This file is vendored **verbatim** from MLCommons (`mlcommons/inference@e59ce58`, `language/deepseek-r1/eval_accuracy.py`, Apache-2.0). We keep it byte-for-byte so it stays diffable against upstream and preserves provenance, so we intentionally don't modify it (empty-except / unused-locals / structure are upstream's). Resolving as won't-fix (vendored).

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
```diff
@@ -0,0 +1,951 @@
...
+
+    Args:
+        df_evaluated: DataFrame with evaluated results
+        logger: Optional logger instance (uses module logger if not provided)
+
+    Returns:
+        Dictionary with evaluation statistics
+    """
+    if logger is None:
+        logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
arekay-nv: This file is vendored **verbatim** from MLCommons (`mlcommons/inference@e59ce58`, `language/deepseek-r1/eval_accuracy.py`, Apache-2.0). We keep it byte-for-byte so it stays diffable against upstream and preserves provenance, so we intentionally don't modify it (empty-except / unused-locals / structure are upstream's). Resolving as won't-fix (vendored).

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
```diff
@@ -0,0 +1,951 @@
...
+
+    # Calculate statistics
+    evaluated = df_evaluated['extracted_answer'].notna().sum()
```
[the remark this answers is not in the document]
arekay-nv: This file is vendored **verbatim** from MLCommons (`mlcommons/inference@e59ce58`, `language/deepseek-r1/eval_accuracy.py`, Apache-2.0). We keep it byte-for-byte so it stays diffable against upstream and preserves provenance, so we intentionally don't modify it (empty-except / unused-locals / structure are upstream's). Resolving as won't-fix (vendored).

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
```diff
@@ -0,0 +1,951 @@
...
+    correct = (df_evaluated['prompt_accuracy'] > 0).sum()
```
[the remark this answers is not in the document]
arekay-nv: This file is vendored **verbatim** from MLCommons (`mlcommons/inference@e59ce58`, `language/deepseek-r1/eval_accuracy.py`, Apache-2.0). We keep it byte-for-byte so it stays diffable against upstream and preserves provenance, so we intentionally don't modify it (empty-except / unused-locals / structure are upstream's). Resolving as won't-fix (vendored).

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
arekay-nv: This file is vendored **verbatim** from MLCommons (`mlcommons/inference@e59ce58`, `language/deepseek-r1/eval_accuracy.py`, Apache-2.0). We keep it byte-for-byte so it stays diffable against upstream and preserves provenance, so we intentionally don't modify it (empty-except / unused-locals / structure are upstream's). Resolving as won't-fix (vendored).

src/inference_endpoint/evaluation/scoring.py
arekay-nv: Use case: it lets an operator point the scorer at a relocated copy of the eval subproject — e.g. baked into a container image at a non-default path, or a shared read-only mount — without editing code. It defaults to the in-repo eval subproject, so it's a no-op unless explicitly set. My recommendation is to keep it (cheap, image-friendly), but leaving this **open** for your call — happy to drop the override if you'd rather not expose it.

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/deepseek_eval_runner.py
arekay-nv: Deferring as a low-priority follow-up. This runs in a short-lived out-of-process runner (fresh interpreter per eval), so the `sys.path` mutation never outlives the process and there's no live collision. `importlib.util.spec_from_file_location` is the cleaner path-based import; leaving this **open** rather than churning the runner in this PR.

src/inference_endpoint/evaluation/scoring.py
arekay-nv: Agreed — `scoring.py` has grown large. Noting it for a follow-up refactor (splitting the DeepSeek-R1 / LiveCodeBench scorers out of the shared module). Leaving this **open** as the tracking note rather than resolving.

PR #379, opened 2026-06-27
feat: accuracy issuer inherits perf concurrency in online mode (#357)

arekay-nv:
When the performance phase runs an **online** load pattern, the accuracy phase now mirrors that same pattern instead of always bursting at MAX_THROUGHPUT, so evaluation exercises the endpoint the same way as the performance run:

- **CONCURRENCY** perf → accuracy inherits the same fixed `target_concurrency`.
- **POISSON** perf → accuracy inherits the same `target_qps`.

It falls back to MAX_THROUGHPUT only when there is no online pattern to mirror: offline (MAX_THROUGHPUT) perf, AGENTIC_INFERENCE perf (which can't drive the non-agentic accuracy datasets — `create_load_strategy` rejects it), or no perf pattern at all.

**Caveat:** inheriting POISSON rate-limits accuracy to the perf QPS (there is no separate accuracy QPS budget yet). This is intentional so accuracy is measured under the same offered load as performance; a dedicated accuracy load-pattern control is planned as a follow-up (see the thread on the agentic case).

Also logs the accuracy issuer's chosen load mode (pattern + parameter) per accuracy dataset, via a new `LoadPattern.__str__`. Adds unit tests for concurrency-inheritance, poisson-inheritance, offline→max_throughput, agentic→max_throughput, none→max_throughput, and the logging cases.

## What does this PR do?

Makes the online accuracy phase reuse the performance phase's load pattern (concurrency or poisson) rather than always running MAX_THROUGHPUT, so accuracy is measured under the same offered load as performance. Offline / agentic / unset perf patterns keep the MAX_THROUGHPUT fallback.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Related issues

Related: #357

## Testing

- [x] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [x] Pre-commit hooks pass


src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -482,6 +482,16 @@ def _build_phases(
         )
     )
 
+    # Accuracy mirrors the perf load pattern so evaluation exercises the
+    # endpoint the same way it was benchmarked. AGENTIC_INFERENCE can't drive
+    # the (non-agentic) accuracy datasets — create_load_strategy rejects it —
+    # so it (and a missing perf pattern) falls back to MAX_THROUGHPUT.
+    perf_lp = ctx.rt_settings.load_pattern
+    if perf_lp is None or perf_lp.type == LoadPatternType.AGENTIC_INFERENCE:
```
nvzhihanj, line 570: @hvagadia is this the right intended behavior for agentic workload? I might be missing something here but thought we should use the same load pattern as well

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -494,11 +504,19 @@ def _build_phases(
                 "AgenticInferenceDataset, which is not yet supported for "
                 "accuracy evaluation."
             )
-        # Accuracy phases run at MAX_THROUGHPUT; inheriting perf_lp (e.g. POISSON)
-        # would silently rate-limit evaluation until an agentic inference accuracy strategy
-        # and QPS-budgeting support are added.
-        acc_load_pattern: LoadPattern | None = LoadPattern(
-            type=LoadPatternType.MAX_THROUGHPUT
+        if acc_load_pattern.type == LoadPatternType.CONCURRENCY:
```
nvzhihanj: Seems like `LoadPatternType` can have a `__str__` (or/and `__repr__`) class so we can print directly instead of if statement here

src/inference_endpoint/commands/benchmark/execute.py
hvagadia: For our standalone accuracy. we will likely have to use much lower concurrency than performance due to docker overhead. @tianmu-li is working on the PR, he will likely have a separate field to control accuracy load pattern.

viraatc (approved):

PR #386, opened 2026-07-01
fix: dont discount open tcp-ports against max-connections on init

viraatc:
## Problem

This fixes false ephemeral-port budget failures when running the test suite in parallel with `pytest -n auto`.

During a large parallel test run, client setup failed with:

```text
--max-connections (10) exceeds the ephemeral port budget (0 = 0 ports x 1 distinct endpoint(s))
```

The client was counting every socket currently open on the machine. Sockets created by other tests made that count reach the system port-range size, so setup incorrectly concluded that no ports remained.

[#371](https://github.com/mlcommons/endpoints/pull/371) / [commit `72c28e0`](https://github.com/mlcommons/endpoints/commit/72c28e0d8d336623d247d9b1a9d8bb6b0535d7b5) correctly gave each distinct endpoint its own budget, but it still started from this changing machine-wide socket count. If that count reported zero available ports, endpoint scaling could not help.

## Fix

Use the configured OS ephemeral-port range as the stable overall budget instead of subtracting sockets that happen to be open during setup. Continue multiplying that budget by the number of distinct endpoints.

If the machine actually runs out of ports later, the connection attempt will report the real runtime error.

## Unchanged behavior

- `max_connections` remains one overall limit.
- `warmup_connections` remains one overall limit.
- Existing worker pool splitting is unchanged.
- CLI defaults and pytest configuration are unchanged.

## CI follow-up

The first PR run exposed an unrelated race in `test_terminate_procs_kills_running_workers`. The worker had exited via SIGTERM, but the test thread and `ProcessPoolExecutor` manager thread raced to reap it. The losing `waitpid()` call saw `ECHILD`, so `Process.is_alive()` briefly returned `True` even though the failure output already showed `exitcode=-SIGTERM`.

[Commit `b084d6b7`](https://github.com/mlcommons/endpoints/commit/b084d6b729530e33ae187fe2fe25c54902b0a972) makes the test observe the process sentinel instead of racing process return-code bookkeeping. It also uses the production `spawn` context, a stdlib-only warm worker probe, and direct cleanup on failure. Production shutdown behavior is unchanged.

## Validation

- Regression test reproduced the exact zero-budget error before the fix and passes afterward.
- HTTP client config tests: `19 passed`.
- Metrics aggregator token tests: `50 passed`.
- Unit-marked suite: `946 passed, 5 skipped`.
- Full non-slow suite passes with `pytest -q -n auto -x -m "not slow and not performance"`.
- Reported warmup and preexisting-dead-worker integration cases passed.
- All pre-commit checks passed.
- GitHub Actions `test`, schema, build, audit, pre-commit, CodeQL, and CLA checks pass.

The separately marked slow 32-worker test still reaches its existing 60-second startup timeout. This PR does not change that separate behavior.

### DIFF
--- a/src/inference_endpoint/endpoint_client/config.py
+++ b/src/inference_endpoint/endpoint_client/config.py
@@ -44,7 +44,7 @@ from .cpu_affinity import (
     get_cpus_in_numa_node,
     get_current_numa_node,
 )
-from .utils import get_ephemeral_port_limit, get_ephemeral_port_range
+from .utils import get_ephemeral_port_range
 
 
 def _endpoint_destination(url: str) -> tuple[str | None, int]:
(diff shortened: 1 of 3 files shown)


arekay-nv (approved):
Thanks for catching and the quick fix!

PR #398, opened 2026-07-06
fix: use cwd for reported sha 

arekay-nv:
## What does this PR do?
The SHA reported in the reports was not setting the cwd correctly, which led to incorrect SHA being reported. If the source repo was different from the endpoints repo, the SHA of the source repo was being used. 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass


nvzhihanj (approved):

src/inference_endpoint/utils/version.py
```diff
@@ -17,30 +17,62 @@
...
     Returns:
-        The short git SHA (7 chars) or None if not in a git repo or git unavailable.
+        The short git SHA (at least 7 chars; git lengthens it if a 7-char
+        prefix is ambiguous), or None if the package is not in a git checkout
+        (e.g. an installed wheel) or git is unavailable.
     """
     try:
         result = subprocess.run(
-            ["git", "rev-parse", "--short=7", "HEAD"],
+            ["git", "rev-parse", "--show-toplevel", "--short=7", "HEAD"],
```
nvzhihanj, line 55: Will this work if we explicitly install into an image?

If not we might also need a version tag (something like 1.0.1) to gatekeep it

arekay-nv: No, an image will not have the git info - we need to add scm support. Will address in followup.

Issue #384, opened 2026-06-30
Refactor commands/benchmark/execute.py — split the benchmark God-module


viraatc:
## Problem

`src/inference_endpoint/commands/benchmark/execute.py` has grown to **~1270 lines** and concentrates most of the run lifecycle's responsibilities in one module. `_run_benchmark_async` alone is **~370 lines** (L663–L1030) and interleaves unrelated concerns, which makes it hard to read, test in isolation, and extend without touching the hot path.

Distinct responsibilities currently living in this one file:

- Setup & context (`setup_benchmark`, `BenchmarkContext`, `_default_report_path`)
- Tokenizer/accuracy resolution (`_check_tokenizer_exists`, `_resolve_accuracy_components`)
- Dataset loading (`_load_datasets`)
- Phase construction (`_build_phases`)
- Profiling triggers (`_derive_profile_urls`, `_post_profile`, `_render_profile_status`, `_write_profiling_section`)
- Power monitoring (`_build_power_payload`) — added in #383
- Snapshot/IO (`_load_final_snapshot_from_disk`, `_salvage_tmpfs`, `_write_scoring_artifacts`)
- Async run orchestration (`_run_benchmark_async`, `run_benchmark_async`)
- Finalization/reporting (`finalize_benchmark`, `run_benchmark`)

## Why now

#383 (power monitoring) had to thread start/stop hooks and a payload builder into an already-large `_run_benchmark_async` + `finalize_benchmark`. Profiling did the same earlier. Each optional sidecar feature (profiling, power, and future ones) adds more interleaved branches to the same long functions. This is a recurring smell — the module wants an extension seam.

## Suggested direction (non-prescriptive)

- Extract the optional "around-the-run" features (profiling, power, future telemetry) behind a small uniform **run-hook seam** (e.g. `on_run_start` / `on_phase_start` / `on_run_stop` / `contribute_to_finalize`) so `_run_benchmark_async` stops growing per feature.
- Move profiling helpers to a `profiling/`-adjacent module and power helpers already live in `power/` — `execute.py` should orchestrate, not implement, each sidecar.
- Split finalization (scoring artifacts, sibling-JSON writers, report.txt sections) into a `finalize`/reporting module.
- Break `_run_benchmark_async` into named stages (service launch → warmup → measure → drain → collect-results).

[...]

PR #400, opened 2026-07-08
feat(metrics): consolidate gpt-oss/deepseek accuracy, dedup reporting, surface TPS

arekay-nv:
## Summary

Reworks accuracy reporting into a uniform **per-dataset list** and aligns the run's report directory with the MLPerf endpoints submission layout ([mlcommons/inference#2628](https://github.com/mlcommons/inference/issues/2628)).

> Note: an earlier revision of this PR added a *composite* gpt-oss consolidation scorer; that was **removed** in favor of the per-dataset list below.

### Per-dataset accuracy list
- `Report.accuracy` and `results.json`'s `accuracy_scores` are now an **always-present list** (empty for perf-only), **one entry per accuracy dataset**. Each entry: `dataset_name`, `extractor`, `ground_truth_column`, `score`, `unit_samples`, `num_repeats`, `total_samples` (= unit × repeats), `duration_s` (the dataset's issue-phase wall-clock), `complete`, and — for multi-subset scorers (DeepSeek-R1, BFCL) — a BFCL-shaped `breakdown`.
- The single breakdown reader `find_accuracy_breakdown` iterates the list; `compliance/checker.py` + `metrics/results_plots.py` are unchanged behind it.
- Rendered per-dataset in `report.txt` / `results_summary.json`: `  <name>: <score> (unit=U, repeats=R, total=T, duration=Ds)`.

### Removed the composite gpt-oss path
- Deleted the composite `gptoss_120b_accuracy` predefined dataset, `GptOss120bAccuracyScorer`, and `_weighted_breakdown`; the per-subset `aime25::gptoss` / `gpqa::gptoss` / `livecodebench::gptoss` datasets are the sole path. DeepSeek-R1 keeps its combined-subset breakdown as one list entry.

### Report directory conforms to #2628
```
<report_dir>/
  config.yaml
  performance/results_summary.json   # the perf-metrics Report
  accuracy/accuracy_results.json     # the accuracy list
  audit/audit_<test_id>.json         # e.g. audit_output_caching_test.json (+ verify_<TEST>.txt)
  results.json                       # full run: config + perf + accuracy_scores + responses + errors
  report.txt · events.jsonl · metrics/ · sample_idx_map.json
```
- `scripts/publish_submission.py` still copies the artifact trio **flat** into the MLPerf `performance/run_1/` submission layout.

### Fixes surfaced during validation
- Coerce numpy scalar scores to native `float`: the base `Scorer.score()` returned `np.mean(...)` (`numpy.float64`), which crashed `Report.to_json` (msgspec) and `results.json` (json) serialization — caught on a live gpt-oss run.
- Updated `scripts/publish_submission.py` and `evaluation/bfcl_v4_multi_turn_cli.py` to the list shape.

### End-to-end validation (GB200 NVL72)
- **gpt-oss-120b** (trtllm-serve): aime25 **0.808** (8×) / gpqa **0.723** (5×) / livecodebench **0.849** (3×) @ 32768 OSL.
- **DeepSeek-R1-NVFP4** (trtllm-serve DEP4): overall **81.04** on the full 4388-sample eval (math500 95.8 / aime 83.3 / lcb 87.7 / mmlu_pro 77.3 / gpqa 66.7) — matches the MLPerf golden (81.36).


src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1206,41 +1192,76 @@ def _salvage_tmpfs(report_dir: Path, tmpfs_dir: Path) -> None:
...
+                subset=cfg.dataset_name.split("::")[0],
+                scorer_cls=cfg.scorer,
+                extractor=cfg.extractor,
+                dataset=cfg.dataset,
+                ground_truth_column=cfg.ground_truth_column,
+                extras=cfg.extras,
+            )
+            for cfg in members
+        ]
+        scorer = GptOssAccuracyScorer(gpt_members, members[0].report_dir)
```
nvzhihanj: Seems hard coded?

arekay-nv: Yes, iterating on consolidating and verifying. Will need some iterations before it is ready. Let me mark it draft.

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -2032,4 +2036,164 @@ def score(self) -> tuple[float | None, int]:
...
+        """Per-subset accuracy breakdown cached by :meth:`score` (BFCL-shaped).
+
+        ``None`` until :meth:`score` runs, or if scoring produced no usable
+        result.
+        """
+        return self._breakdown
+
+
+@dataclass
+class GptOssMember:
```
nvzhihanj: This accuracy bundle is not specific to GPT-OSS and we can reuse for other model as well. Having the model name in the dataclass is not ideal

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -2032,4 +2036,164 @@ def score(self) -> tuple[float | None, int]:
...
+    full_name: str
+    subset: str
+    scorer_cls: type[Scorer]
+    extractor: type[Extractor] | None
+    dataset: Dataset
+    ground_truth_column: str | None
+    extras: dict[str, Any]
+
+
+class GptOssAccuracyScorer:
```
nvzhihanj: I would assume that we will have many different scorer and we can mix and match them for models as needed (instead of hard coding scorer for each of the models)?

tests/unit/scripts/test_publish_submission.py
```diff
@@ -64,6 +65,42 @@ def test_non_dict_metric_returns_none(self):
         assert ps._percentile(None, "99.0") is None
 
 
+class TestVerifyAccuracy:
+    def test_empty_scores_reports_missing(self, tmp_path):
+        ps = _load_publish_submission()
+        run = tmp_path / "acc_run"
+        run.mkdir()
+        (run / "results.json").write_text("{}", encoding="utf-8")
```
nvzhihanj: Leaving a note to change to `accuracy/accuracy_results.json` format

nvzhihanj:
Thanks for the change. Can you help add some examples of the accuracy report (e.g. for gpt-oss) after the change?

arekay-nv (commented):
## Review Council — thorough (Claude)

Codex did not complete this run (its process crashed mid-review), so this is a Claude-only pass. 5 findings below, all verified against HEAD.

src/inference_endpoint/evaluation/accuracy_results.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Thin wrapper over :func:`find_accuracy_entry` for consumers that only need the
+    breakdown block (e.g. the compliance gate, which reads BFCL's
+    ``overall_accuracy`` / ``normalized_single_turn_score`` from it).
+    """
+    entry = find_accuracy_entry(results)
+    return entry.get("breakdown") if entry is not None else None
+
+
+def average_accuracy(accuracy_scores: list[dict[str, Any]]) -> float | None:
```
arekay-nv, line 89: [Claude] **medium** · data-integrity: `average_accuracy` is a plain mean of the per-dataset scalar scores with no scale/metric-homogeneity guard (only `"performance"` and non-numeric scores are excluded). With the new `accuracy_config.scale`, a run that scales some datasets (e.g. gpt-oss `scale:100` → 0–100) but leaves another at the default `1.0` (a pass@1 / ROUGE dataset → 0–1) produces a silently meaningless headline (e.g. mean of 80.0 and 0.3 = 40.15), surfaced as `average_accuracy` in `results.json`/`accuracy_results.json` and the `Average:` line. The shipped examples are homogeneous so it's a latent footgun, not a broken example — consider validating consistent scale, or documenting the homogeneity assumption.

src/inference_endpoint/metrics/report.py
```diff
@@ -99,6 +115,36 @@ def _series_to_metric_dict(stat: dict[str, Any]) -> dict[str, Any]:
...
+    performance one. Returns ``{}`` when ``values`` is empty.
+    """
+    # sig_figs / n_histogram_buckets mirror the aggregator's OSL registration
+    # defaults (MetricsRegistry.register_series, and the --hdr-sig-figs /
+    # --n-histogram-buckets argparse defaults of 3 / 30) so the accuracy block
+    # matches the perf one. Only build_stat(exact=True) is used below, so the HDR
+    # bounds feed an otherwise-unused view; n_histogram_buckets, however, IS read
+    # by the exact histogram path — keep it aligned if the aggregator's is ever
+    # made configurable.
+    sampler = SeriesSampler(
```
arekay-nv: [Claude] **low** · design: `series_metric_dict` reuses the aggregator's private internals (`SeriesSampler`, `_metric_to_dict`, `_DEFAULT_PERCENTILES`) and hardcodes `sig_figs=3`/`n_histogram_buckets=30` + duplicated `_TOKEN_HDR_LOW/HIGH` to match the perf OSL. The 'byte-for-byte same shape' guarantee holds only because the aggregator subprocess is never passed `--hdr-sig-figs`/`--n-histogram-buckets` (so it uses the same 3/30 defaults). If those are ever wired through config, the accuracy OSL keeps 30 buckets while perf uses the configured count — a silent divergence the code comment flags but doesn't prevent. A shared constants source (or exposing these from the aggregator) would remove the drift risk + the private-symbol coupling.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1281,38 +1345,142 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
-        assert eval_cfg.dataset.data is not None
-        num_samples = len(eval_cfg.dataset.data)
+        # Coerce a numpy scalar score (e.g. np.mean from the base Scorer) to a
+        # native Python float so the entry stays serializable by both msgspec
+        # (result_summary.json) and json (results.json). numpy.float64 is a
+        # float subclass, so isinstance(..., float) catches it while leaving
+        # None / dict (RougeScorer) untouched; float(...) drops the numpy type.
+        # accuracy_config.scale then applies (e.g. 100 to report a [0,1] pass@1
+        # score as a percentage); the breakdown, if any, keeps its own scale.
+        if isinstance(score, float):
```
arekay-nv: [Claude] **low** · error-handling: The numpy→native coercion only catches `float` subclasses: `if isinstance(score, float): score = float(score) * scale`. `np.float64` is a float subclass (and is what the base `Scorer.score()` emits today), but `np.float32`/`np.integer` are **not** — a custom scorer returning one would pass through uncoerced, then `json.dump(results)` (:1557) raises and the broad `except … logger.error("Save failed")` (:1578) silently drops **both** `results.json` and `accuracy/accuracy_results.json` while the run still exits 0. Consider coercing via `np.floating`/`np.integer`, and/or not swallowing a serialization failure of the whole artifact set.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1281,38 +1345,142 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
             entry["breakdown"] = breakdown
-        accuracy_scores[eval_cfg.dataset_name] = entry
+
+        # Attach avg/min/max output-token length. Skipped for the perf entry (its
+        # OSL is in result_summary.json) and when no tokenizer is configured. A
+        # tokenization/read failure only drops OSL — it never fails scoring.
+        if osl_tokenizer is not None and eval_cfg.dataset_name != "performance":
+            try:
+                if uuid_to_text is None:
+                    out_df = scorer_instance.get_outputs()
```
arekay-nv: [Claude] **low** · performance: `uuid_to_text` is built from `scorer_instance.get_outputs()`, which returns **every** phase's COMPLETE events (incl. the perf phase), so in `--mode both` it retains the entire run's response-text corpus for the whole finalize even though `_phase_osl_stats` only reads the accuracy datasets' uuids. `del out_df` avoids holding the DataFrame + dict together, but the whole-corpus dict is still kept. Filtering it to the union of the accuracy `sample_index_map` keys would bound memory to the accuracy population. (Finalize-side, so footprint-only.)

src/inference_endpoint/evaluation/accuracy_results.py
```diff
@@ -0,0 +1,137 @@
...
+    complete: bool = True,
+) -> dict[str, Any]:
+    """Build a per-subset breakdown dict.
+
+    ``subset_scores`` values are percentages in ``[0, 100]``. The overall/headline
+    accuracy is intentionally *not* stored here — it lives on the accuracy entry's
+    scalar ``score`` (see :func:`find_accuracy_entry`), so this block carries only
+    the per-subset detail the entry can't.
+
+    ``total_samples`` semantics are producer-defined and not directly comparable
```
arekay-nv, line 133: [Claude] **low** · documentation: `build_breakdown`'s docstring references a producer that no longer exists: *'the gpt-oss roll-up uses the summed unique problem count'*. `build_breakdown` has exactly one caller now — `LegacyMLPerfDeepSeekR1Scorer._cache_breakdown` — and the gpt-oss composite scorer/dataset was removed in this PR (gpt-oss now runs as 3 separate single-metric datasets that emit no breakdown). Drop the gpt-oss clause so it doesn't mislead a future reader into thinking gpt-oss flows through this constructor.

arekay-nv:

Reviewed by: **Claude** (Codex did not complete — its process crashed mid-review) | Depth: **thorough**

Found **5** issues across 3 files (2 new/real, 3 latent/known). All verified against HEAD `ad97ad9`.

### 🟡 Should Fix (medium)

| 1 | `evaluation/accuracy_results.py` | 93 | data-integrity | `average_accuracy` takes a plain mean with no scale-homogeneity guard — mixing `scale:100` and default-`1.0` datasets yields a silently meaningless headline. |

### 🔵 Consider (low)

| 2 | `metrics/report.py` | 134 | design | `series_metric_dict` couples to aggregator privates + duplicates `3/30`/HDR constants; drifts if `--hdr-sig-figs`/`--n-histogram-buckets` are ever wired through. |
| 3 | `commands/benchmark/execute.py` | 1355 | error-handling | numpy coercion only catches `float` subclasses; `np.float32`/`np.integer` slip through → `json.dump` raises → the broad `except` at :1578 silently drops all artifacts (exit 0). |
| 4 | `commands/benchmark/execute.py` | 1395 | performance | `uuid_to_text` retains the full perf+accuracy corpus at finalize; only accuracy uuids are read. |
| 5 | `evaluation/accuracy_results.py` | 128 | documentation | `build_breakdown` docstring references the removed gpt-oss roll-up (only DeepSeek calls it now). |

No critical/high issues — the refactor is sound. #1 is the one worth acting on; #2 and #4 were noted in earlier internal passes and partially mitigated (`del out_df`, a drift-warning comment); #3 and #5 are quick fixes.

> ⚠️ **Commit hygiene**: this PR has 17 commits including ~3 apparent fixups — consider squashing before merge.

examples/04_GPTOSS120B_Example/vllm_gptoss_120b_per_dataset_osl_example.yaml
```diff
@@ -61,6 +63,7 @@ datasets:
       eval_method: "code_bench_scorer"
       extractor: "python_code_extractor"
       num_repeats: 3
+      scale: 100
```
arekay-nv: Remove the scale option.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1231,40 +1237,98 @@ def _salvage_tmpfs(report_dir: Path, tmpfs_dir: Path) -> None:
...
 
-def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
-    """Score accuracy, aggregate results, write JSON."""
-    config = ctx.config
-    result = bench.session
-    collector = bench.collector
-    report = bench.report
+# Bound the memory of one batched tokenization pass over accuracy outputs, which
+# can be up to tens of thousands of tokens each (e.g. gpt-oss lcb at 32768).
+_OSL_TOKENIZE_BATCH = 256
```
arekay-nv: fold into arg.

scripts/publish_submission.py
```diff
@@ -135,7 +141,7 @@ def _percentile(metric: dict[str, Any], key: str) -> Any:
 def _verify_performance(run_dir: Path) -> list[str]:
     """Return human-readable findings for the performance artifacts."""
     findings: list[str] = []
-    summary = _load_json(run_dir / "result_summary.json")
+    summary = _load_json(run_dir / "performance" / "result_summary.json")
     results = _load_json(run_dir / "results.json")
```
nvzhihanj: Why do we still need the results.json? What extra information do we need in that file that is not covered by other reports (I assume we could removed this one)

arekay-nv: This is the submission path - once our internal reporting gets merged we can look at the submission checker side to see if we even need this script and remove it. I would prefer to keep this compared to the txt file as the json format would enable better testing. We can keep it summarized for human readability. 
Can address in followup.

nvzhihanj: As discussed offline, let's remove results.json and decide whether to keep report.txt later

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1231,40 +1234,125 @@ def _salvage_tmpfs(report_dir: Path, tmpfs_dir: Path) -> None:
...
+    """Load the run's tokenizer for accuracy-phase OSL, or ``None`` to skip it.
+
+    OSL is a reporting nicety, so a missing or unloadable tokenizer disables it
+    rather than failing scoring — mirroring the aggregator's no-tokenizer ⇒ no
+    OSL behavior on the perf side.
+    """
+    if name is None:
+        return None
+    try:
+        tokenizer = AutoTokenizer.from_pretrained(name)
```
nvzhihanj: This may be an issue - in benchmarking we lock the model to a specific commit but load by name will by default load the latest tokenizer. (and also there will be HF login issue for semi-public model)

I would recommend loading from local tokenizer model and then optionally from remote HF. @viraatc to review as well

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1231,40 +1234,125 @@ def _salvage_tmpfs(report_dir: Path, tmpfs_dir: Path) -> None:
...
+    except Exception as e:  # noqa: BLE001 - OSL is optional; never fail scoring
+        logger.warning(
+            "Accuracy OSL disabled: could not load tokenizer %r: %s", name, e
+        )
+        return None
+    # Counting tokens on long outputs would otherwise spam the
+    # "sequence longer than model_max_length" warning; raise the cap so it can't.
+    tokenizer.model_max_length = int(1e12)
```
nvzhihanj: This seems like a weird hack - why would we tokenize sequence longer than the max context window?

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1231,40 +1234,125 @@ def _salvage_tmpfs(report_dir: Path, tmpfs_dir: Path) -> None:
...
-            if bench.profiling is not None:
-                _write_profiling_section(f, bench.profiling)
-        logger.info("Report written to %s", report_txt)
 
-    # Sibling profiling.json — kept separate so Report stays a pure
-    # snapshot-derived struct.
-    if bench.profiling is not None:
-        (ctx.report_dir / "profiling.json").write_text(
-            json.dumps(bench.profiling, indent=2)
+def _score_accuracy(
```
nvzhihanj, line 1349: execute.py has bloated to multi thousand lines - should we move accuracy related code to eval folder?

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1231,40 +1234,125 @@ def _salvage_tmpfs(report_dir: Path, tmpfs_dir: Path) -> None:
...
-    # Write scoring artifacts + copy event log from tmpfs to disk
-    _write_scoring_artifacts(ctx, result, bench.tmpfs_dir)
+    # Accuracy-phase output-token lengths (finalize-side, off the hot path): the
+    # aggregator only tokenizes perf-window samples, so re-tokenize the accuracy
+    # responses (already in events.jsonl) here. Tokenizer loaded once; the
+    # uuid→text map is read once from the first scorer's get_outputs (it returns
+    # every phase's COMPLETE events) and reused across datasets. Loaded only when
+    # a real accuracy dataset exists — a perf-only run (or one with just the
+    # inline "performance" entry) must not pay a tokenizer load/download here.
+    has_accuracy = any(ec.dataset_name != "performance" for ec in ctx.eval_configs)
```
nvzhihanj: Why use != perf instead of == accuracy?

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1281,38 +1369,155 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
             entry["breakdown"] = breakdown
-        accuracy_scores[eval_cfg.dataset_name] = entry
+
+        # Response accounting + avg/min/max output-token length. Skipped for the
+        # perf entry (its OSL / failure counts live in result_summary.json). The
+        # counts are computed independent of the tokenizer and of OSL returning a
+        # block — an all-failed phase must still publish scored=0 rather than
+        # silently omitting everything. OSL stays tokenizer-gated. A read/tokenize
+        # failure only drops these blocks — it never fails scoring.
+        if eval_cfg.dataset_name != "performance":
```
nvzhihanj: Why are we using dataset_name not the `type`?

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1281,38 +1369,155 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
+                    )
+                    # Drop the DataFrame so finalize doesn't hold both it and the
+                    # dict (each carrying the full response-text corpus).
+                    del out_df
+                entry["response_counts"] = _phase_response_counts(
+                    scorer_instance.sample_index_map, uuid_to_text
+                )
+                if osl_tokenizer is not None:
+                    t0 = time.perf_counter()
+                    osl = _phase_osl_stats(
```
nvzhihanj, line 1499: Are we repeating the tokenization here (as in metric aggregation) or it's moving the tokenization here?

src/inference_endpoint/evaluation/accuracy_results.py
```diff
@@ -0,0 +1,137 @@
...
+
+from __future__ import annotations
+
+from typing import Any
+
+# Ruleset golden-metric name -> key in the scorer's breakdown block.
+ACCURACY_METRIC_KEYS = {
+    "bfcl_overall_accuracy": "overall_accuracy",
+    "bfcl_normalized_accuracy": "normalized_single_turn_score",
+}
```
nvzhihanj: This seems like quite a hack and will be hard to maintain

(At the very least it shouldn't live in this file)

src/inference_endpoint/metrics/report.py
```diff
@@ -19,16 +19,32 @@
...
 from inference_endpoint.utils.version import get_version_info
 
 from ..utils import monotime_to_datetime
 
+# HDR bounds for a token-count series — mirror the aggregator's OSL/ISL series
+# (metrics_aggregator.aggregator._TOKEN_HDR_LOW/HIGH). Only the exact (raw-array)
+# path is used below, so these feed the otherwise-unused HDR view; kept identical
+# so an accuracy OSL block is constructed the same way as the perf one.
+_TOKEN_HDR_LOW = 1
+_TOKEN_HDR_HIGH = 10_000_000
```
nvzhihanj: I couldn't understand what this is doing just by reading the comment, can you clarify?

nvzhihanj:
## Review Council — architecture review (Concern-driven)

Reviewed by **Claude + Codex** (independent), thorough. Two maintainer concerns; both **confirmed** by both reviewers. Proposing the concrete change for each.

### 🏗️ Concern 1 — `execute.py` is doing too much (~1656 lines)

The accuracy path was grown inside the orchestrator. **Concrete decomposition** (both reviewers agree; removes ~350 lines from `execute.py`):

- **Accuracy scoring/OSL** — `_score_accuracy` (~140 lines), `_phase_osl_stats`, `_phase_response_counts`, `_load_osl_tokenizer`, `_write_scoring_artifacts` → **new `evaluation/accuracy_scoring.py`** (sibling of the new `accuracy_results.py`), taking a small typed input (eval_configs + tokenizer_name + phase timings) instead of `BenchmarkContext`. `finalize_benchmark` becomes a one-line `score = accuracy_scoring.score(...)`.
- **Profiling** — `_derive_profile_urls` / `_post_profile` / `_render_profile_status` / `_write_profiling_section` + `_PROFILE_PATHS` (~95 self-contained lines, zero `BenchmarkContext` coupling) → **new `commands/benchmark/profiling.py`** (a `ProfilerController`).
- Dataset loading (`_load_datasets`, `_resolve_accuracy_components`) → `commands/benchmark/datasets.py` (lower priority).

_Good:_ the new `accuracy_results.py` actually **reduces** duplication (dedups helpers previously copy-pasted in `checker.py` + `results_plots.py`). But `report.py` now imports aggregator-**internal privates** (`SeriesSampler`, `_DEFAULT_PERCENTILES`, `_metric_to_dict`) — leaky cross-layer coupling (see the `report.py:46` thread).

### ⚙️ Concern 2 — accuracy OSL re-tokenizes on the main process (duplicated tokenizer)

Correct that it's a **second `AutoTokenizer` in the main process** + a different implementation from the aggregator's sharded `BatchTokenizer`. It is *not* duplicate work on the same samples (the aggregator only tokenizes the perf window — `is_tracking` is off for accuracy phases), but it is a real finalize-time CPU cost, and it re-tokenizes **scorer-transformed** text (BFCL's `get_outputs()` serializes merged tool-calls → wrong OSL).

**Concrete change (the right architecture):** the OpenAI/sglang responses already carry `usage.completion_tokens`/`prompt_tokens`, but it's dropped in `accumulator.get_final_output` / `session.py` before it reaches `events.jsonl`. **Persist server usage into `QueryResult`/the event log → OSL/ISL become a direct read that eliminates _both_ client re-tokenizers**, with the aggregator's `BatchTokenizer` as the single shared fallback. **Caveat (both reviewers):** MLPerf OSL wants the _reference_ tokenizer for determinism, and some servers omit `usage` (or need `stream_options.include_usage`), so the shared reference-tokenizer fallback stays **mandatory** — "delete all client tokenization" is too strong. The key point: never a bespoke main-thread `AutoTokenizer`; one shared path.

### 🔎 Additional findings (posted inline)
- `scoring.py:224` — base `Scorer.score()` excludes truly-missing samples from the denominator → inflated accuracy; all-missing → `NaN` written as non-standard JSON.
- `accuracy_results.py:93` — `average_accuracy` is a plain mean of heterogeneous scorer scales (0–1 vs 0–100).
- `execute.py:1377` — `isinstance(score, float)` misses `np.float32`.
- `tests/unit/compliance/test_checker.py:203` — fixture false-confidence (`overall_accuracy` that the real DeepSeek breakdown omits).

Concrete change suggestions are also left inline on the existing open threads (`execute.py:1247/1255/1323/1352/1417/1438`, `accuracy_results.py:43`, `report.py:46`, `publish_submission.py:145`).

src/inference_endpoint/evaluation/scoring.py
```diff
@@ -214,7 +218,10 @@ def score(self) -> tuple[float | None, int]:
             scores.append(self.score_single_sample(empirical[i], ground_truths[i]))
 
         n_repeats = len(scores) // self.dataset.num_samples()
-        return np.mean(scores), n_repeats
+        # float(...) so callers get a native Python float, not a numpy scalar:
+        # the score flows into result_summary.json (msgspec) and results.json
+        # (json), neither of which can serialize numpy.float64.
+        return float(np.mean(scores)), n_repeats
```
nvzhihanj: [Review Council] **medium** · data-integrity — base `Scorer.score()` scores only rows that produced a COMPLETE event.

`df = df[df["sample_uuid"].isin(valid_uuids)]` then `n_repeats = len(scores)//num_samples()` and `float(np.mean(scores))`. A failed request that still logs a COMPLETE-with-empty-output is counted as 0 (fine), but a sample with **no** COMPLETE event (drain-timeout / crash) is **excluded from the denominator** → accuracy is inflated over the surviving subset. And if *all* target rows are missing, `np.mean([])` → `NaN`, which is then serialized as non-standard `NaN` JSON. **Fix:** use `sample_index_map` size as the denominator (left-join outputs by uuid; missing → failure), and reject non-finite scores before writing artifacts (or return `None` / set `complete=False`).

src/inference_endpoint/evaluation/accuracy_results.py
nvzhihanj, line 89: [Review Council] **low** · design — `average_accuracy` takes a plain mean of per-dataset `score`, but scorer scales are not homogeneous: pass@1 / LiveCodeBench return 0–1 while the DeepSeek and BFCL scorers return 0–100. Today's shipped configs happen to be homogeneous (gpt-oss = three 0–1 scorers) so no live miscalculation, but a config mixing a 0–1 and a 0–100 dataset yields a silently meaningless mean. **Fix:** normalize/assert a common scale, or document the homogeneity precondition on the function.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1281,38 +1369,155 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
             ) from e
         score, n_repeats = scorer_instance.score()
-        assert eval_cfg.dataset.data is not None
-        num_samples = len(eval_cfg.dataset.data)
+        # Coerce a numpy scalar score (e.g. np.mean from the base Scorer) to a
+        # native Python float so the entry stays serializable by both msgspec
+        # (result_summary.json) and json (results.json). numpy.float64 is a
+        # float subclass, so isinstance(..., float) catches it while leaving
+        # None / dict (RougeScorer) untouched; float(...) drops the numpy type.
+        if isinstance(score, float):
```
nvzhihanj: [Review Council] **low** · error-handling — `if isinstance(score, float): score = float(score)` only coerces `float` subclasses. `np.float64` is one (works), but `np.float32` is **not** — a scorer returning `np.float32` passes through uncoerced and crashes `msgspec`/`json` encoding. Since every in-tree scorer already returns a native `float` (`float(np.mean(...))`), this guard is dead for the case it cites and incomplete for the case that matters. **Fix:** `isinstance(score, numbers.Real) and not isinstance(score, bool)`, or drop it as redundant.

tests/unit/compliance/test_checker.py
```diff
@@ -198,6 +200,33 @@ def test_accuracy_gate_fails_on_too_few_samples():
     assert not samples.passed
 
 
+@pytest.mark.unit
```
nvzhihanj, line 207: [Review Council] **low** · testing — `test_accuracy_gate_accepts_non_bfcl_breakdown` puts `"overall_accuracy": 83.0` in the breakdown, but the real DeepSeek path (`build_breakdown`) **omits** `overall_accuracy` (asserted elsewhere), and `check_accuracy` reads `overall_accuracy` only from the breakdown block with no fallback to the scalar `entry['score']` (unlike `extract_accuracy`, which does fall back). So a genuine DeepSeek breakdown gated on `bfcl_overall_accuracy` would fail 'metric missing' — the test gives false confidence. **Fix:** drop `overall_accuracy` from the fixture and add the entry-score fallback to `check_accuracy` for parity with `extract_accuracy`.

nvzhihanj (commented):
## Review Council — net-new findings

Inline findings not already covered by the open threads (architecture proposal is in the top-level summary). Reviewed by Claude + Codex.

src/inference_endpoint/commands/benchmark/execute.py
nvzhihanj: Confirmed — two issues here. `AutoTokenizer.from_pretrained(name)` pins **no `revision`**, so it loads the *latest* tokenizer, not the one at the benchmarked model commit (reproducibility hole you flagged). It also omits `trust_remote_code=True`, which the aggregator (`token_metrics.py`) requires — so DeepSeek/custom tokenizers silently fail to load and accuracy OSL is quietly disabled. **Concrete change:** don't load a tokenizer here at all — persist the server's `usage.completion_tokens` (see the :1438 thread) and read OSL directly. If a client fallback is kept, it must reuse the aggregator's `BatchTokenizer` construction (same `trust_remote_code`, same pinned revision), not a second `AutoTokenizer`.

src/inference_endpoint/commands/benchmark/execute.py
nvzhihanj: It suppresses the `sequence longer than model_max_length` warning while *counting* tokens on long outputs (not generating) — a symptom of re-tokenizing on the client. **Concrete change:** if OSL comes from the server's `usage.completion_tokens` (or the shared aggregator tokenizer, which already avoids this), the `model_max_length = 1e12` hack disappears entirely — drop the whole `_load_osl_tokenizer` path.

src/inference_endpoint/commands/benchmark/execute.py
nvzhihanj: Agreed. **Concrete decomposition** (both Claude and Codex proposed this independently — ~350 lines out of `execute.py`): extract the accuracy cluster (`_score_accuracy`, `_phase_osl_stats`, `_phase_response_counts`, `_load_osl_tokenizer`, `_write_scoring_artifacts`) into a new `evaluation/accuracy_scoring.py` next to `accuracy_results.py`, taking a small typed input (eval_configs + tokenizer_name + phase timings) rather than `BenchmarkContext`; `finalize_benchmark` then just calls `score_accuracy(...)`. Separately, the self-contained profiling subsystem (`_derive_profile_urls`/`_post_profile`/`_render_profile_status`/`_write_profiling_section`, ~95 lines) → a new `commands/benchmark/profiling.py` (`ProfilerController`). That leaves `execute.py` a thin lifecycle driver.

src/inference_endpoint/commands/benchmark/execute.py
nvzhihanj: Minor but yes — string-comparing to `perf`/`performance` is fragile (a dataset literally named that is misclassified). **Concrete change:** branch on `eval_cfg.dataset.type` (or the phase `type`), not the string name. Same root cause as the :1417 thread.

src/inference_endpoint/commands/benchmark/execute.py
nvzhihanj: Same root as :1352 — `if eval_cfg.dataset_name != 'performance'` compares the *name*. It's also carrying a subtle correctness dependency (the comment notes `get_outputs()` returns *all* phases), which makes it doubly fragile. **Concrete change:** key on dataset `type`/phase type, and restrict `uuid_to_text` to the accuracy datasets' `sample_index_map` keys so finalize doesn't hold the entire *perf* response-text corpus in a main-process dict.

src/inference_endpoint/commands/benchmark/execute.py
nvzhihanj: It's re-tokenizing — a **second `AutoTokenizer` in the main process**, a different implementation from the aggregator's sharded `BatchTokenizer`. It's not duplicate work on the *same* samples (the aggregator only tokenizes the perf window), but it *is* duplicated logic + a heavy second load, and it re-tokenizes **scorer-transformed** text — BFCL's `get_outputs()` serializes merged tool-calls, so its OSL is wrong. **Concrete change (right architecture):** the response already carries `usage.completion_tokens`/`prompt_tokens`, but it's dropped in `accumulator.get_final_output` / `session.py` before `events.jsonl`. Persist server usage into `QueryResult`/the event log → OSL/ISL become a direct read that eliminates *both* client tokenizers; keep the aggregator's `BatchTokenizer` as the single shared fallback (servers that omit usage / streaming without `include_usage`). Caveat: MLPerf OSL wants the *reference* tokenizer, so the fallback stays mandatory — but it should be one shared path, never a bespoke main-thread `AutoTokenizer`.

src/inference_endpoint/evaluation/accuracy_results.py
nvzhihanj: Agreed it's fragile. Concretely, `find_accuracy_entry` (:60) only matches entries that carry `breakdown.subset_scores`, so DeepSeek-style headline accuracy — which lives on the scalar `entry['score']`, not a breakdown — is **invisible** to every consumer using this helper (the checker gate, the plots). **Concrete change:** replace the breakdown-shape sniffing with metric-specific selectors that can read `entry['score']`, `average_accuracy`, and named breakdown metrics by dataset/scorer — and move that selection logic out of a place where it 'shouldn't live'.

src/inference_endpoint/metrics/report.py
nvzhihanj: This is `series_metric_dict`: it rebuilds an OSL block by importing aggregator-**internal privates** — `SeriesSampler`, `_DEFAULT_PERCENTILES` from `metrics_aggregator/registry.py` and `_metric_to_dict` from `snapshot.py` — and hardcodes `sig_figs=3, n_histogram_buckets=30` to mirror the aggregator's argparse defaults. **Concrete change:** `report.py` (a snapshot *consumer*) shouldn't import producer-side aggregation primitives; promote a small shared stats builder to a neutral module both import (e.g. `metrics/series_summary.py`), or expose a public builder from the aggregator. Otherwise it silently drifts if the aggregator changes its HDR bounds/sig-figs/buckets.

scripts/publish_submission.py
nvzhihanj: With `result_summary.json` now self-complete (qps/tps + distributions) and accuracy in `accuracy/accuracy_results.json`, `results.json` is largely redundant for our own tooling — but MLCommons `run_verification.py` / submission checkers still read it, so it can't be dropped until the submission side migrates. **Concrete change:** keep `results.json` as a compat-only artifact *generated from* the consolidated reports (single source of truth), and document it as such. Separately, `_verify_accuracy` (:186) silently returns nothing if handed an old dict-shaped `accuracy_scores` — for a submission verifier, emit an explicit 'unsupported legacy shape' finding instead of silent emptiness.

src/inference_endpoint/commands/benchmark/execute.py
nvzhihanj: We probably cannot use the API completion_tokens?

arekay-nv: Yes, the other option was to do hot-path tokenization with the batch tokenizer, so the alternative was to do post-processing tokenization for the accuracy without adding impact. Let me see if there are other options.

src/inference_endpoint/commands/benchmark/execute.py
arekay-nv: Should i address this here or in followup as tracked in #384

src/inference_endpoint/metrics/results_plots.py
```diff
@@ -94,39 +98,24 @@ class RunArtifacts:
...
-                return block
-    return None
-
-
 def extract_accuracy(results: dict[str, Any]) -> AccuracyBreakdown | None:
-    """Pull the accuracy breakdown from a BFCL ``results.json`` dict."""
-    score = _find_accuracy_score(results)
-    if score is None:
+    """Pull the accuracy breakdown from a BFCL ``accuracy_results.json`` dict."""
+    entry = _find_accuracy_entry(results)
```
viraatc, line 147: **gpt-oss accuracy plots render empty.** After the composite scorer was removed, the three per-subset gpt-oss entries (`aime25::gptoss`, `gpqa::gptoss`, `livecodebench::gptoss`) carry no `breakdown` — `PassAt1Scorer`/`LiveCodeBenchScorer` don't override `score_breakdown()`. `_find_accuracy_entry` only matches entries with `subset_scores`, so it returns `None` here and `extract_accuracy` bails *before* the scalar-score fallback below can run. Net: gpt-oss runs get no accuracy plot.

**Fix:** when no entry has a `breakdown`, build plot inputs from each entry's scalar `score` (one bar per dataset), or attach a `build_breakdown` for the three components at finalize.

src/inference_endpoint/evaluation/accuracy_results.py
```diff
@@ -0,0 +1,137 @@
...
+    result equals the single dataset's score when there is only one. The inline
+    ``"performance"`` entry — a scored perf dataset, not an accuracy component —
+    and any non-numeric score are excluded. Returns ``None`` when no component has
+    a numeric score.
+    """
+    values = [
+        float(entry["score"])
+        for entry in accuracy_scores
+        if isinstance(entry, dict)
+        and entry.get("dataset_name") != "performance"
```
viraatc: Two issues in `average_accuracy`:

1. **Mixed scales.** Headline `score` is a 0–1 fraction for gpt-oss/BFCL but 0–100 for DeepSeek-R1 (0.808 vs 81.04 in this PR's own numbers). A plain mean across scorer families is meaningless, and the `report.txt` `Average:` line reads inconsistently between runs. Normalize to one unit (fraction or %) before averaging/displaying.
2. **Name-based exclusion defeats the new `dataset_type`.** This PR added `AccuracyConfiguration.dataset_type` *specifically* so a dataset named `performance` isn't magic — but the serialized entry drops it, so this still filters on `dataset_name != 'performance'`. Persist `dataset_type` (or `is_performance: bool`) on the entry and filter on that.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1281,32 +1423,147 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
-        assert eval_cfg.dataset.data is not None
-        num_samples = len(eval_cfg.dataset.data)
-        if eval_cfg.dataset_name == "performance":
-            num_samples = sum(phase.issued_count for phase in result.perf_results)
+        # Coerce a numpy scalar score (e.g. np.mean from the base Scorer) to a
+        # native Python float so the entry stays serializable by both msgspec
+        # (result_summary.json) and json (results.json). numpy.float64 is a
+        # float subclass, so isinstance(..., float) catches it while leaving
+        # None / dict (RougeScorer) untouched; float(...) drops the numpy type.
+        if isinstance(score, float):
```
viraatc: **Coercion is dtype-narrow.** `isinstance(score, float)` catches `np.float64` (a `float` subclass) but not `np.float32` or numpy ints. If any scorer returns one, it leaks into the `json.dump` for `accuracy_results.json` → `TypeError` → swallowed by the broad `except` → the accuracy artifact is silently not written and the run still exits 0. No current scorer hits this (all `float(np.mean(...))`), so it's latent.

**Fix:** coerce with `float(score)` guarded by `isinstance(score, numbers.Real)` (covers every numpy scalar), or `np.asarray(score).item()`.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1342,34 +1599,37 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
+            # for at least one dataset — gating on the key's presence, not the
+            # rounded wall-clock, so a sub-millisecond total (tiny outputs) still
+            # records 0.0 rather than silently dropping the field.
+            osl_computed = any("osl_tokenize_s" in e for e in accuracy_scores)
+            osl_tokenization_s = round(
+                sum(e.get("osl_tokenize_s", 0.0) for e in accuracy_scores), 3
+            )
+            accuracy_dir = ctx.report_dir / "accuracy"
+            accuracy_dir.mkdir(parents=True, exist_ok=True)
+            accuracy_results_path = accuracy_dir / "accuracy_results.json"
```
viraatc: Three things at this write site:

1. **Stale-artifact leakage on report-dir reuse.** `performance/` is always written but `accuracy/` only when scored. The old single `results.json` was overwritten every run; now a perf-only rerun into a dir that previously held an accuracy run leaves a stale `accuracy/accuracy_results.json` (and vice-versa). Moot if report dirs are always fresh/timestamped — worth a glance.
2. **`results.json` is gone, but `collector.responses` is still collected (ACC/BOTH mode) and now never serialized.** Dead work, and the PR description's dir tree still lists `results.json # … responses + errors`. Either drop the collector path or fix the description (response text lives in `events.jsonl`, so no data loss).
3. **Non-atomic write.** Audit artifacts use atomic `tmp+fsync+rename` (`compliance/result.py`); this and `result_summary.json` use plain `open()/write`, so a crash mid-write leaves truncated JSON. Consider the same atomic writer.

viraatc (commented):
## Review Council — local pass (Codex + Grok + manual verification)

Scope: integration + the report-layout / accuracy-list migration. **The core migration is sound** — I verified the BFCL and DeepSeek-R1 compliance/breakdown paths and that the list-shape consumers (`checker.py`, `results_plots.py`, `publish_submission.py`, `bfcl_v4_multi_turn_cli.py`) are all migrated. Most issues below are user-facing artifacts the diff didn't touch. (BFCL/dataset coupling inside the accuracy module is intentionally out of scope — tracked separately.)

Inline comments: gpt-oss plots empty · `average_accuracy` scale + `dataset_type` · numpy coercion · the accuracy-write site (stale artifacts / dead `collector.responses` / non-atomic).

### 🔴 Not inline (files unchanged by this PR) — broken examples
The layout change breaks committed examples users follow verbatim:
- `examples/11_Edge_Agentic_Example/README.md:244,252` — reads `results.json` + dict `accuracy_scores['bfcl_v4::function_calling']`; it's now `accuracy/accuracy_results.json` + a **list** → `FileNotFoundError` / `TypeError`.
- `examples/03_BenchmarkComparison/compare_with_vllm.py:352` — reads `report_dir / 'result_summary.json'`; it's now under `performance/`.

**Fix:** update both to the new paths + list shape (mirror `bfcl_v4_multi_turn_cli.py:229-247`).

### Notes (low)
- Accuracy-write failure is swallowed (`except Exception: logger.error`) → exit-0 with no accuracy artifact; consider hard-failing when scoring produced entries.
- `find_accuracy_breakdown` returns the *first* `subset_scores` entry — fine for BFCL-only / DSR1-only; a hypothetical mixed run could grab the wrong block by ordering.

_Gemini errored out; Opus didn't return in time — this is Codex + Grok + manual verification._

tests/unit/compliance/test_checker.py
arekay-nv: Fixed in 82531a4. The fixture now uses a realistic DeepSeek-R1-style breakdown — no `overall_accuracy` in the block (the headline lives on the entry's scalar `score`, a percentage) — and `check_accuracy` falls back to the entry `score` for the `overall_accuracy` metric when the block omits it, matching the plot path. Also added a bounded-fallback test (`test_accuracy_gate_reports_missing_when_no_overall_anywhere`) so the gate reports "metric missing" rather than silently passing when neither the block nor the entry carries the overall.

src/inference_endpoint/commands/benchmark/execute.py
arekay-nv: It's a separate post-run pass, not duplicated hot-path work. `_phase_osl_stats` tokenizes the accuracy datasets' final outputs at `finalize` time (off the hot path) to produce exact per-dataset OSL for the accuracy report, reusing the metrics aggregator's shared reference tokenizer (`load_reference_backend` / `encode_lengths` in `token_metrics.py`) rather than a bespoke tokenizer or unreliable server `completion_tokens`. So the tokenizer/loader logic is shared (consistent counts), and this tokenization runs once in `finalize`, guarded by try/except and timed via `osl_tokenize_s` — no impact on the live run.

src/inference_endpoint/evaluation/accuracy_results.py
arekay-nv: Documented the homogeneity precondition on `average_accuracy` rather than guessing scales. Real runs score one model's same-family datasets (gpt-oss: three fraction `[0, 1]` scorers; DeepSeek-R1: one percentage `[0, 100]` scorer), so the components share a scale. We deliberately did **not** add a magnitude-based guard — it false-omits valid low-percentage runs (a near-zero percentage is indistinguishable from a fraction by magnitude alone). Homogenizing the scorer `score()` contract to a single unit is the real fix and is tracked separately.

src/inference_endpoint/commands/benchmark/execute.py
arekay-nv: Agreed — the accuracy cluster (`_score_accuracy` plus the OSL / response-count / uuid-bound helpers) has grown large enough to belong under `evaluation/`. Deferring to a dedicated follow-up PR to keep this one scoped to the accuracy-reporting consolidation and avoid tangling a large mechanical move with the fixes here. Leaving this thread open to track the follow-up.

viraatc (approved):
LGTM, thanks!

PR #330, opened 2026-06-02
Add configurable options to drain_in_flight timeout in different phases

Victor49152:
## What does this PR do?

1. Add DrainConfig that accepts three parameters: drain timeout for warmup/perf/acc phase
2. Update the yaml template and unit tests
3. Update drain timeout settings in vlm so that offline scenario could finish properly

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [x] Tests added/updated
- [ ] All tests pass locally
- [ ] Manual testing completed

## Checklist

- [x] Code follows project style
- [x] Pre-commit hooks pass
- [ ] Documentation updated (if needed)

PR #402, opened 2026-07-08
test: cover per-phase drain timeout in BenchmarkSession

wu6u3tw:
## Summary

The original version of this PR made per-phase in-flight drain timeouts configurable. That feature has since landed on main via #330 and #337 (`settings.drain` / `DrainConfig`, `PhaseConfig.drain_timeout`, unbounded accuracy drain), so this PR is reduced to the one piece main still lacks: a session-level regression test.

- `test_drain_respects_per_phase_timeout`: a phase whose in-flight samples never complete must exit the drain wait after `drain_timeout` seconds instead of hanging.

## Test plan

- `uv run pytest tests/unit/load_generator/test_async_session.py -k drain` (3 passed)
- `pre-commit run` on the changed file (ruff, mypy, license header all pass)

### DIFF
--- a/tests/unit/load_generator/test_async_session.py
+++ b/tests/unit/load_generator/test_async_session.py
@@ -447,6 +447,30 @@ class TestBenchmarkSession:
         assert session._current_phase_stopped is True
 
     @pytest.mark.asyncio
+    async def test_drain_respects_per_phase_timeout(self, caplog):
+        """PhaseConfig.drain_timeout bounds the drain wait for stuck samples."""
+        loop = asyncio.get_running_loop()
+        publisher = FakePublisher()
+
+        issuer = FakeIssuer()
+        issuer._loop = loop
+        issuer._auto_respond = False  # samples issue but never complete
+
+        session = BenchmarkSession(issuer, publisher, loop)
+        phases = [
+            PhaseConfig(
+                "perf",
+                _make_settings(n_samples=2),
+                FakeDataset(2),
+                drain_timeout=0.1,
+            ),
+        ]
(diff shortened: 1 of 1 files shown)


wu6u3tw:
Applied both review suggestions: drain log lines now format the timeout via a shared timeout_str (no %.0f on a possibly-None float, and sub-second values like 0.1 s print exactly), and the test asserts the precise "Drain timed out after 0.1 s" message. Also trimmed the inline comments/docstrings. 101 session+config tests pass.

src/inference_endpoint/config/schema.py
```diff
@@ -651,149 +484,30 @@ class WarmupConfig(BaseModel):
...
     model_config = ConfigDict(extra="forbid", frozen=True)
 
-    warmup_timeout_s: Annotated[
-        float | None,
-        cyclopts.Parameter(
-            alias="--warmup-drain-timeout",
-            help="Warmup drain timeout in seconds (None = wait indefinitely)",
-        ),
-    ] = Field(
+    warmup_timeout_s: float | None = Field(
```
nvzhihanj: I would in general avoid adding more fine-granularity flag to control timeout at each phase, because it adds burden to the whole system and makes config even more difficult to understand/write.

@arekay-nv @viraatc do you think we should just set an arbitrary large timeout and fail out if the timeout is hit? The silent timeout has caused us lots of trouble in tuning because it will pretend successful run while some samples are actually dropped.

wu6u3tw:
Closed because it is covered in #330/#337

PR #403, opened 2026-07-08
feat: run wan2.2 VBench accuracy from the prebuilt image — bake scorer in + fix cold-start failures

wu6u3tw:
Make the wan2.2 VBench accuracy scorer both **shippable** and **runnable** from the prebuilt endpoints image. Two parts:

1. **Package it** — bake the VBench accuracy project into the image (it lives under `examples/`, so the stock `Dockerfile.dev` never included it).
2. **Fix it** — three independent cold-start failures in the scoring path, each surfacing only after a full generation pass (30-50 min on 18 GB300 nodes).

With both, a full wan2.2 Offline accuracy run completed on GB300-NVL72 and scored **0.69998 over 248 samples**. Supersedes #414 (packaging split), which is now closed. Split from #401 (part 2 of 2; the drain-timeout config is in the companion PR).

## Packaging: bake the scorer into the image (`scripts/Dockerfile.dev`)

Adds a `PROVISION_VBENCH` build arg (**default `1`**, consistent with the existing `PROVISION_DSR1=1`) that COPYs `examples/09_Wan22_VideoGen_Example/accuracy` to `/opt/vbench_accuracy` and builds its **isolated** `.venv` (own uv-managed Python 3.11; `UV_PROJECT_ENVIRONMENT` overridden per call so its old/heavy deps never leak into the main `/opt/venv`). The pyproject is patched to relax `requires-python` and swap `decord`->`decord2` for wheel-only installs on both **x86_64 and aarch64/GB300**. Also installs `libgl1`/`libglib2.0-0` (opencv). Build with `--build-arg PROVISION_VBENCH=0` for a lean image without wan2.2 accuracy. Accuracy runs then reuse the baked venv via `UV_NO_SYNC=1` (`vbench_project_path=/opt/vbench_accuracy`) — no host checkout.

## Fix 1: stage videos as `.mp4` regardless of source suffix

`VBenchScorer._stage_videos` kept the source extension (`src.suffix or '.mp4'`). VBench's `load_video` dispatches purely on the extension and raises bare `NotImplementedError` for anything but `.mp4`/`.gif`/frame dirs. trtllm-serve emits MJPEG `.avi` when ffmpeg is not installed server-side, so a full accuracy pass generated all 248 videos and then died at scoring. decord (libav) detects the container by content, not extension; an MJPEG-AVI symlinked under an `.mp4` name decodes correctly (verified on GB300: 81 frames, 720x1280). Staged symlinks are now always named `{prompt}-{idx}.mp4`.

## Fix 2: `vbench_runner.py` fails on torch >= 2.6 checkpoint loading

The wan2.2 accuracy subproject resolves torch 2.12, where `torch.load` defaults to `weights_only=True` (changed in 2.6). VBench's reference checkpoints (motion_smoothness AMT, RAFT) are full pickles and fail with `UnpicklingError`. The runner sets torch's official escape-hatch env var `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1` (via `os.environ.setdefault`) at the top of the scorer subprocess — no monkeypatching of `torch.load`. It is read at `torch.load` call time and only overrides the default (callsites passing `weights_only` explicitly are unaffected), so the effect is scoped to this subprocess and the parent benchmark process keeps stock semantics.

## Fix 3: opaque failure when `wget`/`unzip` are missing

`vbench.utils.init_submodules` downloads per-dimension weights via literal `wget`/`unzip` subprocesses on a cold cache. In a minimal client container this dies mid-evaluation with `FileNotFoundError: 'wget'`, after the videos were already generated. The runner now preflights both tools and exits early with a structured error naming what is missing and how to fix it. (If you rely on the cold-cache download, provide `wget`/`unzip` or pre-stage the weights; otherwise bake/mount the weights.)

## Testing

- New `test_stage_videos_renames_non_mp4_sources_to_mp4`.
- `tests/unit/evaluation/test_scoring.py` and the full unit suite pass with zero new failures vs the unmodified base commit.
- Hardware validation on GB300-NVL72: VBench scored 248/248 staged videos end to end (score 0.69998), including cold-cache weight downloads.
- Image build: `docker build -f scripts/Dockerfile.dev --build-arg PROVISION_VBENCH=1 .` (and `PROVISION_VBENCH=0` leaves the stock image behavior for everything but the two apt packages).

### DIFF
--- a/examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
+++ b/examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
@@ -28,6 +28,7 @@ Exit codes:
 
 import argparse
 import json
+import os
 import sys
 import traceback
 from importlib.resources import files as _pkg_files
@@ -35,6 +36,17 @@ from importlib.resources import files as _pkg_files
 import torch
 import vbench as _vbench_pkg
 from vbench import VBench
+
+# VBench checkpoints are full pickles; this env var makes torch.load default to
+# weights_only=False in this subprocess. Read at call time, not import time.
+os.environ.setdefault("TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD", "1")
+
+# torch.hub validates repos via an unauthenticated GitHub API call that hits
+# HTTP 403 rate limits on shared cluster nodes. The repos are trusted; skip it.
+try:
+    torch.hub._validate_not_a_forked_repo = lambda *a, **k: None
+except AttributeError:  # pragma: no cover
+    pass
 
 
 def _emit_error(exc: BaseException) -> None:
@@ -105,9 +117,7 @@ def main() -> int:
 
     device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
     dimension_list = [d.strip() for d in args.dims.split(",") if d.strip()]
-    # VBench's `full_info_dir` arg is actually a JSON *file* path (despite
-    # the name). vbench_standard mode (used here) requires a valid file;
-    # passing None crashes inside VBench's load_json().
+    # full_info_dir is actually a JSON file path despite the misleading name.
     full_info_json = args.full_info_json or str(
         _pkg_files(_vbench_pkg).joinpath("VBench_full_info.json")
     )
--- a/scripts/Dockerfile.dev
+++ b/scripts/Dockerfile.dev
@@ -2,6 +2,11 @@
 # From project root:
 # docker build -f scripts/Dockerfile.dev --build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g) -t inference-endpoint-dev .
 # docker run -v $(pwd):/mnt/inference-endpoint -it --shm-size=512m inference-endpoint-dev bash
+#
+# VBench (WAN 2.2) accuracy scorer is ON by default (consistent with the DeepSeek-R1
+# evaluator, PROVISION_DSR1=1). It lives under examples/, not src/, so it is COPYed
+# in explicitly. To skip it (leaner image, no wan22 accuracy), build with PROVISION_VBENCH=0:
+# docker build -f scripts/Dockerfile.dev --build-arg PROVISION_VBENCH=0 --build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g) -t inference-endpoint-dev .
 
 FROM python:3.12.11-slim
 
@@ -15,12 +20,11 @@ ENV PYTHONUNBUFFERED=1 \
     UV_LINK_MODE=copy \
     UV_PROJECT_ENVIRONMENT=/opt/venv
 
-# Install system dependencies
-# git + curl + ca-certificates support the legacy_mlperf_deepseek_r1 evaluator's setup_eval.sh
-# (it git-clones prm800k / LiveCodeBench at pinned commits) in the provisioning RUN below;
-# without them that step fails and mlperf_eval/ is never baked. (eval_accuracy.py is vendored.)
+# libgl1 + libglib2.0-0: opencv runtime deps for the VBench scorer.
+# git + curl + ca-certificates: DSR1 setup_eval.sh clones prm800k / LiveCodeBench.
 RUN apt-get update && \
     apt-get install -y --no-install-recommends build-essential procps git curl ca-certificates \
+        libgl1 libglib2.0-0 \
     && rm -rf /var/lib/apt/lists/*
 
 RUN mkdir /mnt/inference-endpoint
@@ -38,9 +42,14 @@ RUN if ! getent group ${GROUP_ID}; then \
     fi && \
     useradd -u ${USER_ID} -g ${GROUP_ID} --create-home --shell /bin/bash appuser --no-log-init && \
     chown -R ${USER_ID}:${GROUP_ID} /mnt/inference-endpoint && \
-    mkdir -p /opt/venv && chown ${USER_ID}:${GROUP_ID} /opt/venv
+    mkdir -p /opt/venv /opt/uv-python /opt/vbench_accuracy /opt/vbench_cache && \
+    chown ${USER_ID}:${GROUP_ID} /opt/venv /opt/uv-python /opt/vbench_accuracy /opt/vbench_cache
 USER appuser
-ENV PATH="/opt/venv/bin:/home/appuser/.local/bin:$PATH"
+ENV PATH="/opt/venv/bin:/home/appuser/.local/bin:$PATH" \
+    UV_PYTHON_INSTALL_DIR=/opt/uv-python \
+    UV_NO_SYNC=1 \
+    VBENCH_PROJECT_PATH=/opt/vbench_accuracy \
+    VBENCH_CACHE_DIR=/opt/vbench_cache
 
 # Install dependencies first (cached unless pyproject.toml/uv.lock change)
 RUN uv sync --frozen --no-install-project --extra dev --extra test
@@ -76,4 +85,20 @@ RUN if [ "${PROVISION_DSR1}" = "1" ]; then \
         UV_PROJECT_ENVIRONMENT="$(pwd)/.venv" bash setup_eval.sh ; \
     else \
         echo "PROVISION_DSR1=${PROVISION_DSR1}: skipping DeepSeek-R1 evaluator provisioning" ; \
+    fi
+
+# VBench (WAN 2.2) accuracy scorer. Isolated venv at /opt/vbench_accuracy/.venv
+# (Python 3.11; tokenizers 0.13.3 has no cp312 wheel). prefetch_weights.py
+# pre-downloads the three dims that init_submodules would otherwise wget/unzip
+# at evaluation time. PROVISION_VBENCH=0 skips this block.
+ARG PROVISION_VBENCH=1
+COPY --chown=${USER_ID}:${GROUP_ID} examples/09_Wan22_VideoGen_Example/accuracy /opt/vbench_accuracy
+RUN if [ "${PROVISION_VBENCH}" = "1" ]; then \
+        cd /opt/vbench_accuracy && \
+        UV_PROJECT_ENVIRONMENT="$(pwd)/.venv" uv sync --frozen --python 3.11 && \
+        UV_PROJECT_ENVIRONMENT="$(pwd)/.venv" uv run python prefetch_weights.py && \
+        UV_PROJECT_ENVIRONMENT="$(pwd)/.venv" uv run python -c \
+            "import torch, decord, tokenizers, vbench; print('vbench', vbench.__file__, '| decord', decord.__version__, '| tokenizers', tokenizers.__version__, '| torch', torch.__version__)" ; \
+    else \
+        echo "PROVISION_VBENCH=${PROVISION_VBENCH}: skipping VBench provisioning" ; \
     fi
--- a/examples/09_Wan22_VideoGen_Example/accuracy/prefetch_weights.py
+++ b/examples/09_Wan22_VideoGen_Example/accuracy/prefetch_weights.py
@@ -1,0 +1,106 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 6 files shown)


examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
```diff
@@ -37,6 +39,25 @@
...
+    weights_only=True default. Explicit weights_only callers are unaffected.
+    """
+    orig_load = torch.load
+
+    @functools.wraps(orig_load)
+    def _load(*args, **kwargs):
+        kwargs.setdefault("weights_only", False)
+        return orig_load(*args, **kwargs)
+
+    torch.load = _load
```
nvzhihanj: Monkey patch in general should be avoided unless unnecessary (it will impact other workloads and make system hard to debug)

examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
```diff
@@ -37,6 +39,31 @@
...
+
+
+_default_torch_load_to_full_pickles()
```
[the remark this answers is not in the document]
wu6u3tw: Fixed in c0d006d.

examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
wu6u3tw: Fixed in c0d006d.

scripts/Dockerfile.dev
```diff
@@ -77,3 +86,24 @@ RUN if [ "${PROVISION_DSR1}" = "1" ]; then \
...
+
+# VBench (WAN 2.2) accuracy scorer, gated by PROVISION_VBENCH (default 1, like PROVISION_DSR1).
+# It lives under examples/ (not src/), so COPY it in and build its own .venv at
+# /opt/vbench_accuracy; accuracy runs reuse it via UV_NO_SYNC=1. Isolated like DSR1: pinned to
+# Python 3.11 (tokenizers 0.13.3 has no cp312 wheel) with UV_PROJECT_ENVIRONMENT set per call so
+# its old deps stay out of /opt/venv. pyproject patches (requires-python, decord->decord2) keep
+# the install wheel-only on aarch64. PROVISION_VBENCH=0 skips it.
+ARG PROVISION_VBENCH=1
+ENV UV_PYTHON_INSTALL_DIR=/opt/uv-python
+COPY --chown=${USER_ID}:${GROUP_ID} examples/09_Wan22_VideoGen_Example/accuracy /opt/vbench_accuracy
```
arekay-nv, line 95: **high** — flagged by Codex, Grok, Claude

The subproject is copied here to `/opt/vbench_accuracy`, but `VBenchScorer` resolves its default project path to `<repo>/examples/09_Wan22_VideoGen_Example/accuracy` (`scoring.py:1348`) and nothing sets `VBENCH_PROJECT_PATH`. That default path does not exist in the image, so `--mode acc` fails at scorer init (`vbench_runner.py` not found) and the baked venv is never used. Fix: `ENV VBENCH_PROJECT_PATH=/opt/vbench_accuracy` (and confirm the scorer reads it).

_Verified locally (built this PR's `Dockerfile.dev` with `PROVISION_VBENCH=0`): the scorer's own `_resolve_project_path(None)` resolves to `/mnt/inference-endpoint/examples/09_Wan22_VideoGen_Example/accuracy` — `runner.exists()=False` — and raises `FileNotFoundError: vbench_runner.py not found at …`. Setting `VBENCH_PROJECT_PATH=/opt/vbench_accuracy` resolves it (`runner.exists()=True`)._

scripts/Dockerfile.dev
```diff
@@ -19,8 +24,11 @@ ENV PYTHONUNBUFFERED=1 \
 # git + curl + ca-certificates support the legacy_mlperf_deepseek_r1 evaluator's setup_eval.sh
 # (it git-clones prm800k / LiveCodeBench at pinned commits) in the provisioning RUN below;
 # without them that step fails and mlperf_eval/ is never baked. (eval_accuracy.py is vendored.)
+# libgl1 + libglib2.0-0: opencv (cv2), pulled in by the VBench accuracy scorer's video
+# reader, needs libGL.so.1 / libgthread at import time (only used when PROVISION_VBENCH=1).
 RUN apt-get update && \
     apt-get install -y --no-install-recommends build-essential procps git curl ca-certificates \
+        libgl1 libglib2.0-0 \
```
arekay-nv, line 27: **high** — flagged by Codex, Grok, Claude

`vbench_runner.py` hard-fails (`return 2`, `MissingSystemDependency`, lines 98/117) when `wget` or `unzip` is missing on a cold VBench cache, but this apt layer installs only `libgl1 libglib2.0-0`. `python:3.12-slim` ships neither tool, so a default `PROVISION_VBENCH=1` image fails on the first accuracy run. Add `wget unzip` to the install list.

_Verified locally (same image): `command -v wget` and `command -v unzip` are both MISSING, and the runner's `shutil.which(("wget","unzip"))` guard returns exit code 2._

scripts/Dockerfile.dev
```diff
@@ -77,3 +86,24 @@ RUN if [ "${PROVISION_DSR1}" = "1" ]; then \
...
+ARG PROVISION_VBENCH=1
+ENV UV_PYTHON_INSTALL_DIR=/opt/uv-python
+COPY --chown=${USER_ID}:${GROUP_ID} examples/09_Wan22_VideoGen_Example/accuracy /opt/vbench_accuracy
+RUN if [ "${PROVISION_VBENCH}" = "1" ]; then \
+        cd /opt/vbench_accuracy && \
+        sed -i 's#requires-python = ">=3.12"#requires-python = ">=3.10"#' pyproject.toml && \
+        sed -i 's#    "vbench==0.1.5",#    "vbench==0.1.5",\n    "decord2>=3.4.0",#' pyproject.toml && \
+        sed -i "s#^package = false#package = false\noverride-dependencies = [\"decord; sys_platform == 'never'\"]#" pyproject.toml && \
+        UV_PROJECT_ENVIRONMENT="$(pwd)/.venv" uv lock --python 3.11 && \
+        UV_PROJECT_ENVIRONMENT="$(pwd)/.venv" uv sync --python 3.11 && \
```
arekay-nv: **medium** — flagged by Grok

`uv sync` pulls torch from the default index with no CUDA pin. On the aarch64/GB200 target the default wheel may be CPU-only, which then fails the runtime `torch.cuda.is_available()` gate (`vbench_runner.py:119`). Please verify the synced torch is a CUDA build on aarch64 — on `linux/amd64` the default wheel is CUDA-enabled, so this likely only bites arm64.

scripts/Dockerfile.dev
```diff
@@ -77,3 +86,24 @@ RUN if [ "${PROVISION_DSR1}" = "1" ]; then \
     else \
         echo "PROVISION_DSR1=${PROVISION_DSR1}: skipping DeepSeek-R1 evaluator provisioning" ; \
     fi
+
+# VBench (WAN 2.2) accuracy scorer, gated by PROVISION_VBENCH (default 1, like PROVISION_DSR1).
+# It lives under examples/ (not src/), so COPY it in and build its own .venv at
+# /opt/vbench_accuracy; accuracy runs reuse it via UV_NO_SYNC=1. Isolated like DSR1: pinned to
```
arekay-nv: **medium** — flagged by Grok, Claude

This comment says accuracy runs reuse the venv via `UV_NO_SYNC=1`, but `UV_NO_SYNC` is set nowhere in the image or the scorer. At runtime `uv run --project` may therefore re-resolve/sync (and hit the network), defeating the self-containment this PR is for. Either set `ENV UV_NO_SYNC=1` (and confirm the scorer honors it) or drop the claim.

examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
```diff
@@ -85,6 +93,29 @@ def main() -> int:
     )
     args = parser.parse_args()
 
+    # vbench downloads weights via wget/unzip subprocesses on a cold cache;
+    # fail fast instead of a FileNotFoundError mid-evaluation.
+    missing_tools = [t for t in ("wget", "unzip") if shutil.which(t) is None]
```
arekay-nv: **medium** — flagged by Claude

This `wget`/`unzip` guard runs unconditionally, so it aborts even when the VBench weights are already cached — contradicting the remediation text below ("...or pre-populate the cache"). Consider checking cache presence first, or downgrading to a warning when the cache is already populated.

scripts/Dockerfile.dev
```diff
@@ -77,3 +86,24 @@ RUN if [ "${PROVISION_DSR1}" = "1" ]; then \
...
+# the install wheel-only on aarch64. PROVISION_VBENCH=0 skips it.
+ARG PROVISION_VBENCH=1
+ENV UV_PYTHON_INSTALL_DIR=/opt/uv-python
+COPY --chown=${USER_ID}:${GROUP_ID} examples/09_Wan22_VideoGen_Example/accuracy /opt/vbench_accuracy
+RUN if [ "${PROVISION_VBENCH}" = "1" ]; then \
+        cd /opt/vbench_accuracy && \
+        sed -i 's#requires-python = ">=3.12"#requires-python = ">=3.10"#' pyproject.toml && \
+        sed -i 's#    "vbench==0.1.5",#    "vbench==0.1.5",\n    "decord2>=3.4.0",#' pyproject.toml && \
+        sed -i "s#^package = false#package = false\noverride-dependencies = [\"decord; sys_platform == 'never'\"]#" pyproject.toml && \
+        UV_PROJECT_ENVIRONMENT="$(pwd)/.venv" uv lock --python 3.11 && \
```
arekay-nv: **medium** — flagged by Grok

`uv lock --python 3.11` regenerates the lockfile at build time instead of syncing the committed `examples/09_Wan22_VideoGen_Example/accuracy/uv.lock`, so image dependencies can drift between rebuilds and diverge from the committed lock. Prefer `uv sync --frozen` against a committed 3.11 lock (or commit the patched pyproject + lock) for reproducibility.

examples/09_Wan22_VideoGen_Example/accuracy/vbench_runner.py
```diff
@@ -85,6 +93,29 @@ def main() -> int:
...
+                        "wget/unzip on first use. Install them in the client "
+                        "environment (e.g. `apt-get install wget unzip`) or "
+                        "pre-populate the VBench cache directory."
+                    ),
+                }
+            ),
+            file=sys.stderr,
+            flush=True,
+        )
+        return 2
```
arekay-nv: **medium** — flagged by Grok, Claude

Exit code `2` is returned here for missing tools, but the CUDA-unavailable path (line 135) also returns `2`, while the module docstring (line 26) documents `2` as "CUDA required but unavailable" only. Callers cannot distinguish the two failures — update the docstring (ideally use distinct codes).

arekay-nv (commented):

Reviewed by **Codex** + **Grok** (cursor-cli) + **Claude** · depth: **thorough**

Posting **7** findings (2 high, 5 medium). Filtered: 3 low-severity `consider` items and 5 pre-existing/out-of-scope items on `scoring.py`/tests not touched by this diff. The two high findings were each flagged **independently by all three reviewers** and **reproduced locally** by building this PR's image.

### 🔴 Must fix — both defeat the self-contained `--mode acc` image
| File:line | Issue |
|---|---|
| `Dockerfile.dev:98` | Baked at `/opt/vbench_accuracy` but scorer default path is `examples/09_…/accuracy` and no `VBENCH_PROJECT_PATH` is set → `FileNotFoundError` at init. |
| `Dockerfile.dev:31` | Runner hard-fails without `wget`/`unzip`; apt installs only `libgl1 libglib2.0-0`. |

### 🟡 Should fix
| File:line | Issue |
|---|---|
| `Dockerfile.dev:105` | `uv sync` no CUDA pin → possibly CPU-only torch on aarch64 (verify). |
| `Dockerfile.dev:92` | Comment claims `UV_NO_SYNC=1` reuse, but it is set nowhere. |
| `vbench_runner.py:98` | Missing-tools guard is unconditional; aborts even with a warm cache. |
| `Dockerfile.dev:104` | Build-time `uv lock` regenerates rather than syncing the committed lock → drift. |
| `vbench_runner.py:117` | Exit code `2` overloaded (missing-tools vs CUDA); docstring stale. |

_Out of scope (pre-existing, not this diff): `scoring.py` partial-completeness (1597), unguarded results read (1645), prompt-filename collision (1374). The `.mp4` staging change itself was reviewed and is correct._

arekay-nv (approved):
LGTM - thanks!

scripts/Dockerfile.dev
wu6u3tw: Fixed. See d2349e1.

scripts/Dockerfile.dev
wu6u3tw: Fixed. See d2349e1.

PR #404, opened 2026-07-08
fix(endpoint-client): retry pre-response connection resets (opt-in)

Palanivelg:
## Summary
- Adds an **opt-in** transport retry that re-issues a request on a fresh connection when the server closes an idle keep-alive socket *before any response byte arrives* (the single-slot-server reuse/reset race). It never fires once headers/chunks have started, so it cannot duplicate output.
- New `HTTPClientConfig.transport_max_retries` — **default `0` = prior fail-fast behavior, so every existing use case is unchanged**. The edge-agentic accuracy reference config opts in explicitly.
- Root cause: reused-then-reset sockets on single-slot servers failed requests before any response arrived, silently zeroing the affected accuracy sample. The retry recovers them without disabling connection reuse.

## Changes
- `endpoint_client/worker.py`: `_read_headers_with_retry` helper
- `endpoint_client/config.py`: `transport_max_retries` field (default 0)
- `examples/11_Edge_Agentic_Example/online_edge_full_run.yaml`: opts in
- regenerated `_full` config templates (schema-driven)
- `tests/unit/endpoint_client/test_worker_retry.py`: unit tests

## Test plan
- [x] `pytest tests/unit/endpoint_client/test_worker_retry.py` — all pass
- [x] `pre-commit run` on all changed files (ruff, ruff-format, mypy, template regen, license) — all pass
- [x] Full accuracy pass with the fix enabled: no connection resets; accuracy back on the reference baseline (previously degraded by dropped requests)

### DIFF
--- a/examples/11_Edge_Agentic_Example/online_edge_full_run.yaml
+++ b/examples/11_Edge_Agentic_Example/online_edge_full_run.yaml
@@ -106,6 +106,12 @@ settings:
     num_workers: 1
     max_connections: 1
     warmup_connections: 0
+    # Single-slot edge servers (e.g. llama.cpp -np 1) close idle keep-alive
+    # sockets between samples; a reused-then-reset socket would otherwise zero
+    # the affected accuracy sample. Retry the request on a fresh connection when
+    # the reset happens before any response byte (safe: the server produced no
+    # output). Off by default globally; enabled here for the gated run.
+    transport_max_retries: 2
 
 endpoint_config:
   endpoints:
--- a/src/inference_endpoint/compliance/checker.py
+++ b/src/inference_endpoint/compliance/checker.py
@@ -157,6 +157,21 @@ def check_config_lock(config: dict[str, Any]) -> list[Check]:
         detail += f" target_concurrency={target_concurrency}"
     checks.append(Check("single_stream", single_stream, detail))
 
+    # Lock the pre-response reset retry: it is a reproducibility knob that lets
+    # the client silently re-issue after a SUT-side reset, so its value must be
+    # recorded (and within the schema bound) rather than passing for any value.
+    # Disclosure that a recovered reset didn't mask a dropped turn is a manual
+    # attestation added in check_submission (transport_retries_total lives in the
+    # run's result summary, not in this config).
+    retries = _get(config, "settings", "client", "transport_max_retries", default=0)
+    checks.append(
+        Check(
+            "transport_max_retries_locked",
+            isinstance(retries, int) and 0 <= retries <= 5,
+            f"transport_max_retries={retries}",
+        )
+    )
+
     return checks
 
 
@@ -309,6 +324,22 @@ def check_submission(
             report.add("config_valid", False, f"{config_path} failed to load: {e}")
         else:
             report.checks.extend(check_config_lock(config))
+            retries = _get(
+                config, "settings", "client", "transport_max_retries", default=0
+            )
+            if retries:
+                # A recovered pre-response reset produces a clean success, so
+                # the no_dropped_turns gate can't see it from scores.json alone.
+                # Require the submitter to confirm the disclosed retry value and
+                # that transport_retries_total (in the run's result summary) did
+                # not mask a dropped-and-re-issued turn.
+                report.notes.append(
+                    f"transport_max_retries={retries}: pre-response connection "
+                    "resets are retried. Confirm this matches the disclosed "
+                    "ruleset value and that no retry-recovered reset (see "
+                    "transport_retries_total in the run's result summary) masked "
+                    "a dropped turn."
+                )
 
     results_path = report_dir / "results.json"
     scores_path = report_dir / "scores.json"
--- a/src/inference_endpoint/endpoint_client/config.py
+++ b/src/inference_endpoint/endpoint_client/config.py
@@ -170,6 +170,23 @@ class HTTPClientConfig(WithUpdatesMixin, BaseModel):
         4.0, description="Discard connections idle longer than this (seconds)"
     )
 
+    # Retry a request when the connection fails BEFORE any response byte is
+    # received (e.g. "connection closed before headers received" — the server
+    # closed an idle keep-alive socket at the instant we wrote the request).
+    # Safe because the server produced no output; the retry re-issues on a fresh
+    # connection. Never fires once headers/chunks have started (would risk
+    # duplicate output). Guards single-stream localhost runs (e.g. llama.cpp with
+    # -np 1) where a stale-socket reset otherwise zeroes an accuracy sample.
+    #
+    # Opt-in: default 0 preserves the prior fail-fast behavior for every use
+    # case. Configs that need it (e.g. the edge-agentic accuracy reference)
+    # enable it explicitly via `client.transport_max_retries`.
+    transport_max_retries: int = Field(
+        0,
+        ge=0,
+        description="Retries on a pre-response connection reset (0=disabled)",
+    )
+
     # Minimum required connections for http-client to initialize.
     # Will log warning if not enough ephemeral ports are available during warmup.
     #
--- a/src/inference_endpoint/endpoint_client/config.py
+++ b/src/inference_endpoint/endpoint_client/config.py
@@ -181,10 +181,13 @@ class HTTPClientConfig(WithUpdatesMixin, BaseModel):
     # Opt-in: default 0 preserves the prior fail-fast behavior for every use
     # case. Configs that need it (e.g. the edge-agentic accuracy reference)
     # enable it explicitly via `client.transport_max_retries`.
+    # Bounded above so a genuinely-down server can't turn fail-fast into a long
+    # reconnect/re-write spin that masks a hard SUT failure and multiplies load.
     transport_max_retries: int = Field(
         0,
         ge=0,
-        description="Retries on a pre-response connection reset (0=disabled)",
+        le=5,
+        description="Retries on a pre-response connection reset (0=disabled, max 5)",
     )
 
     # Minimum required connections for http-client to initialize.
--- a/src/inference_endpoint/endpoint_client/worker.py
+++ b/src/inference_endpoint/endpoint_client/worker.py
@@ -28,6 +28,8 @@ from collections.abc import AsyncGenerator
 from typing import Any
 from urllib.parse import urlparse
 
+import msgspec
+
 from inference_endpoint.async_utils.transport import (
     ReceiverTransport,
     SenderTransport,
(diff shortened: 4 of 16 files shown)


examples/11_Edge_Agentic_Example/online_edge_full_run.yaml
```diff
@@ -106,6 +106,12 @@ settings:
     num_workers: 1
     max_connections: 1
     warmup_connections: 0
+    # Single-slot edge servers (e.g. llama.cpp -np 1) close idle keep-alive
+    # sockets between samples; a reused-then-reset socket would otherwise zero
+    # the affected accuracy sample. Retry the request on a fresh connection when
+    # the reset happens before any response byte (safe: the server produced no
+    # output). Off by default globally; enabled here for the gated run.
+    transport_max_retries: 2
```
nvzhihanj, line 114: @viraatc can you help review this PR? In general I want to avoid adding more tuning flags to our http client if possible

src/inference_endpoint/endpoint_client/worker.py
```diff
@@ -378,14 +378,50 @@ async def _fire_request(self, req: InFlightRequest) -> bool:
...
+        No output was produced, so re-issuing the identical request on a fresh
+        connection is safe and idempotent. Returns the HTTP status code, or None
+        if retries are exhausted (an error response has been emitted).
+        """
+        attempt = 0
+        while True:
+            conn = req.connection
+            try:
+                status_code, _ = await conn.protocol.read_headers()
+                return status_code
```
arekay-nv, line 404: [Council: Claude + manual trace] **high — perf-regression masking**

A *successful* retry returns the status here with **zero trace** — no log, no metric, no counter, no per-sample flag, no event. Only the exhausted path (`_handle_error`, ~L404) and the disabled path record anything. So when a server resets connections under load — a "SUT can't keep up" signal you'd use to tune `max_connections`/`max_idle_time`/`warmup_connections`/server keep-alive — the retry converts it into a clean success and the signal never reaches the metrics/report.

Suggest: `logger.warning(...)` on each recovery + an aggregatable counter (e.g. `transport_retries_total`) + a per-sample retry flag on `QueryResult`, so the Report can surface how many samples needed re-issue.

src/inference_endpoint/endpoint_client/worker.py
```diff
@@ -378,14 +378,50 @@ async def _fire_request(self, req: InFlightRequest) -> bool:
             logger.error(f"Request {req.query_id} failed: {type(e).__name__}: {e}")
             return False
 
+    async def _read_headers_with_retry(self, req: InFlightRequest) -> int | None:
```
arekay-nv, line 383: [Council: Claude + manual trace] **high — latency pollution attributed to the server**

This whole method runs inside the timed `_process_response`, so a retried sample's elapsed window now folds in: waiting for `connection_lost` on the dead socket (server FIN/RST timing) + `_pool.acquire()` of a fresh conn + re-`write()` + a **full second server processing pass** — with no marker distinguishing it. On single-slot accuracy runs (the exact scenario this targets) resets are the *common* case, so this inflates p99/max for precisely the samples the feature exists to handle.

Suggest flagging retried samples in `QueryResult` so the report can segregate/exclude them instead of blending polluted latency into the server's distribution.

src/inference_endpoint/endpoint_client/config.py
```diff
@@ -170,6 +170,23 @@ class HTTPClientConfig(WithUpdatesMixin, BaseModel):
...
+    # closed an idle keep-alive socket at the instant we wrote the request).
+    # Safe because the server produced no output; the retry re-issues on a fresh
+    # connection. Never fires once headers/chunks have started (would risk
+    # duplicate output). Guards single-stream localhost runs (e.g. llama.cpp with
+    # -np 1) where a stale-socket reset otherwise zeroes an accuracy sample.
+    #
+    # Opt-in: default 0 preserves the prior fail-fast behavior for every use
+    # case. Configs that need it (e.g. the edge-agentic accuracy reference)
+    # enable it explicitly via `client.transport_max_retries`.
+    transport_max_retries: int = Field(
```
arekay-nv, line 186: [Council: Claude + manual trace] **high — reproducibility / compliance hole on this knob**

1. `compliance/checker.py:check_config_lock` pins seed + single-stream (`num_workers`/`max_connections`/`target_concurrency`) but **never inspects `transport_max_retries`**, even though the resolved config carries it at `settings.client.transport_max_retries` and the edge-agentic reference (`online_edge_full_run.yaml:114`) sets it to `2`. The config-lock gate therefore passes for *any* value, it isn't surfaced as a manual attestation, and a retry-recovered reset can let the `no_dropped_turns` gate pass on a run where a turn was effectively dropped-and-re-issued.
2. The field is unbounded (`ge=0`, no `le=`): against a genuinely-down server every attempt re-acquires/re-writes/re-raises, turning fast-fail into a long connection-churn spin that masks a hard SUT failure and multiplies load.

Suggest: pin `transport_max_retries` in `check_config_lock` to the ruleset value (0 for perf, the disclosed value for gated accuracy), add `le=` (e.g. 5), and record recovered resets so "reproducible + 0 dropped turns" can't be satisfied silently. (Dovetails with the maintainer note about limiting tuning flags.)

src/inference_endpoint/endpoint_client/worker.py
```diff
@@ -378,14 +378,50 @@ async def _fire_request(self, req: InFlightRequest) -> bool:
...
+        connection is safe and idempotent. Returns the HTTP status code, or None
+        if retries are exhausted (an error response has been emitted).
+        """
+        attempt = 0
+        while True:
+            conn = req.connection
+            try:
+                status_code, _ = await conn.protocol.read_headers()
+                return status_code
+            except ConnectionError as e:
```
arekay-nv, line 405: [Council: Codex + Claude + manual trace] **medium — retries on a *partial*-response close, not just a true zero-byte close**

`connection_lost` raises `ConnectionResetError` whenever headers aren't complete (`http.py` ~L216), so a peer that emitted *partial* status-line/header bytes and then dropped still lands in this `except` and re-issues. Client output isn't duplicated (the accumulator only runs after a 200), but the server may have already **received and processed** the request → duplicate generation work, i.e. extra invisible load that worsens the very overload that caused the reset. The docstring's "no output was produced… safe and idempotent" and the comment "no headers arrived" overstate this — it's client-output-idempotent, not server-side-idempotent.

Suggest tracking whether any response bytes were received and only retrying true zero-byte closes; scope the docstring claim to client output.

src/inference_endpoint/endpoint_client/worker.py
```diff
@@ -406,8 +442,9 @@ async def _process_response(self, req: InFlightRequest) -> None:
             logger.warning(f"Request {req.query_id} failed: {type(e).__name__}: {e}")
 
         finally:
-            # Release connection back to pool if not already
-            self._pool.release(conn)
+            # Release the current connection (body handlers release early; this
+            # is idempotent and also covers the swapped connection after a retry)
+            self._pool.release(req.connection)
```
arekay-nv, line 463: [Council: Claude + manual trace] **medium — "idempotent" is inaccurate for a dead connection**

`ConnectionPool.release` (`http.py:586`) only early-returns on `not conn.in_use`, and its dead branch closes the transport **without clearing `in_use`**. In the retry-exhausted and retries-disabled paths the last (dead) connection is released once inside `_read_headers_with_retry` (~L402) and again here, so the second call re-enters the dead branch and fires a spurious `_notify_waiter()`. Impact is low in single-slot mode (a spurious wakeup that re-waits), but the comment will mislead maintainers.

Suggest either resetting `in_use` in the dead branch of `release`, or dropping the idempotency claim and guarding the double-release explicitly.

tests/unit/endpoint_client/test_worker_retry.py
```diff
@@ -0,0 +1,223 @@
...
+    assert req.connection is fresh
+    # Dead connection discarded exactly once; request re-written on the fresh one.
+    assert pool.released == [dead]
+    assert fresh.protocol.written == [req.http_bytes]
+    assert worker._handle_error_calls == []
+
+
+@pytest.mark.unit
+@pytest.mark.asyncio
+async def test_retry_exhausted_emits_error_and_returns_none():
```
arekay-nv, line 185: [Council: Claude] **medium — test gaps on this PR's risk areas**

Every test calls `_read_headers_with_retry` directly and never drives `_process_response`, so the `finally: self._pool.release(req.connection)` double-release path (worker.py L447) is never exercised. `_FakeConn`/`_FakePool` don't model `in_use`, so they couldn't detect the dead-conn idempotency defect even if driven. There's no multi-retry-then-success case (e.g. `["reset","reset",200]` with `max_retries=2`), and nothing asserts observability of a successful retry (there's nothing to assert — itself the gap).

Suggest an integration-style test that drives `_process_response` end-to-end with an `in_use`-tracking fake pool.

arekay-nv (commented):

Reviewed by: **Codex** (gpt-5.5, xhigh) + **Claude** + manual trace | Depth: **thorough**

Small diff, but two design decisions land on the reproducibility / perf-masking concerns raised for this PR. 6 issues posted inline.

### 🔴 Must Fix (high)
| File:Line | Category | Summary |
|---|---|---|
| `worker.py:396` | observability | Successful retry is **invisible** (no log/metric/counter/flag/event) → server resets under load masked as clean successes instead of surfaced to tune connection settings |
| `worker.py:381` | performance | Retried-sample latency (dead-socket wait + reacquire + re-write + full 2nd server pass) is folded into the sample and **attributed to the server** with no marker; inflates p99/max on exactly the single-slot samples this targets |
| `config.py:184` | reproducibility | `transport_max_retries` **not pinned by `check_config_lock`** and **unbounded** (`ge=0`, no `le=`); recovered resets can silently satisfy `no_dropped_turns`; reference config sets it to 2 |

### 🟡 Should Fix (medium)
| File:Line | Category | Summary |
|---|---|---|
| `worker.py:397` | correctness | Retries on a **partial**-response close, not just zero-byte → possible duplicate server-side generation; "safe and idempotent" overstated *(Codex + Claude)* |
| `worker.py:447` | concurrency | `release` **not idempotent for a dead conn** (never clears `in_use`) → double-release fires spurious `_notify_waiter`; comment inaccurate |
| `test_worker_retry.py:124` | testing | Tests never drive `_process_response`, don't model `in_use`, no multi-retry-then-success, no observability assertion |

_Note: the earlier agent comment on `worker.py` (write-before-assign leak) is resolved by commit `b9ca904`._

arekay-nv (changes requested):
Can we root cause this more accurately. Ideally we would like to tune the other settings and not have silent retries impacting the performance.

Palanivelg:
With tensorrt_edge_llm and vllm, the issue is not reproducible. But with llama.cpp, especially without MTP the failures are observed occasionally. Will revisit this after the v6.1 submissions.

src/inference_endpoint/endpoint_client/worker.py
```diff
@@ -378,14 +380,64 @@ async def _fire_request(self, req: InFlightRequest) -> bool:
...
+                status_code, _ = await conn.protocol.read_headers()
+                return status_code
+            except ConnectionError as e:
+                partial = conn.protocol.bytes_received
+                self._pool.release(conn)
+                if (
+                    partial
+                    or req.transport_retries >= self.http_config.transport_max_retries
+                ):
+                    await self._handle_error(req.query_id, e)
```
viraatc, line 412: Should we count/log retries on failed samples too? A sample that retries twice and then exhausts goes through `_handle_error`, which builds the error result without the count — so `transport_retries_total` stays 0 on a run where every retry was exhausted. That's the run where you most want the signal, and the compliance note points people at this counter.

src/inference_endpoint/load_generator/session.py
```diff
@@ -600,6 +600,21 @@ def _handle_response(self, resp: QueryResult | StreamChunk) -> None:
...
+            # regardless of the tracking window.
+            if resp.transport_retries:
+                retry_record = EventRecord(
+                    event_type=SampleEventType.TRANSPORT_RETRY,
+                    timestamp_ns=time.monotonic_ns(),
+                    sample_uuid=query_id,
+                    conversation_id=conv_id_str,
+                    turn=turn_num,
+                )
+                for _ in range(resp.transport_retries):
```
viraatc, line 615: Could publish once with the count instead of N duplicates:

```python
if resp.transport_retries:
    self._publisher.publish(
        EventRecord(
            event_type=SampleEventType.TRANSPORT_RETRY,
            timestamp_ns=time.monotonic_ns(),
            sample_uuid=query_id,
            conversation_id=conv_id_str,
            turn=turn_num,
            data=resp.transport_retries,
        )
    )
```
with `registry.increment(..., delta=...)` on the aggregator side. N identical rows with the same timestamp in `events.jsonl` makes post-processing guess.

src/inference_endpoint/compliance/checker.py
```diff
@@ -157,6 +157,21 @@ def check_config_lock(config: dict[str, Any]) -> list[Check]:
...
+    # the client silently re-issue after a SUT-side reset, so its value must be
+    # recorded (and within the schema bound) rather than passing for any value.
+    # Disclosure that a recovered reset didn't mask a dropped turn is a manual
+    # attestation added in check_submission (transport_retries_total lives in the
+    # run's result summary, not in this config).
+    retries = _get(config, "settings", "client", "transport_max_retries", default=0)
+    checks.append(
+        Check(
+            "transport_max_retries_locked",
+            isinstance(retries, int) and 0 <= retries <= 5,
```
viraatc, line 170: Can we avoid the hard-code 5 here?
The schema already enforces `ge=0, le=5` — a config outside that range can't produce a run. Re-asserting the bound here hardcodes a second `5` that will drift if the schema changes. If the point is recording the value in the report, a record-only check (always-pass, value in `detail`) does that without the magic number.

viraatc (commented):
my apologies for letting this review slip,
overall implementation looks really clean!

ran a few tests locally:

- Detection is immediate (RST arrives, no timeout wait): ~0.1ms
- Reconnect + re-issue on a fresh connection: ~0.3ms
- Net added latency per retried sample: ~0.2–0.5ms (roughly one TCP reconnect — the second pass replaces the first, it doesn't stack, and for a true zero-byte reset the server never started on attempt 1, so no duplicated generation either)

arekay-nv:
@Palanivelg can you rebase this on main.

PR #405, opened 2026-07-09
feat :Testing workflow to push images

arekay-nv:
## What does this PR do?
Adds a reproducible way to build and publish the endpoints client image to
[`ghcr.io/mlcommons/endpoints`](https://github.com/orgs/mlcommons/packages/container/endpoints/996854134):

- **`scripts/push_docker_image.sh`** — builds `scripts/Dockerfile.dev` at a chosen
  git ref via `docker buildx` and pushes it, tagged with both the short commit SHA
  and the ref name so consumers can pin either.
- **`.github/workflows/publish-image.yml`** — a manually-triggered
  (`workflow_dispatch`) workflow that runs the script, so a publish can be kicked
  off from the Actions tab against any selected branch/tag/SHA without a local
  Docker setup.

The workflow reuses the script (single source of truth for the tag scheme) rather
than duplicating build logic.

## Behavior & design

| Aspect | Choice |
|---|---|
| Default platform | `linux/amd64` (what benchmark clients run on); multi-arch is one flag/input away |
| Multi-arch | `--multi-arch` / `linux/amd64,linux/arm64` publishes a manifest list usable on x86 **and** arm |
| Tags | `:<short-sha>` + `:<sanitized-ref>` |
| Builder | self-provisions a `docker-container` buildx builder (the default `docker` driver can neither `--push` nor build multi-platform) |
| Auth | script logs in via `GHCR_USER`/`GHCR_TOKEN` if set, else assumes `docker login`; workflow uses `github.actor` + the automatic `GITHUB_TOKEN` |
| Safety | refuses to publish a `:$SHA` tag from a dirty tree (`--allow-dirty` to override); optional ref checkout is restored on exit |
| `PROVISION_DSR1` | passthrough build-arg (default = Dockerfile default `1`); set `0` for a lean client-only image |

## CI prerequisites (one-time, outside this PR)

- Workflow must land on the default branch before the "Run workflow" button appears.
- `permissions: packages: write` is set in the workflow; the org must allow the
  `GITHUB_TOKEN` to write packages, and the `endpoints` package must grant this repo
  Write access (Package → *Manage Actions access*). Usually automatic since the
  package lives under this repo's org.

> Note: `GITHUB_TOKEN` authenticates as `github-actions[bot]` (scoped to this repo);
> `username: github.actor` only sets the login name. For a push attributed to a
> specific account, supply that account's PAT as a secret instead.

## Validation

- **Multi-arch dry build** (`--multi-arch --no-push`): both `linux/amd64` (emulated)
  and `linux/arm64` (native) compiled all stages, exit `0`.
- Script is clean under **macOS bash 3.2.57** and `bash -n`; the empty-array
  expansion and tag sanitizer were unit-checked; the dirty-tree guard was exercised
  across push/dry/allow-dirty/clean/checked-out cases.
- Reviewed via a two-model council (Codex + Claude); all findings fixed: detached-HEAD
  ref resolution, Docker-tag sanitization beyond `/`, bash-3.2 empty-array crash, and
  the dirty-tree publish guard. Workflow security posture confirmed (inputs passed via
  `env:` not inline `${{ }}`, actions pinned to SHAs, minimal `contents:read` +
  `packages:write`).

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass


scripts/push_docker_image.sh
```diff
@@ -0,0 +1,208 @@
...
+# Docker tags allow only [A-Za-z0-9_.-], must not start with '.' or '-', and cap at
+# 128 chars. Map every other char (e.g. '/', '+', '~', '#') to '-' so a legitimate
+# git ref name can't yield an invalid tag that fails `-t` only after the whole build.
+REF_TAG="$(printf '%s' "$ENDPOINTS_REF" | tr -c 'A-Za-z0-9_.-' '-')"
+[[ "$REF_TAG" == [A-Za-z0-9_]* ]] || REF_TAG="_${REF_TAG}"
+REF_TAG="${REF_TAG:0:128}"
+
+# Building the working tree as-is (no distinct ref checked out): don't publish a
+# :$SHORT_SHA tag whose '.' build context doesn't match commit $SHORT_SHA.
+if [[ -z "${DID_CHECKOUT:-}" && "$ALLOW_DIRTY" != "1" ]] && tree_has_tracked_changes; then
```
arekay-nv, line 159: [Codex + Claude] **medium** (build reproducibility + context): The dirty guard here uses `git diff --quiet HEAD`, which **ignores untracked files**. But `scripts/Dockerfile.dev:49` does `COPY src/ ./src/`, so an untracked file under `src/` gets baked into the image — a local `--push` can publish `:${SHORT_SHA}` whose contents don't match commit `${SHORT_SHA}`.

Separately, the comment at line 124 claims *".dockerignore governs the build context"*, but **no `.dockerignore` exists** in the repo, so the whole working tree is uploaded as build context (`docker buildx build … .`, line 197), pulling in large untracked dirs (`cached_datasets/`, `gtc_results/`, `dsr1_v61_seed_logs_*/`) on every local run.

**Fix:** add a `.dockerignore` (exclude artifacts + anything not `COPY`ed); for the SHA-tag correctness, also include untracked files under `COPY`ed paths in the cleanliness check, or build from `git archive`.

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,82 @@
...
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout selected ref
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          # Check out the ref *name* (not github.sha) so a local branch head exists
+          # and the script can resolve/tag it; blank input falls back to the
+          # dispatched branch/tag. Full history for `git rev-parse --short`.
+          ref: ${{ inputs.ref || github.ref_name }}
```
arekay-nv, line 56: [Codex] **medium** (correctness): This checks out the dispatch ref directly, replacing the tree before the run step. Building a ref that **predates this workflow** means `./scripts/push_docker_image.sh` won't exist and the run fails — even though the UI advertises building arbitrary refs. It also lets the target ref supply the script that runs *after* GHCR login.

**Fix:** keep the checkout on the workflow's own (trusted) ref, and pass the target via `ENDPOINTS_REF` so the *script* switches refs only for the Docker build. (Ensure the target ref is fetched — the script's `origin/<ref>` fallback needs the object present.)

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,82 @@
...
+          fetch-depth: 0
+
+      # Registers QEMU binfmt handlers so the docker-container builder can build
+      # the non-native arch when platforms includes linux/arm64.
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8 # v4.2.0
+
+      - name: Set up Docker Buildx
```
arekay-nv, line 64: [Claude] **low** (design): This provisions a buildx builder that the script never uses — `push_docker_image.sh` always creates and selects its own `endpoints-buildx` (docker-container) builder (lines 173/191). buildx is preinstalled on `ubuntu-latest`, so this step is effectively idle. Either drop it, or have the script honor a caller-supplied builder. (The `Set up QEMU` step above **is** needed for multi-arch — keep it.)

arekay-nv (commented):

Reviewed by **Codex** (gpt-5.5, xhigh) + **Claude**. 3 inline findings below (2 medium, 1 low; no critical/high).

arekay-nv:

Reviewed by: **Codex** (gpt-5.5, xhigh) + **Claude** | Depth: **standard** (low-severity C included on request)

Found **3** issues, all posted inline. No critical/high.

| A | `scripts/push_docker_image.sh` | 158 | medium | reproducibility | Both | Dirty guard uses `git diff --quiet HEAD` → ignores untracked files; `Dockerfile.dev` `COPY src/` can bake them into `:{sha}`. Also no `.dockerignore` exists (contradicts the line-124 comment) → whole tree uploaded as context. |
| B | `.github/workflows/publish-image.yml` | 56 | medium | bug | Codex | Checks out arbitrary dispatch ref directly; refs predating this workflow lack the script and fail, and the target ref supplies the post-login script. |
| C | `.github/workflows/publish-image.yml` | 64 | low | design | Claude | `Set up Docker Buildx` provisions a builder the script never uses (it always creates its own `endpoints-buildx`). QEMU step is needed — keep. |

> ℹ️ The existing `gemini-code-assist` comment on `push_docker_image.sh:196` (array expansion "missing outer quotes") is a **false positive** — `${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}` is already the correct Bash 3.2 / `set -u` empty-array guard; the suggested rewrite is not an improvement.

<sub>All comments are neutral (`event: COMMENT`) — no approve/request-changes.</sub>

scripts/push_docker_image.sh
```diff
@@ -0,0 +1,238 @@
...
+    )
+fi
+
+BUILD_VERB=$([[ "$PUSH" == "1" ]] && echo "and pushing" || echo "(dry run, --no-push)")
+echo ">> Building ${IMAGE}:${SHORT_SHA} (ref: ${ENDPOINTS_REF}) for ${PLATFORM} ${BUILD_VERB}"
+docker buildx build \
+    --builder "$BUILDX_BUILDER" \
+    --platform "$PLATFORM" \
+    -f "$DOCKERFILE" \
+    -t "${IMAGE}:${SHORT_SHA}" \
```
viraatc, line 262: One thing I noticed: re-running this for the same commit with a different `PLATFORM`/`PROVISION_DSR1` re-pushes `:${SHORT_SHA}` and quietly swaps out the bytes a consumer may have pinned. `push_image.sh` in this PR already has a nice guard for exactly this (`imagetools inspect` + `--force`) — maybe reuse it here, or settle on one canonical config per SHA?

scripts/push_docker_image.sh
```diff
@@ -0,0 +1,238 @@
...
+        echo "       Commit/stash the changes, or pass --allow-dirty to publish anyway." >&2
+        exit 1
+    fi
+    echo ">> warning: ${dirty_msg} (dry run; nothing pushed)." >&2
+fi
+
+# ---------------------------------------------------------------------------
+# Ensure a docker-container builder (supports --push and multi-platform).
+# ---------------------------------------------------------------------------
+if ! docker buildx inspect "$BUILDX_BUILDER" >/dev/null 2>&1; then
```
viraatc, line 202: Small thing: this mirrors the builder setup in `livecodebench/push_image.sh`, but that one also recreates a stale builder when QEMU was registered after it bootstrapped (`unsupported_platforms()` + recreate-once, ~L180-196). Without that, `--multi-arch` on an older `endpoints-buildx` builder fails with a raw buildx error. Worth porting that bit over — or factoring the shared builder logic into a sourced helper like `_image_env.sh`?

viraatc (approved):

PR #406, opened 2026-07-10
fix: Salt failure is a hard error

arekay-nv:
## What does this PR do?

Makes salt failures a **hard error** instead of a silent skip. Previously, when `--warmup-salt` was on (the default) but a sample couldn't be salted, `Dataset._apply_salt` logged a warning (or silently passed the sample through) and the warmup ran with **no** cache-busting — the exact condition salt exists to prevent, failing quietly.

Now `salt=True` requires every sample to be a `dict` with a text (`str`) `prompt` and no `input_tokens`; anything else raises `DatasetValidationError` **at dataset-load time**, before any load is issued.

**Key changes:**

| Area | Change |
|------|--------|
| Validation | New `Dataset.validate_saltable()` — rejects non-`dict` samples, `input_tokens` (pre-tokenized; adapters send these verbatim so a salted `prompt` never reaches the server), missing `prompt`, and non-`str` `prompt`. Error names the offending sample index + remediation. |
| Fail-fast placement | Called from `_load_datasets` when `warmup.enabled and warmup.salt`, so an unsaltable dataset fails **before** worker/aggregator subprocesses spawn. Also called from `with_salt()` so the salting mechanism stays self-protecting. |
| Hot path | `_apply_salt` collapses to a single dict-merge (`{**data, "prompt": f"[{salt}] {data['prompt']}"}`) — the multimodal-list handling, `input_tokens` warnings, and silent passthroughs are deleted; the contract is guaranteed upstream. |

### ⚠️ Behavior change (note for reviewers)

- **Multimodal / pre-tokenized warmup + salt now hard-errors.** A `prompt` that is a list (image/video workloads) or a dataset with `input_tokens` (e.g. gpt-oss-120b, DeepSeek-R1 via `/v1/completions`) will fail warmup instead of silently skipping salt. Fix: set `--warmup-salt=false` / `warmup.salt: false`.
- **Whole-dataset validation.** The run is rejected if *any* sample is non-conforming — including samples the `n_requests` subset would never issue. Stricter than the old per-sample skip; intended by the hard-error design.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactor/cleanup

## Testing

- [x] Tests added/updated
- [x] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass


src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -437,55 +455,57 @@ def load_sample(self, index: int) -> Any:
...
+        never reach the server). A sample salt cannot bust would silently defeat
+        cache-busting, so it is rejected rather than skipped. Called before any
+        load is issued — at benchmark setup and again from with_salt().
+
+        Raises:
+            DatasetValidationError: naming the first offending sample.
+        """
+        if self.data is None:
+            return
+        for i, sample in enumerate(self.data):
```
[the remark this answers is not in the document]
arekay-nv: the `data` member is never a pandas DataFrame: its only assignment is dataset.py:432

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -437,55 +455,57 @@ def load_sample(self, index: int) -> Any:
...
-                        {**part, "text": f"[{salt}] {part['text']}"},
-                        *prompt[i + 1 :],
-                    ]
-                    return {**data, "prompt": salted_parts}
-            self.logger.warning(
-                "salt=True: multimodal prompt has no text part — "
-                "salt cannot be applied; KV-cache reuse may not be prevented"
-            )
-        return data  # unsupported prompt type — skip salting
+        return {**data, "prompt": f"[{salt}] {data['prompt']}"}
```
nv-alicheng, line 517: I have 2 questions:
1. Have we studied whether or not injecting these query-irrelevant random strings at the beginning affects accuracy in a format-sensitive workload like GPT-OSS that uses harmonize?
2. Is it possible to salt a token_ids list by injecting a random list of ints sampled from a set of known tokens that do not represent special markers?

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -437,55 +455,57 @@ def load_sample(self, index: int) -> Any:
...
+
+        Raises:
+            DatasetValidationError: naming the first offending sample.
+        """
+        if self.data is None:
+            return
+        for i, sample in enumerate(self.data):
+            reason = _salt_violation(sample)
+            if reason is not None:
+                raise DatasetValidationError(
```
nv-alicheng, line 482: Should have been caught earlier, but that is something to address -
`exceptions.py` is very poorly designed and only contains bare name aliases for particular failure states.

This construction of error message and `_salt_violation` should be part of the DatasetValidationError class:

```
class DatasetValidationError:
    ERR_MSG_TEMPLATE = "salt=True requires ... but sample {index} {reason} ..."

    def __init__(self, dataset_idx: int, sample: Sample):
        self.reason = self._violation_reason(sample)
        err_msg = DatasetValidationError.ERR_MSG_TEMPLATE.format(index=dataset_idx, reason=self.reason)
        super().__init__(err_msg)

    def _violation_reason(self, sample: Sample):
        # Body of `_salt_violation` here.
```

Ideally, reason should be some enum or object rather than a raw string, smth like
```
class DatasetValidationError:
     ...

    class Reason(Enum):
         TypeMismatch = ...
         InputTokensShadowing = ...
         PromptMissing = ...
         PromptTypeMismatch = ...
         Other = ...

         def fmt_str(self, sample):
              <handle conversion to full error reason string>
     ...
```

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -257,6 +258,23 @@ def load_from_huggingface(
     return ds[split].to_pandas()
 
 
+def _salt_violation(sample: Any) -> str | None:
+    """Return a human-readable reason a sample cannot be salted, or None if it can.
```
nv-alicheng: code quality - this method is named poorly and is performing 2 tasks at once. IMO this can be remedied by the enum solution mentioned: Rename to `_can_salt`, returning either None or a DatasetValidationError.Reason enum.

DatasetValidationError itself parses the enum -> human readable message.

Otherwise, returning raw strings like this is brittle.

nv-alicheng:
## Review Council — PR #406 (3× Claude, depth: thorough)

Three independent Claude reviewers (bugs/correctness, design/edge, testing/docs) plus a code-quality pass, deduped and severity-recalibrated. **"R"** marks issues flagged independently by multiple reviewers. All line numbers verified against HEAD (`44fc32d7`).

**Verdict:** Solid, focused bug-fix. Fail-fast placement is correct — validation at `_load_datasets` (execute.py:387) runs before any worker/aggregator subprocess spawns, and warmup's `with_salt()` (execute.py:546) salts the same `ctx.dataloader` it re-validates. Transforms apply eagerly in `load()`, so `validate_saltable()` scans the exact post-transform shape `load_sample()` returns — no mismatch. No security issues, green suite. The findings are about **strictness + a default**, not broken logic.

### 🟡 Should fix

| # | File:line | Cat | R | Issue |
|---|-----------|-----|---|-------|
| 1 | `config/schema.py:762` + `execute.py:386` | design | **3×** | **`warmup.salt` defaults `True` + the new hard error = footgun for first-class pre-tokenized workloads.** `openai_completions` (gpt-oss-120b), `sglang`, and DeepSeek-R1 run `Harmonize()`+`ColumnFilter(["input_tokens"])` at load, so their samples always carry `input_tokens` and no `prompt`. Plain `--warmup` (without `--warmup-salt false`) now **aborts setup** for every such run that previously warned and ran unsalted. Options: default `salt=False`, auto-downgrade to warning+unsalted when `input_tokens` is present, or (minimum) state the text-`prompt` requirement in the field `help`/description. |
| 2 | `tests/unit/commands/test_benchmark.py:248` | testing | | The 3 `TestLoadDatasetsSaltValidation` tests `@patch.object(Dataset, "validate_saltable")`, so they only assert called/not-called — they never exercise **real** validation through `_load_datasets`. The exact integration this PR adds (unsaltable perf dataset → `DatasetValidationError` before subprocess spawn) has zero coverage. Add a test that writes an `input_tokens`/int-`prompt` JSONL and asserts `pytest.raises(DatasetValidationError)` through `_load_datasets` unpatched. |
| 3 | `execute.py:386` | api-contract | | **Whole-dataset strictness vs the warmup subset.** Validation rejects the entire run if *any* sample is non-saltable, but salt only touches warmup, which issues just a seeded `n_requests` subset — and the perf phase runs unsalted regardless. A mostly-text dataset with one anomalous row now hard-fails at load even if that row is never issued. Consider scoping validation to the warmup-issued subset. |
| 4 | `dataset_manager/dataset.py:470` | error-handling | | `if self.data is None: return` silently passes validation on an unloaded dataset — contradicts the PR's own fail-loud thesis and masks a "dataset not loaded" ordering bug. A truly-empty loaded dataset is `[]` (already a no-op via the loop), not `None`; `EmptyDataset` is never actually instantiated. Distinguish: raise on `None` (not-loaded), no-op on `[]`. |
| 5 | `tests/unit/dataset_manager/test_salted_dataset.py:196` | testing | | `match=r"\b1\b"` is a weak/brittle index assertion — `re.search` passes on any word-bounded `1` anywhere in the message and fails for two-digit indices. Anchor to `match=r"sample 1\b"`. |

### 🔵 Consider

- **`dataset.py:270`** — both-keys (`input_tokens` + valid str `prompt`) is now rejected, but the deleted code salted the `prompt` in that case. A prompt-consuming adapter on a mixed-schema dataset loses a salt it used to get. Confirm no dataset+adapter pairing sends `prompt` while `input_tokens` is present.
- **`dataset.py:495` / `execute.py:387`** (2×) — the full dataset is scanned twice (load-time guard + `with_salt` re-scan) over unchanged shared data. Cold-path and fine, but for the 50k+-sample corpora this project targets, add a one-line comment marking the early call as the intentional pre-spawn fail-fast so a future reader doesn't "dedupe" it away.
- **`dataset.py:477`** — the error names `sample {i}`, an index into post-transform/filtered loaded order, not the user's source-file line; a JSONL with column filters won't map cleanly. Clarify "index into loaded order."
- **Testing** — the online/agentic path and `accuracy_only` (dataloader `None` → skip) are uncovered; pin the intended skip so a refactor can't start validating a `None` dataloader.

### 🧹 Code quality

- **`dataset.py:464` & `:490`** — docstrings narrate removed behavior ("rejected **rather than skipped**", "fails here **rather than silently issuing unsalted prompts**"). Per AGENTS.md ("describe current state, not development history"), reframe positively and drop the "rather than …" clauses.
- **`dataset.py:506`** — `assert self._salt_rng is not None` vanishes under `python -O` → a bare `AttributeError` on the warmup path. For a PR about making salt failures loud, a raised error fits better (same applies to the pre-existing `assert self.data is not None` on the hot access path).

---

_Convergence: all three reviewers independently landed on the salt-default-`True` footgun (#1) — the one decision worth the author's attention; the rest is polish._

src/inference_endpoint/dataset_manager/dataset.py
arekay-nv: 1. - not yet, we can do that but the goal is to eliminate kv-cache reuse between the warmup and performance by changing the prefix of the user prompt. For something like gpt-oss, we need a haromize aware salting mechanism. 
2. That is the long term plan - to have each adapter add sensible salt. For now, with the text "prompt" column datasets, we can inject salt, but others require understanding the structure and picking valid tokens.

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -257,6 +258,24 @@ def load_from_huggingface(
...
+def _can_salt(sample: Any) -> DatasetValidationError.Reason | None:
+    """Return the Reason a sample cannot be salted, or None if it can.
+
+    Salt requires a dict sample with a str 'prompt' and no 'input_tokens' (which
+    adapters send verbatim, so a salted 'prompt' would not reach the server).
+    """
+    Reason = DatasetValidationError.Reason
+    if not isinstance(sample, dict):
+        return Reason.TYPE_MISMATCH
+    if "input_tokens" in sample:
```
viraatc, line 276: `_can_salt()` misses another shadowing case. The default OpenAI adapter preserves `messages` alongside `prompt`, then `to_endpoint_request()` prefers `messages` over `prompt` (`openai_msgspec_adapter.py:151-152`). A sample containing both fields therefore passes validation, `_apply_salt()` changes only `prompt`, and the unchanged `messages` payload is sent.

This makes `salt=True` silently fail to bust the cache. Please either reject samples containing `messages` with a dedicated reason or salt the authoritative `messages` payload, and add an adapter-level regression test.

src/inference_endpoint/config/schema.py
```diff
@@ -759,10 +759,19 @@ class WarmupConfig(BaseModel):
...
-            help="Prepend a unique random hex salt to each warmup prompt",
+            help=(
+                "Prepend a unique random hex salt to each warmup prompt. Requires "
+                "text-'prompt' samples; enabling it on a pre-tokenized "
+                "('input_tokens') dataset is a hard error."
+            ),
         ),
     ] = Field(
-        True, description="Prepend a unique random hex salt to each warmup prompt"
+        False,
```
viraatc, line 769: Should we emit a warning when warmup is enabled without salt?

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -257,6 +258,29 @@ def load_from_huggingface(
...
+        return Reason.INPUT_TOKENS_SHADOWING
+    if "messages" in sample:
+        return Reason.MESSAGES_SHADOWING
+    if "prompt" not in sample:
+        return Reason.PROMPT_MISSING
+    if not isinstance(sample["prompt"], str):
```
nv-alicheng, line 284: [Codex + Claude] **medium (regression / api-contract)**: This `str`-only check makes multimodal **list-form** prompts hard-error at load with `PROMPT_TYPE_MISMATCH`. Both reviewers independently verified that list prompts were *previously salted correctly* by prefixing the first `type == "text"` part (covered by the now-deleted `test_multimodal_list_prompt_first_text_part_is_salted` / `test_multimodal_image_first_text_at_index_1_is_salted`) — they were **not** silently skipped. So the PR's "salt was silently skipped" premise doesn't hold for the list case: this removes tested, working behavior, and any existing `warmup.salt: true` VLM/multimodal config now aborts at setup.

Two options:
1. If dropping multimodal salt is intended, call it out as a breaking change and give list prompts a **distinct, actionable** `Reason` — the current message ("sample 'prompt' is not a str") is misleading, since list prompts *can* be salted.
2. If unintended, restore the first-text-part salting branch for list prompts.

src/inference_endpoint/commands/benchmark/cli.py
```diff
@@ -67,9 +67,14 @@ def _run(
                 f"{'.'.join(str(p) for p in err['loc'])}: {err['msg']}"
                 for err in e.errors()
             )
-            raise DatasetValidationError(f"Invalid --dataset: {msgs}") from e
+            # --dataset parse failures aren't yet mapped to a specific Reason.
```
nv-alicheng: [Code-Quality] **low**: `# --dataset parse failures aren't yet mapped to a specific Reason.` narrates future work without a tracking issue, which the project's comment guidelines disallow. Either state the current fact (dataset-string parse failures have no specific `Reason`, hence `UNSPECIFIED`) and drop "yet", or attach a tracking issue link. Same pattern in both `except` arms.

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -440,55 +464,62 @@ def load_sample(self, index: int) -> Any:
             data = self._apply_salt(data)
         return data
 
+    def validate_saltable(self) -> None:
+        """Raise if any loaded sample cannot be salted.
+
+        salt requires a dict sample with a text ('str') 'prompt' and no
+        'input_tokens' (adapters send those verbatim, so a salted 'prompt' would
```
nv-alicheng: [Claude] **low (documentation)**: This docstring says salt requires "no `input_tokens`" only, but `_can_salt` (and the `Reason` enum) also reject `messages` (`MESSAGES_SHADOWING`). The docstring is incomplete — mention `messages` alongside `input_tokens` so the contract matches the code.

nv-alicheng (commented):

Reviewed by: **Codex** + **Claude** + **Code-Quality** | Depth: **standard**

Found **3** issues (1 cross-confirmed by two reviewers). Overall a clean, well-tested refactor — typed `Reason` enum, fail-fast placement, and the collapsed `_apply_salt` are sound.

| 1 | `dataset_manager/dataset.py` | 279 | medium | regression/api-contract | **Codex + Claude** | Multimodal list-form prompts now hard-error at load; list prompts were previously salted (first text part), not silently skipped — tested behavior removed. |
| 2 | `commands/benchmark/cli.py` | 70 | low | code-quality | Code-Quality | Comment narrates future work ("aren't yet mapped") without a tracking issue. |
| 3 | `dataset_manager/dataset.py` | 471 | low | documentation | Claude | `validate_saltable` docstring omits `messages` (code also rejects it). |

**#1 is the decision point:** the "silent skip" justification doesn't hold for list-form prompts — confirm the multimodal-salt drop is intended (and give list prompts a distinct `Reason`), or restore the branch.

src/inference_endpoint/config/schema.py
```diff
@@ -759,10 +759,19 @@ class WarmupConfig(BaseModel):
         bool,
         cyclopts.Parameter(
             alias="--warmup-salt",
-            help="Prepend a unique random hex salt to each warmup prompt",
+            help=(
```
nv-alicheng: **Doc-string duplication here: `cyclopts.Parameter(help=...)` vs `Field(description=...)`**

This `salt` field (and `drain` just below) sets the same string twice — once as cyclopts `help=`, once as `Field(description=...)`. That duplication is avoidable.

**Behavior (verified against cyclopts 4.10.0):** cyclopts reads `Field(description=...)` **natively** for `--help`. When both are set, `help=` wins; delete `help=` and cyclopts falls back to `description` → **identical CLI output**. And `scripts/regenerate_templates.py` (`_collect_comments._help`) already reads `description` *first*, `help` only as fallback. So `Field(description=...)` can be the single source of truth for both the CLI `--help` text and the YAML template comments.

**Proposal:** make `Field(description=...)` canonical and drop the redundant `help=`. Keep `cyclopts.Parameter(...)` only where it carries CLI-only metadata (`alias`, `negative`, `name`, `show`, `required`) — those have no Pydantic equivalent. No `@auto_cyclopts`-style decorator needed: cyclopts already does the injection, so this is *less* machinery, not more. (Avoid the bare attribute-docstring form — Pydantic doesn't put it in `FieldInfo.description`, so the template generator would lose the comment.)

**Scope on HEAD (`6d28c00`):** 16 fields set both `help` + `description` → **10 identical** (safe to collapse), **6 silently divergent**; plus 17 help-only and 50 description-only (already single-source).

Identical (just drop `help=`): `ModelParams.seed`, `WarmupConfig.{enabled,n_requests,salt,drain,warmup_random_seed}`, `DrainConfig.{warmup_timeout_s,performance_timeout_s,accuracy_timeout_s}`, `ProfilingConfig.engine`.

⚠️ **Silently divergent — the CLI `--help` and the YAML comment already say different things** (single-sourcing prevents exactly this drift; decide per field whether the split is intentional):

| field | `help=` (CLI) | `Field(description=)` (YAML/schema) |
|---|---|---|
| `EndpointConfig.endpoints` | Endpoint URL(s) | Endpoint URL(s). Must include scheme, e.g. 'http://host:port'. |
| `EarlyStoppingConfig.enabled` | Report MLPerf early-stopping percentile estimates for TTFT/TPOT/latency | Early-stopping percentile estimates (default on) |
| `Settings.service_ready_timeout_s` | Seconds to wait for metrics/event-logger services to start | Seconds to wait for metrics-aggregator/event-logger services to become ready. |
| `ProfilingConfig.urls` | Override URL(s) for profiler triggers; defaults to endpoint_config.endpoints | URL(s) the profiler start/stop triggers are derived from. When None, derived from endpoint_config.endpoints instead… |
| `DrainConfig.metrics_drain_timeout_s` | …Set to 0 to wait indefinitely. Increase for very large datasets… | …after ENDED (default: 0 = unlimited). An incomplete drain is surfaced via n_pending_tasks > 0, never silently dropped. |
| `DrainConfig.metrics_tokenizer_workers` | …0 defers all tokenization to the end-of-run drain, which always uses the auto-sized sharded pool. | …(default: 2; 0 = defer everything to the end-of-run drain). |

Mostly pre-existing, not blocking — raising it here because this PR edits the `salt` help text in both places, adding two more duplicated fields to the pile. Could be this PR or a follow-up.

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -257,6 +258,29 @@ def load_from_huggingface(
     return ds[split].to_pandas()
 
 
+def _can_salt(sample: Any) -> DatasetValidationError.Reason | None:
```
nv-alicheng: nit: I feel like this naming is inverted:
_can_salt implies truth-y value if it can, but here, None is the 'success' state, and Reasons (truth-y) are failure states.

Rename to `_check_unsaltable`

src/inference_endpoint/dataset_manager/dataset.py
```diff
@@ -440,55 +464,62 @@ def load_sample(self, index: int) -> Any:
...
+
+        Raises:
+            DatasetValidationError: naming the first offending sample. The index
+                is into the loaded, post-transform sample order, not the source
+                file line.
+        """
+        assert self.data is not None, "Dataset not loaded. Call load() first."
+        for i, sample in enumerate(self.data):
+            reason = _can_salt(sample)
+            if reason is not None:
```
nv-alicheng:
```
if not (reason := _check_unsaltable(sample)):
    raise ...
```

nv-alicheng (approved):
Approved, but comments need to be addressed.

src/inference_endpoint/config/templates/concurrency_template_full.yaml
```diff
@@ -93,7 +93,7 @@ settings:
   warmup:
     enabled: false  # Enable warmup phase before performance run
     n_requests: null  # Warmup request count (None = full dataset once)
-    salt: true  # Prepend a unique random hex salt to each warmup prompt
+    salt: false  # Prepend a unique random hex salt to each warmup prompt. Requires text-'prompt' samples; enabling it on a pre-tokenized ('input_tokens') dataset is a hard error.
```
nvzhihanj, line 96: Changing the default behavior to false might fail a lot of the submission silently (simply because the submitter forgot to turn them on). 
Can we warm up with random token if the salt cannot be done, or some other alternative approach (e.g. add random tokens)? Based on the endpoints v0.7 and inference 6.1 I am really worried a non submit-compatible default flag will invalidate many submissions

arekay-nv: By default, both warmup and warmup salting are disabled. Enabling warmup alone leaves salting disabled; the run logs a warning and proceeds with unsalted warmup requests. A hard setup error occurs only when both warmup and salt are enabled and the loaded dataset contains a sample the implementation cannot safely salt. The bug this PR addresses is that, previously, requesting salt could still result in an unsalted outgoing request—sometimes silently and sometimes with only a warning. 
This change would ensure that warmup with salting is either observed correctly or flagged as an error.

src/inference_endpoint/dataset_manager/dataset.py
arekay-nv: Thanks for the suggestion - definitely better.

src/inference_endpoint/config/schema.py
arekay-nv: Fixed the one change made here, created issue #451 to track the remaining.

src/inference_endpoint/config/templates/concurrency_template_full.yaml
nvzhihanj: Yes, that part makes sense. But for submission, only a warmup with SALT will ever be accepted for single-turn/agentic LLM workload. Will "SALT by default enabled" work? Or can a ruleset force that to be true? Thinking about reducing human mistakes and amount of context knowledge here

This is given that all future LLM/VLM workload will use text prompt whenever possible

arekay-nv: Agree on reducing the potential for mistakes. 

For submission we should only allow warmup with salting, but that can change if we want to benchmark a scenario where all the requests share the same prefix (system prompt) and want to capture the performance in that scenario. This will be captured in the benchmark specification via rules and enforced by the system.

As we move towards task specific datasets, the dataset will encode how to salt it given a random seed, and that will be more robust instead of relying on a general salting mechanism. So, the scenario above will be handled by either salting the system prompt (fill cache invalidation) or salting the user prompt (system prompt cached, user prompt invalidated).

src/inference_endpoint/dataset_manager/dataset.py
arekay-nv: Updated the description.

viraatc (approved):

src/inference_endpoint/exceptions.py
```diff
@@ -37,7 +39,58 @@ class InputValidationError(CLIError):
...
+        self,
+        reason: "DatasetValidationError.Reason",
+        detail: str | None = None,
+    ) -> None:
+        self.reason = reason
+        self.detail = detail
+        message = reason.value if detail is None else f"{reason.value}: {detail}"
+        super().__init__(message)
+
+    def __reduce__(
```
viraatc: `__reduce__` is compensating for `BaseException.args` not matching this exception’s constructor: `__init__` takes `(reason, detail)`, but `args` contains only the formatted message. 

Should we instead pass `(reason, detail)` to `super().__init__()` and implement `__str__()` for the human-readable message? Python’s default exception machinery then handles copy/deepcopy/pickle without bespoke serialization, allowing us to delete `__reduce__` and simplify the pickle-only test matrix.

PR #407, opened 2026-07-12
fix: reserve ephemeral-port headroom for the auto max-connections default

nvzhihanj:
## Problem

`client.max_connections` defaults to `-1`, which resolves to the **entire**
ephemeral-port range (`system_maximum_ports × distinct_endpoints`). When
connections are established in a burst — e.g. `max_throughput` at start-up — the
pool tries to bind every ephemeral port at once. The OS holds recently-closed
ports in `TIME_WAIT`, so the tail of `bind()` calls intermittently fails with
`EADDRNOTAVAIL` ("Cannot assign requested address"). The run recovers, but only
after a burst of dropped/retried connections; it's worst when the client shares a
host with other socket activity.

A connection pool should never try to claim 100% of the ephemeral range — that
leaves nothing for `TIME_WAIT` recycling, the OS, or other processes on the box.

## Fix

Resolve the auto (`-1`) limit to a fraction of the budget instead of the whole
range, so the zero-config default reserves headroom:

- `AUTO_MAX_CONNECTIONS_BUDGET_FRACTION = 0.5` — auto claims half the per-endpoint
  budget. That is still a large pool (thousands of connections per endpoint) and
  continues to scale with the number of distinct endpoints.
- An **explicit** `max_connections` is unchanged and may still use the full
  budget (the over-budget guard is unchanged).

Also removes the now-dead `get_ephemeral_port_limit()` / `get_used_port_count()`
helpers — the static reserve replaces them without reading the racy,
process-global live socket count.

## Why this default

The client should work well without the user having to reason about ephemeral
ports or hand-tune `max_connections`. A static reserve keeps that property: the
default is safe on a busy/co-located host, while workloads that genuinely need
more concurrency add endpoints (the budget scales per distinct destination) or
set an explicit value. Empirically, claiming the full range reliably produced
`EADDRNOTAVAIL` under a bursty `max_throughput` start-up, whereas reserving
headroom eliminated it with no change to server-side batch or throughput.

## Tests

Updated `TestEndpointBudgetScaling` for the reserved-headroom default and added a
case asserting the auto limit leaves headroom while an explicit value may use the
full budget.

---


nvzhihanj:
Close this one in favor of @viraatc 's dynamic fix

PR #411, opened 2026-07-14
fix(client): pace connection establishment; auto warmup claims 25% of the port budget

viraatc:
## Problem

Since #386, the full-auto config (`max_connections=-1`, `warmup_connections=-1`) resolves the pool ceiling to the entire ephemeral-port budget (`range × distinct_endpoints`, ~28k on default Linux). Auto warmup derives from that ceiling (`connections_per_worker // 2`), and `ConnectionPool.warmup()` fires the whole count as one unbounded `asyncio.gather` — so a default run opens **~half the port range as one simultaneous connect burst at t=0** (~14k SYNs to a single endpoint). The `max_throughput` t=0 wave can grow the pool the same unbounded way through `acquire()`.

That burst is what regressed real runs: it can overflow the server accept queue (SYN drops → seconds of RTO retransmits) and transiently exhaust client ephemeral ports (`EADDRNOTAVAIL`), and a failed connect currently surfaces as a failed query.

An alternative mitigation was proposed in #407 (halve the auto `max_connections`). That helps empirically, but it works by indirectly halving the warmup burst: the ceiling only counts live pool members (`established + creating`), so it cannot bound TIME_WAIT/port pressure in principle — and it permanently halves the client's steady-state connection roofline (silently: excess demand just queues in the pool) to fix a startup-time event. It also loses its protection the moment someone sets an explicit `max_connections`, since warmup scales off it.

## Fix

Keep the full auto ceiling; fix the burst directly:

1. **Pace connection establishment** — `ConnectionPool` bounds in-flight `connect()` calls with a connect limiter (`MAX_CONCURRENT_CONNECTS = 128` per pool). This covers **both** warmup and runtime pool growth, and never touches the pooled-connection reuse path. Connects complete in ~1 RTT, so establishment now proceeds in fast waves instead of one flood — total warmup time is barely affected. `_creating` is reserved before waiting on the limiter so queued acquires can't overshoot the `max_connections` ceiling.
2. **Bound the auto warm set** — when both `max_connections` and `warmup_connections` are `-1`, warmup resolves to **25% of the port budget** (`AUTO_WARMUP_BUDGET_FRACTION`) instead of half the pool (= half the port range). The ceiling stays at the full budget; connections beyond the warm set are opened on demand at runtime (paying the connect cost lazily, now paced). Warming far beyond steady-state concurrency only wastes ports — the LIFO idle stack means excess warm connections sit established at the bottom holding ephemeral ports until shutdown — and inflates the startup burst.
3. **Cleanup** — removes the dead `get_used_port_count()` / `get_ephemeral_port_limit()` helpers (unused since #386); fixes `docs/ENDPOINT_CLIENT.md`'s auto-configuration section, which still described the pre-#386 live-socket subtraction and a wrong 90% figure for `min_required_connections` (code: 12.5%); pins the `regenerate-templates` pre-commit hook to the project env via `uv run --no-sync` (a bare `python` entry resolves whatever is first on PATH and can regenerate templates against a sibling checkout's schema — which briefly happened on this very branch and was reverted in `09fa7b95`).

Unchanged behavior:

- Explicit `max_connections` and explicit `warmup_connections` values are untouched (over-budget guard unchanged).
- With an explicit `max_connections`, auto warmup remains 50% of the pool (worker-side).
- No change to the request hot path; the limiter is only on the pool-growth path.
- YAML config templates are unchanged (verified byte-for-byte against `main` and round-tripped through `BenchmarkConfig.from_yaml_file`).

## Notes

- On a busy/shared host, actual port contention still surfaces at `connect()` time as an `OSError` with no automatic retry — a bounded EADDRNOTAVAIL retry in `_create_connection` is a natural follow-up to turn transient port pressure into backpressure instead of a failed query (MLPerf run-validity relevance).
- Host-wide simultaneous connects are bounded at `128 × num_workers` (e.g. ~2k with 16 workers); the constant is deliberately conservative and easy to revisit with measured errno counts.

## Tests

- `TestAutoWarmupResolution` (new): full-auto resolves warmup to 25% of the budget, scales with distinct endpoints, explicit/disabled warmup preserved, `-1` sentinel survives when `max_connections` is explicit.
- `test_connect_pacing_bounds_inflight_creates` (new): against the real echo server, peak in-flight `connect()` never exceeds the limiter during warmup.
- Updated `TestEndpointBudgetScaling` guard test (dead-helper patch removed).
- `tests/unit/endpoint_client/`: 79 passed. Template round-trip + template tests: 11 passed. Full unit suite: no new failures vs `main` (an identical pre-existing set of xdist-parallel failures in `metrics_aggregator` tests reproduces on clean `main`).
- Hardened hook verified against the failure mode: with a PATH whose `python` imports a sibling checkout, `pre-commit run regenerate-templates --all-files` now produces zero template diff.
- `pre-commit run --all-files` green.


nvzhihanj:
Tested this branch (rebased onto current `main`) on a **co-located** setup — client and server on the same node, single `localhost` endpoint — under offline `max_throughput`. Sharing a data point since this is exactly the port-exhaustion case #386 introduced.

**The pacing + bounded warmup clearly help.** Full-auto (`max_connections=-1`, `warmup_connections=-1`) went from the large majority of requests failing with `Cannot assign requested address` on the pre-fix build down to a **small single-digit percentage** here — a big improvement, and the single t=0 SYN burst is gone.

**But it doesn't fully eliminate the client-side exhaustion for a single low-latency endpoint.** The remaining failures are all a ~1–2 s burst at t=0. With `max_connections=-1` the ceiling is still the whole ephemeral range, and for a `localhost` (sub-ms RTT) endpoint the 128-permit semaphore doesn't slow establishment enough — the pool reaches the full-range ceiling in <1 s and exhausts client ports. A failed connect surfaces as a failed query (not retried), so those samples are lost and the run is marked partial.

Adding an explicit `max_connections=16384` (a ceiling *below* the ephemeral range) on top of this branch → **0 failures, no throughput regression**. So the pacing (protects the server accept queue / kills the SYN flood) and a bounded ceiling (protects the client's ephemeral ports) look like they address **different halves** of the problem and are complementary, rather than one subsuming the other.

Might be worth also bounding the auto ceiling below the full range (a fraction, as the earlier #407 sketched), or making the pacing time-based rather than concurrency-based, or retrying a failed connect — so the zero-config default is clean even for a single low-latency endpoint under `max_throughput`.

Overall a solid improvement — just flagging that single-endpoint low-RTT `max_throughput` still needs a ceiling bound to reach zero errors. Happy to share details offline.

arekay-nv (approved):
Thanks @viraatc.

viraatc:
> Overall a solid improvement — just flagging that single-endpoint low-RTT max_throughput still needs a ceiling bound to reach zero errors. Happy to share details offline.

thanks for testing! will look into it and address in a followup.
Merging this to unblock new container build.

nvzhihanj (approved):

PR #412, opened 2026-07-14
Add License Scanning and Reporting for Endpoints Changes on Main.

arav-agarwal2:
## What does this PR do?

As part of the rollout for our AI coding policy, we'd like to add a small licensing checker to this codebase. It shouldn't block anything - it just reports changes on an action report we can review asynchronously. 

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally
- [X] Manual testing completed

## Checklist

- [x] Pre-commit hooks pass


.github/workflows/scanoss.yml
```diff
@@ -0,0 +1,207 @@
...
+            scanoss-py scan "${KEY_ARG[@]}" \
+              --settings .github/scanoss.json \
+              --apiurl https://api.osskb.org \
+              --files-from changed_files.txt \
+              -o "$OUT" || SCAN_RC=$?
+          else
+            echo "No changed files to scan (delta empty or all filtered)."
+          fi
+          [ -s "$OUT" ] || echo '{}' > "$OUT"
+          echo "SCAN_RC=$SCAN_RC" >> "$GITHUB_ENV"
```
arekay-nv, line 79: **Silent "clean" on scan failure** — flagged by all three reviewers (Codex P2, Grok high, Claude). `SCAN_RC` is captured and exported to `$GITHUB_ENV`, but no later step reads it. On a failed scan (bad/expired key, quota exhausted, API or CLI error) the `[ -s "$OUT" ] || echo '{}'` fallback leaves `results.json` as `{}`, so the Summary prints "no components detected in this delta" — indistinguishable from a genuinely clean scan. For a license/BOM check, a failed run that looks compliant is the worst failure mode. Keep it non-blocking, but make failure visible:

```suggestion
          [ -s "$OUT" ] || echo '{}' > "$OUT"
          if [ "$SCAN_RC" -ne 0 ]; then
            echo "::warning title=SCANOSS scan failed::scanoss-py exited $SCAN_RC; the report below may be incomplete."
            echo "SCAN_FAILED=1" >> "$GITHUB_ENV"
          fi
          echo "SCAN_RC=$SCAN_RC" >> "$GITHUB_ENV"
```

Then have the report builder print a "⚠️ scan failed — results incomplete" banner when `SCAN_FAILED=1`, so the Summary can't be misread as clean.

.github/workflows/scanoss.yml
```diff
@@ -0,0 +1,207 @@
...
+            : > changed_files.txt
+          else
+            git diff --name-only --diff-filter=ACMR "$BEFORE" "$AFTER" > changed_files.txt
+          fi
+          echo "Merged delta ($(wc -l < changed_files.txt) file(s), before scanoss.json skip filters):"
+          cat changed_files.txt || true
+          SCAN_RC=0
+          if [ -s changed_files.txt ]; then
+            KEY_ARG=()
+            [ -n "${SCANOSS_API_KEY:-}" ] && KEY_ARG=(--key "${SCANOSS_API_KEY}")
```
arekay-nv, line 65: **API key on the process argv** (Grok, Claude). `KEY_ARG=(--key "${SCANOSS_API_KEY}")` puts the secret in argv, readable via `/proc/<pid>/cmdline` for the scan's lifetime. Recent `scanoss-py` reads `SCANOSS_API_KEY` from the environment (already exported in this step's `env:`), so you can drop the `--key` arg and the `KEY_ARG` array entirely. If you keep `--key`, confirm scanoss never echoes its argv. Exposure is bounded (ephemeral single-tenant runner), hence low severity.

.github/workflows/scanoss.yml
```diff
@@ -0,0 +1,207 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install scanoss-py
+        run: pip install scanoss
```
arekay-nv, line 37: **Pin `scanoss`** (Grok, Claude). `pip install scanoss` installs whatever PyPI serves at run time — non-reproducible, and a breaking/compromised release silently changes scan/convert/inspect behavior (and the `scanoss.winnowing` import the quota probe needs). Pin to the version you validate against, e.g. `pip install scanoss==<x.y.z>`, ideally via a constraints/requirements file.

.github/workflows/scanoss.yml
```diff
@@ -0,0 +1,207 @@
...
+
+jobs:
+  scanoss:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
arekay-nv, line 27: **Optional: SHA-pin the actions.** `actions/checkout@v4` (also `setup-python@v5`, `upload-artifact@v4`) are floating major tags. For a workflow that handles a secret on `main`, pinning to full commit SHAs (version in a trailing comment) stops a moved tag from changing runner behavior without a diff. Low priority — all three are first-party GitHub actions.

.github/workflows/scanoss.yml
```diff
@@ -0,0 +1,207 @@
...
+
+          # ---- current file-scan quota (per-key X-Ratelimit-* headers; needs the key) ----
+          key = os.environ.get("SCANOSS_API_KEY", "")
+          if key:
+              try:
+                  import requests
+                  from scanoss.winnowing import Winnowing
+                  wfp = Winnowing().wfp_for_contents(".github/scanoss.json", False, open(".github/scanoss.json", "rb").read())
+                  rid = str(uuid.uuid4())
+                  r = requests.post(
```
arekay-nv, line 143: **Quota probe spends the quota it measures** (Grok, Claude). This POSTs a winnowed wfp to `/scan/direct` every run *only* to read the `X-Ratelimit-*` headers — consuming one file-scan against the same per-key quota it reports, and adding a `requests` + `scanoss.winnowing` dependency for a Summary-only cosmetic. Consider dropping it, or reading the rate-limit headers off the main scan's HTTP response instead of issuing a separate billable request.

.github/scanoss.json
```diff
@@ -0,0 +1,49 @@
...
+        "purl": "pkg:github/mlcommons/endpoints"
+      },
+      {
+        "purl": "pkg:github/nvidia/tensorrt-edge-llm"
+      },
+      {
+        "purl": "pkg:github/nvlabs/cosmos-policy"
+      }
+    ],
+    "exclude": [
```
arekay-nv, line 43: **Confirm the BOM `include` scope** (Grok, Claude). `include` lists `pkg:github/nvidia/tensorrt-edge-llm` and `pkg:github/nvlabs/cosmos-policy` alongside this repo. Declaring other orgs' repos as BOM members tells SCANOSS their matches are *declared*, which can mask undeclared/copyleft findings in this tree. If `mlcommons/endpoints` doesn't vendor those, restrict `include` to this repo:

```suggestion
    "include": [
      {
        "purl": "pkg:github/mlcommons/endpoints"
      }
    ],
```

Keep the extra entries if the cross-repo declaration is intentional.

arekay-nv (commented):
## Review-council: SCANOSS license-scan workflow

Three independent reviewers (Codex `gpt-5.5`, Grok 4.5, Claude) looked at this PR. **The workflow is well-built and security-conscious** — correct `push`-to-`main` trigger (not `pull_request_target`), `permissions: contents: read`, and defensive shell (handles the all-zeros first-push SHA, empty deltas, and a missing key). No critical/high issues: `${{ }}` interpolation into `run:` is limited to commit SHAs + `env.REPORTS_DIR` (not injectable), and the `data.items()` results walk matches SCANOSS's output shape.

Findings below are reliability/hardening, most-severe first. The **top one (flagged by all three reviewers)** is the only thing I'd treat as merge-worthy — a failed scan currently reports as "clean." The rest are optional. Nothing here blocks the PR.

.github/scanoss.json
arav-agarwal2: These were found in the scan - so rejecting suggestion.

.github/workflows/scanoss.yml
arav-agarwal2: This is fair - however it's challenging to do within the context of the scanoss pipeline. I'll drop it and upgrade this issue when I get time to do so and if it makes sense to do so.

.github/workflows/scanoss.yml
arav-agarwal2: Yea - as these are major tools, I think I can safely ignore this for now. Let me know if this is needed.

.github/workflows/scanoss.yml
arav-agarwal2: I've confirmed scanoss doesn't echo it's output, so we should be fine.

.github/workflows/scanoss.yml
```diff
@@ -0,0 +1,211 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install scanoss-py
+        run: pip install scanoss
```
arekay-nv, line 37: Can we pin the version for scanoss here

.github/workflows/scanoss.yml
```diff
@@ -0,0 +1,211 @@
...
+          w("|---|---|---|---|---|")
+          for fpath, comp, pct, lics, flag in rows:
+              w(f"| `{fpath}` | {comp} | {pct} | {lics} | {flag} |")
+          if not rows:
+              w("| _no components detected in this delta_ | | | | |")
+          w()
+
+          with open(os.environ["GITHUB_STEP_SUMMARY"], "a") as f:
+              f.write("\n".join(out) + "\n")
+          PY
```
arekay-nv, line 203: Can we lift this out to a separate file `.github/scripts/scanoss_report.py`

.github/workflows/scanoss.yml
```diff
@@ -0,0 +1,211 @@
...
+                  return "copyleft"
+              if n in PERMISSIVE:
+                  return None
+              return "review"   # not clearly permissive — worth a look
+
+          w("## SCANOSS — merged-delta report")
+          w()
+          w(f"Commit `{os.environ.get('GITHUB_SHA','')[:12]}` on `{os.environ.get('GITHUB_REF_NAME','')}`")
+          w()
+
```
arekay-nv, line 134: We should also be checking `SCANOSS_RC` to see if the tool ran to differentiate the call failing versus a clean scan.

.github/scanoss.json
```diff
@@ -0,0 +1,49 @@
...
+    "include": [
+      {
+        "purl": "pkg:github/mlcommons/endpoints"
+      },
+      {
+        "purl": "pkg:github/nvidia/tensorrt-edge-llm"
+      },
+      {
+        "purl": "pkg:github/nvlabs/cosmos-policy"
+      }
```
arekay-nv, line 41: Are these two needed?

.github/scanoss.json
arekay-nv, line 43: Is "exclude" a valid category? Didn't find it [here](https://docs.scanoss.com/en/latest/poc/process-integrations/github-actions#step-4-create-your-scanoss-json-file)

PR #414, opened 2026-07-14
feat: optional VBench (WAN 2.2) accuracy scorer in Dockerfile.dev (PROVISION_VBENCH)

wu6u3tw:
## What does this PR do?
The image built from `scripts/Dockerfile.dev` only `COPY src/`s the package, so the WAN 2.2 VBench accuracy scorer — which lives at `examples/09_Wan22_VideoGen_Example/accuracy/`, outside `src/` — is not present. As a result `inference-endpoint benchmark --mode acc|both` for the wan22 videogen example has no scorer available in the image.

This adds a **`PROVISION_VBENCH` build arg, defaulting to `1` (on)** — consistent with the existing `PROVISION_DSR1=1` default, which already bakes the DeepSeek-R1 accuracy evaluator into the stock image. It materializes the VBench accuracy uv project at `/opt/vbench_accuracy` in its **own isolated `.venv`**, so accuracy runs reuse it via `UV_NO_SYNC=1` (`vbench_project_path=/opt/vbench_accuracy`) — no host checkout, no runtime network. Build with `--build-arg PROVISION_VBENCH=0` to skip it (leaner image with no wan22 accuracy). Also adds `libgl1` + `libglib2.0-0` (opencv runtime the scorer's video reader needs at import).

Because DSR1 is already provisioned by default, defaulting VBench on keeps both bundled accuracy scorers consistent. Note this adds the VBench venv (torch + scoring models) to the default image; pass `PROVISION_VBENCH=0` for a lean build. The DSR1 path is untouched.

**Isolation:** the Dockerfile sets a global `ENV UV_PROJECT_ENVIRONMENT=/opt/venv`. The VBench project pins Python 3.11 + old/heavy deps (torch, decord, vbench 0.1.5 -> transformers 4.33.2 -> tokenizers 0.13.3) that must not leak into the main 3.12 `/opt/venv`, so every `uv` call in the VBench block overrides `UV_PROJECT_ENVIRONMENT` to the subproject's own `.venv` — the same rule the DSR1 block already uses.

**Cross-arch (x86_64 + aarch64):** built on a uv-managed Python 3.11 because tokenizers 0.13.3 has no cp312 wheel; the pyproject is patched to relax `requires-python` and swap `decord` -> `decord2` (decord 0.6.0 has no arm64 wheel). These make the venv install wheel-only — required on aarch64, harmless on x86_64. The base image and uv-managed Python are multi-arch and `uv lock` re-resolves per arch, so one Dockerfile serves both; build the image for the target arch (`docker buildx --platform linux/amd64,linux/arm64`).

## Type of change
- [x] New feature

## Testing
- [ ] `docker build -f scripts/Dockerfile.dev --build-arg PROVISION_VBENCH=1 -t ie-vbench .`
- [ ] In-container import check: `UV_PROJECT_ENVIRONMENT=/opt/vbench_accuracy/.venv uv run --project /opt/vbench_accuracy python -c "import torch, decord, tokenizers, vbench"`
- [ ] Default build (`PROVISION_VBENCH` unset) unchanged; DSR1 path unaffected
- [ ] Cross-arch build on linux/arm64 via buildx

## Checklist
- [ ] Pre-commit hooks pass

### DIFF
--- a/scripts/Dockerfile.dev
+++ b/scripts/Dockerfile.dev
@@ -2,6 +2,10 @@
 # From project root:
 # docker build -f scripts/Dockerfile.dev --build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g) -t inference-endpoint-dev .
 # docker run -v $(pwd):/mnt/inference-endpoint -it --shm-size=512m inference-endpoint-dev bash
+#
+# VBench (WAN 2.2) accuracy scorer is OFF by default (lives under examples/, not
+# src/). To bake it in, build with --build-arg PROVISION_VBENCH=1:
+# docker build -f scripts/Dockerfile.dev --build-arg PROVISION_VBENCH=1 --build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g) -t inference-endpoint-vbench .
 
 FROM python:3.12.11-slim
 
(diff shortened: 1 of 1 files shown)


wu6u3tw:
Folding this Dockerfile change into #403 (which fixes the VBench scorer bugs it depends on), so the packaging and the runtime fixes land together. Closing in favor of #403.

PR #417, opened 2026-07-16
feat(metrics): opt-in MLPerf early-stopping percentile estimates (LoadGen parity)

nvzhihanj:
## What

Adds an **on-by-default MLPerf-LoadGen-style early-stopping percentile estimate** to the tail-latency metrics (`ttft` / `tpot` / `latency`) for scenarios that measure percentile latencies (Server/Poisson, fixed concurrency) — closing a parity gap with LoadGen (`loadgen/early_stopping.cc`; mlcommons/inference#2188, mlcommons/inference#1095; MLPerf Inference paper arXiv:1911.02549). LoadGen's fixed-sample requirement (~270k queries for Server p99) is impractical for slow modern workloads (LLM, and especially T2V at minutes per query); early stopping replaces it with a data-dependent binomial-confidence test so short runs surface an honest, conservative tail estimate instead of a silently under-reported empirical percentile.

- `metrics/early_stopping.py` — pure port of the LoadGen binomial-confidence math (continued-fraction regularized incomplete beta, no scipy), validated against LoadGen reference values (e.g. `h_min(t=0, p99) = 459`; SingleStream-style floors: p90 = 64, p99 = 662).
- **On by default** (⚠️ review question): the feature is non-invasive — computed once at run COMPLETE from data the aggregator already keeps (the exact path sorts that array once, in place, and shares it with the percentile-grid computation), hot path untouched, and the output field is purely additive — so every run gets loadgen-parity tail certification without knowing the flag exists. `settings.early_stopping.enabled: false` / `--no-early-stopping` is the single opt-out (for consumers that strictly validate the summary schema). `CONFIDENCE = 0.99` and `TOLERANCE = 0.0` are LoadGen-parity algorithm constants (`results.cc:157-158`), and the percentile targets derive from each series' own report percentile grid (filtered to ≥ p50 — an ES estimate is a tail certification), so the scenario gate percentile is always covered — nothing to tune per config, no way to accidentally weaken the statistical claim.
- The metrics aggregator computes the estimates once at run COMPLETE from the exact sorted series (cold path; hot path untouched). `result_summary.json` gains a compact `early_stopping_percentile` map per metric — keys mirror the `percentiles` grid (≥ p50), values are the conservative estimate or `null` when the run is below that percentile's floor. The rich detail (empirical/n/min_queries/discarded) is INFO-logged by the aggregator and reproducible offline.
- `docs/early_stopping.md` — design, layering, and the multi-percentile marginal-confidence caveat.
- `scripts/early_stopping_estimate_from_events.py` — post-hoc tool that recomputes ES for **any historical run** from its `events.jsonl`. It decodes lines with the product's typed `EventRecord` (the JSONL writer's mirror decoder), so event names, payload shapes, TPOT chunk semantics, tokenizer counting, and the ES math all come from product modules.

Estimate-only by design: no target-latency pass/fail gate and no dynamic mid-run halt (scope in the doc).

## Validation

- **Unit** (suite: 1351 passed): math vs LoadGen reference values, conservative/below-floor behavior, registry → snapshot → report integration, post-hoc script gating/parsing.
- **No-GPU smoke** (echo server, n = 300): the `early_stopping_percentile` map reports estimates for p50–p97 (floors 11–130) and `null` for p99/p99.9 (floors 662/6636) — the short-run behavior the feature exists for.
- **End-to-end** on a DS-R1 B300×8 Server (Poisson) 40k-query run: the in-band blocks **exactly match** an independent post-hoc recomputation from the run's event log (estimate/empirical/n/discarded, all three metrics — including TPOT's tokenization path). At n = 40k the p99 estimate is the expected conservative rank shift (400th → 353rd-highest sample); the ES-vs-empirical delta is small on a well-provisioned run and grows on short/subsampled runs exactly as the binomial math predicts. (Absolute perf figures from the internal run are intentionally omitted.)

## Notes for review

- Draft for early feedback: naming/placement (`settings.early_stopping` vs nesting under a metrics section), the report block shape, and whether `scripts/es_from_events.py` should graduate to a CLI subcommand.
- Planned follow-up: T2V/VLM validation at p90 with fixed concurrency = 1.


src/inference_endpoint/config/schema.py
```diff
@@ -925,6 +925,29 @@ def _validate_url_scheme(cls, v: list[str] | None) -> list[str] | None:
 
 
 @cyclopts.Parameter(name="*")
```
nvzhihanj: [Codex+Claude] **critical / api-contract**: Inserting `EarlyStoppingConfig` here **stole the `@cyclopts.Parameter(name="*")` decorator that belonged to `Settings`** (on `main` the decorator sits directly above `class Settings`; at HEAD `Settings` is undecorated). Consequences, verified at HEAD:
- every documented dotted flag under settings is broken: `--client.warmup-connections` now fails with `Unknown option ... Did you mean "--settings.client.warmup-connections"?` — `tests/integration/commands/test_cli.py` (which uses `--client.warmup-connections`) fails at HEAD (unit suite doesn't catch it);
- because `EarlyStoppingConfig` is itself flattened, `--help` advertises `--settings.enabled` bound to BOTH `warmup.enabled` and `early_stopping.enabled` — parsing it can silently toggle warmup instead of ES.

**Fix**: restore `@cyclopts.Parameter(name="*")` on `Settings` and drop it from `EarlyStoppingConfig` (the field already has the `--early-stopping` alias). Add a CLI-surface unit test so decorator placement can't silently regress.

src/inference_endpoint/metrics/early_stopping.py
```diff
@@ -0,0 +1,220 @@
...
+def _odds(h: int, t: int, p: float, d: float) -> float:
+    # P(<= t over-latency in h + t total | over-latency rate = 1 - p + d)
+    return _betai(h, 1 + t, p - d)
+
+
+def find_min_passing(t: int, p: float, d: float = TOLERANCE, c: float = CONFIDENCE) -> int:
+    """Minimum ``h`` such that ``_odds(h, t, p, d) <= 1 - c``. ``_odds`` decreases in ``h``."""
+    target = 1.0 - c
+    lo, hi = 1, 2
+    while _odds(hi, t, p, d) > target:
```
nvzhihanj, line 218: [Claude] **medium / bug**: `find_min_passing`'s doubling loop has no domain validation or iteration cap. For `p >= 1.0`, `_betai(h, 1+t, 1.0)` returns 1.0 for every `h`, so `while _odds(hi, t, p, d) > target: hi *= 2` **never terminates**; for `c == 1.0` (`target == 0.0`) it terminates only via float underflow and returns a meaningless value. Unreachable in-band (constants only), but `scripts/es_from_events.py --percentiles 1.0` hangs the tool and `--confidence 1.0` prints garbage. **Fix**: raise `ValueError` unless `0 < p < 1` and `0 < c < 1` in `es_percentile_estimate`/`find_min_passing`, and validate the script's CLI args.

src/inference_endpoint/metrics/early_stopping.py
```diff
@@ -0,0 +1,220 @@
...
+
+def es_percentile_estimate(
+    sorted_latencies: Sequence[float],
+    percentile: float,
+    confidence: float = CONFIDENCE,
+) -> EarlyStoppingResult:
+    """Conservative early-stopping percentile estimate over an ascending-sorted latency series."""
+    n = len(sorted_latencies)
+    min_queries = find_min_passing(1, percentile, TOLERANCE, confidence) + 1
+    emp_idx = min(n - 1, max(0, math.ceil(percentile * n) - 1)) if n else 0
```
nvzhihanj: [Claude] **medium / data-integrity**: the block's `empirical` uses `ceil(p*n)-1` while the percentile grid in the same `result_summary.json` uses `np.percentile(..., method="lower")` = `floor(p*(n-1))` (registry `_exact_stat`). They silently diverge by one order statistic whenever `floor(p*(n-1)) != ceil(p*n)-1` — e.g. n=10050/p=0.99 → arr[9949] vs arr[9948]; n=2001 exposes it too. So `early_stopping.empirical` can disagree with `percentiles["99.0"]` two lines above it. **Fix**: derive `empirical` with the grid convention (`floor(p*(n-1))`), and add a test pinning block-empirical == grid value for a non-degenerate n (e.g. 2001).

src/inference_endpoint/metrics/report.py
```diff
@@ -102,6 +102,12 @@ def _series_to_metric_dict(stat: dict[str, Any]) -> dict[str, Any]:
             "counts": [c for _, c in histogram],
         },
     }
+    # Early-stopping estimate block — present only when the feature is enabled
+    # (COMPLETE snapshots for TTFT/TPOT/latency).
+    early_stopping = stat.get("early_stopping")
+    if early_stopping is not None:
+        metric["early_stopping"] = early_stopping
```
nvzhihanj: [Codex+Claude] **medium / bug**: the ES blocks reach `result_summary.json` but nothing renders them on the display path — `Report.display()`/`_display_metric` never reads `early_stopping`, so `report.txt`/console show nothing even when the flag is on, while `docs/early_stopping.md` claims "The text report renders one extra line per metric." **Fix**: render one line per metric in `_display_metric` (the `sufficient=false` warning is the whole point of the feature) — or, second best, correct the doc.

docs/early_stopping.md
```diff
@@ -0,0 +1,92 @@
...
+}
+```
+`sufficient=false` with `estimate=null` means the run had fewer than `min_queries` samples to claim
+that percentile at the requested confidence. The text report renders one extra line per metric.
+
+## Explicitly out of scope
+
+- No `target_latency` pass/fail gate (LoadGen's Server ES path).
+- No dynamic mid-run halt — evaluated once at COMPLETE.
+- Offline/max_throughput scenarios are not tail-latency-bounded, so ES is a no-op there even if enabled.
```
nvzhihanj: [Claude] **low / documentation**: "ES is a no-op there even if enabled" is factually wrong — nothing gates ES by scenario: `execute.py` appends `--early-stopping` for any load pattern and the aggregator attaches the spec to `sample_latency_ns` unconditionally, so an offline run WILL emit blocks (and ttft/tpot when streaming). Reword (computed but not meaningful for throughput-bound scenarios) or implement the gating described.

docs/early_stopping.md
```diff
@@ -0,0 +1,92 @@
+# Early stopping — design
+
+Optional, default-off feature that adds an MLPerf-LoadGen-style **early-stopping percentile
+estimate** to tail-latency metrics (TTFT / TPOT / total latency). It reports a *conservative,
+confidence-backed* percentile alongside the empirical one, so a run with too few samples to trust its
+raw p99/p90 is surfaced honestly instead of silently under-reporting the tail.
+
+It is an **estimate only** — no target-latency pass/fail, no dynamic mid-run halt. See the study /
+gap analysis in the companion `endpoints-early-stopping` repo.
```
nvzhihanj: [Claude] **low / documentation**: "See the study / gap analysis in the companion `endpoints-early-stopping` repo" is a dangling pointer for the upstream audience (violates AGENTS.md "Documentation references — no local-only artifacts"). Link a public URL, inline the key conclusions, or drop the sentence.

scripts/es_from_events.py
```diff
@@ -0,0 +1,269 @@
...
+        batch_deltas.clear()
+        batch_texts.clear()
+
+    with open(path, "rb") as f:
+        for i, line in enumerate(f):
+            if progress and i and i % 200_000 == 0:
+                print(f"  ...{i} events", file=sys.stderr)
+            try:
+                e = _loads(line)
+            except (ValueError, msgspec.DecodeError):
```
nvzhihanj: [Claude] **medium / error-handling**: two robustness gaps for corrupt/untrusted JSONL: (1) undecodable lines are silently swallowed (`except (ValueError, msgspec.DecodeError): continue`) with no counter — a truncated log silently yields smaller series and different ES verdicts (contrast with the tool-call skip, which warns; also `msgspec.DecodeError` is a `ValueError` subclass, listing both is redundant); (2) a line decoding to valid non-dict JSON (`123`, `"str"`) crashes at `e.get(...)` with `AttributeError`, and dict events missing `sample_uuid`/`timestamp_ns` crash with `KeyError`. **Fix**: count+warn on malformed lines, guard `isinstance(e, dict)`, use `.get()` with skip-and-count for required keys.

scripts/es_from_events.py
```diff
@@ -0,0 +1,269 @@
...
+                continue
+            et = e.get("event_type")
+            if et == "session.start_performance_tracking":
+                tracking = True
+            elif et == "session.stop_performance_tracking":
+                tracking = False
+            elif et == "sample.issued":
+                if tracking:
+                    rows[e["sample_uuid"]] = [e["timestamp_ns"], None]
```
nvzhihanj: [Claude] **low / bug**: duplicate `sample.issued` (retry) semantics diverge from the aggregator: this overwrites the row and resets `recv_first` (`rows[uuid] = [ts, None]`), while `MetricsTable._create_row` logs and REUSES the existing row (recv_first preserved). On runs with retries the in-band aggregator records a TPOT the script skips → `--compare` reports spurious MISMATCHes. **Fix**: on duplicate uuid, keep the row and update only the issued timestamp.

scripts/es_from_events.py
```diff
@@ -0,0 +1,269 @@
...
+def es_blocks(values, percentiles, confidence: float) -> list[dict]:
+    """Sorted-input ES blocks, same shape as result_summary.json ``early_stopping``."""
+    values = sorted(values)
+    return [
+        es_percentile_estimate(values, p, confidence).as_dict() for p in percentiles
+    ]
+
+
+def _make_counter(tokenizer_path: str):
+    from inference_endpoint.async_utils.services.metrics_aggregator.token_metrics import (
```
nvzhihanj: [Claude] **low / design+cleanup** (3 small items): (1) function-level import of `token_metrics` violates the repo's no-lazy-imports rule (AGENTS.md) — move to top-level (module import is cheap; the heavy work happens in `load_reference_backend` at call time); (2) `json.load(open(args.compare))` leaks the handle — use a context manager like the `--json` writer does; (3) each series is sorted twice (`es_blocks` sorts a copy internally, then `main` sorts again for `_cross_check`) — sort once and share.

scripts/es_from_events.py
```diff
@@ -0,0 +1,269 @@
...
+    completed = summary.get("n_samples_completed")
+    if completed is not None and len(values) != completed:
+        # ttft/tpot legitimately count fewer (streaming-only / empty-rest skips);
+        # latency should equal n_samples_completed.
+        note = " ** MISMATCH **" if series_name == "sample_latency_ns" else " (info)"
+        print(f"  XCHECK count: ours={len(values)} n_samples_completed={completed}{note}")
+    inband = md.get("early_stopping")
+    if not inband:
+        return
+    if isinstance(inband, dict):  # pre-list single-percentile shape
```
nvzhihanj: [Claude] **low / design**: dead compatibility branch — `if isinstance(inband, dict):  # pre-list single-percentile shape` supports a wire shape that only existed in intermediate commits of this same branch; nothing released ever emits a dict here, and the comment narrates development history (AGENTS.md disallows). Delete the branch.

src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
```diff
@@ -315,6 +324,18 @@ def _exact_stat(self) -> SeriesStat:
             for i in range(len(edges) - 1)
         ]
 
+        # Early-stopping estimates (COMPLETE path only): conservative confidence-backed
+        # bound per configured percentile, all off one sorted raw array. Cold path —
+        # the sort is one-time at run end and each estimate is a few beta evaluations.
+        early_stopping: list[dict[str, float | int | bool | None]] | None = None
+        if self._es_spec is not None:
```
nvzhihanj: [Claude] **low / bug**: the empty-series early-return above (`if self._count == 0: return SeriesStat(...)`) skips this ES logic entirely, so with ES enabled a target series that recorded nothing (e.g. all requests failed → no `ttft_ns`) omits the `early_stopping` key — indistinguishable from the feature being off. `es_percentile_estimate` already handles n=0 cleanly (`sufficient:false, estimate/empirical:null`). **Fix**: emit the blocks in the empty arm too when `self._es_spec is not None` (or document the omission).

src/inference_endpoint/async_utils/services/metrics_aggregator/snapshot.py
```diff
@@ -246,6 +250,9 @@ def _metric_to_dict(m: MetricStat) -> dict:
             for rng, c in m.histogram
         ],
     }
+    if m.early_stopping is not None:
+        series["early_stopping"] = m.early_stopping
```
nvzhihanj: [Claude] **low / design**: every numeric field in `_metric_to_dict` goes through `_scrub_nonfinite` (the dict form feeds `json.dumps(..., allow_nan=False)`), but the ES block is passed through raw. Today's producers can't yield NaN/Inf, but any future float series in `ES_TARGET_METRICS` with a non-finite sample would make the final-snapshot write raise. Cheap fix: scrub `estimate`/`empirical` here.

src/inference_endpoint/config/templates/concurrency_template_full.yaml
```diff
@@ -89,14 +89,16 @@ settings:
     metrics_drain_timeout_s: 0.0  # Wall-clock budget (seconds) to finish tokenizing buffered samples after ENDED (default: 0 = unlimited). An incomplete drain is surfaced via n_pending_tasks > 0, never silently dropped.
     metrics_tokenizer_workers: 2  # In-process tokenizer threads for live (mid-run) ISL/OSL/TPOT (default: 2; 0 = defer everything to the end-of-run drain).
   warmup:
-    enabled: false  # Enable warmup phase before performance run
+    enabled: false
```
nvzhihanj: [Claude] **low / documentation**: all three regenerated templates silently LOST the warmup comment (`enabled: false  # Enable warmup phase before performance run` → bare), and the new `early_stopping.enabled` has no comment either. Root cause: `regenerate_templates.py:_collect_comments` keys inline comments by bare field name and drops them on collision — adding `EarlyStoppingConfig.enabled` collides with `WarmupConfig.enabled`. Fix the generator (qualify comment keys by parent section path) rather than accepting the regression — it will recur for every future `enabled` field.

tests/unit/async_utils/services/metrics_aggregator/test_early_stopping_integration.py
```diff
@@ -0,0 +1,91 @@
...
+            reg.build_snapshot(state=SessionState.COMPLETE, n_pending_tasks=0)
+        )
+        assert "early_stopping" not in _series(d, "ttft_ns")
+        assert "early_stopping" not in _series_to_metric_dict(_series(d, "ttft_ns"))
+
+    def test_live_snapshot_has_no_estimate(self):
+        # ES is COMPLETE-only (needs the exact sorted raw array); LIVE must not carry it.
+        reg = _registry_with_data(EarlyStoppingSpec())
+        d = snapshot_to_dict(
+            reg.build_snapshot(state=SessionState.LIVE, n_pending_tasks=1)
```
nvzhihanj: [Claude] **medium / testing**: coverage gaps in the aggregator wiring: (1) INTERRUPTED terminal snapshots (SIGTERM path, non-exact) — nothing pins that they carry no `early_stopping`; (2) the empty-series arm with ES enabled; (3) only `ttft_ns` (int dtype) is exercised — `tpot_ns` (the only float member of `ES_TARGET_METRICS`) and `sample_latency_ns` never get an ES assertion; (4) nothing asserts block-`empirical` equals the grid percentile for a matching p — a test with n=2001 would have caught the ceil/floor divergence.

tests/unit/scripts/test_early_stopping_estimate_from_events.py
```diff
@@ -0,0 +1,124 @@
...
+        assert s["tpot_ns"] == [(400 - 150) / 2]
+
+    def test_no_tokenizer_skips_tpot(self, tmp_path):
+        s = mod.compute_series(self._fixture(tmp_path), count_tokens=None)
+        assert s["tpot_ns"] == []
+        assert len(s["sample_latency_ns"]) == 3
+
+
+class TestEsBlocks:
+    def test_blocks_use_product_math_and_shape(self):
```
nvzhihanj: [Claude] **low / testing**: everything downstream of `compute_series`/`es_blocks` is unexercised: `main()` (arg parsing incl. the invalid values that hang `find_min_passing`, `--json` writing), `_cross_check` (match/mismatch/count branches), malformed-line handling (undecodable line, valid non-dict JSON, missing keys), and the `tool_call_skipped` warning path. Given the tool audits historical (possibly imperfect) logs, the corrupt-input branches deserve direct tests.

nvzhihanj (commented):

Reviewed by: **Codex** (gpt-5.5, xhigh) + **Claude** | Depth: **thorough**

Posting 14 inline findings (1 critical, 5 medium, 8 low). Verified against HEAD before posting; summary comment follows.

nvzhihanj:

Reviewed by: **Codex** (gpt-5.5, xhigh) + **Claude** | Depth: **thorough** (diff ~1.3k lines)

Found **15** issues across **10** files. All line numbers and claims verified against HEAD (`c720c8c`) before posting; two findings were independently flagged by both reviewers.

| 1 | `config/schema.py` | 927 | api-contract | Both | `EarlyStoppingConfig` stole `Settings`' `@cyclopts.Parameter(name="*")` decorator → every dotted CLI flag broken (`--client.warmup-connections` fails; integration test fails at HEAD) + ambiguous `--settings.enabled` bound to both warmup and ES |

### 🟡 Should Fix (medium)
| 2 | `metrics/early_stopping.py` | 122 | bug | Claude | `find_min_passing` hangs for `p>=1.0`, garbage for `c==1.0` — reachable via the post-hoc script's unvalidated `--percentiles`/`--confidence` |
| 3 | `metrics/early_stopping.py` | 205 | data-integrity | Claude | block `empirical` (`ceil(p·n)−1`) diverges from the percentile grid (`floor(p·(n−1))`, `method="lower"`) in the same JSON |
| 4 | `metrics/report.py` | 109 | bug | Both | ES blocks never rendered in report.txt/console; doc claims they are |
| 5 | `scripts/es_from_events.py` | 107 | error-handling | Claude | silent malformed-line swallowing; non-dict JSON / missing keys crash with raw tracebacks |
| 6 | `tests/.../test_early_stopping_integration.py` | 89 | testing | Claude | INTERRUPTED/empty-series/tpot+latency/empirical-vs-grid branches untested |

### 🔵 Consider (low)
| 7 | `docs/early_stopping.md` | 92 | documentation | "ES is a no-op for offline" is factually wrong — nothing gates by scenario |
| 8 | `docs/early_stopping.md` | 9 | documentation | dangling reference to a non-public companion repo (AGENTS.md rule) |
| 9 | `scripts/es_from_events.py` | 116 | bug | duplicate-ISSUED semantics diverge from aggregator → spurious `--compare` mismatches on retry runs |
| 10 | `scripts/es_from_events.py` | 158 | design | lazy import (AGENTS.md), unclosed `--compare` handle, double sort per series |
| 11 | `scripts/es_from_events.py` | 184 | design | dead dict-shape compat branch for a format that never shipped |
| 12 | `.../registry.py` | 331 | bug | empty target series omits the ES block entirely (looks like feature-off) |
| 13 | `.../snapshot.py` | 254 | design | ES block bypasses `_scrub_nonfinite` → future NaN would break final-snapshot write |
| 14 | `templates/*_full.yaml` | 92 | documentation | regeneration lost the warmup comment (generator comment-key collision) |
| 15 | `tests/unit/scripts/test_es_from_events.py` | 113 | testing | `main()`/`_cross_check`/corrupt-input branches unexercised |

Fixes for all findings are being applied to this branch iteratively.

nvzhihanj:
## Review Council — fixes applied (`166f39a`)

All 15 findings addressed:

| # | Finding | Resolution |
|---|---------|------------|
| 1 | 🔴 `Settings` lost `name="*"` decorator | **Fixed** — decorator restored on `Settings`, `EarlyStoppingConfig` un-flattened. New `tests/unit/commands/test_cli_flag_surface.py` pins the flag namespace (`--client.num-workers` top-level, no `--settings.*` leak, warmup keeps bare `--enabled`, ES stays namespaced). `tests/integration/commands/test_cli.py` passes again (5/5). |
| 2 | `find_min_passing` hangs for p≥1 | **Fixed** — domain validation (`0<p<1`, `0<c<1`, `0≤d<p`) raises `ValueError`; the script also rejects bad `--percentiles`/`--confidence` up front. The hang was reproduced live by the new test before the fix. |
| 3 | `empirical` convention diverges from grid | **Fixed** — now `floor(p·(n−1))` (matches `np.percentile(..., method="lower")`); test pins block-empirical == grid value. |
| 4 | ES not rendered in text report | **Fixed** — `_display_metric` renders an "Early-stopping estimates" section (insufficient percentiles say so explicitly). |
| 5 | script silent/crashy on corrupt JSONL | **Fixed** — malformed lines counted + WARN'd; non-dict JSON and missing `sample_uuid`/`timestamp_ns` skip-and-count instead of crashing. One review sub-claim did **not** survive verification: `msgspec.DecodeError` is *not* a `ValueError` subclass — the original two-exception tuple was correct and is kept (with a comment); the new junk-line test caught this. |
| 6 | aggregator test gaps | **Fixed** — INTERRUPTED pinned ES-free; empty-series blocks; all three target series (incl. float `tpot_ns`) asserted; empirical-vs-grid pin. |
| 7 | doc: offline "no-op" claim | **Fixed** — reworded to match reality (computed, not gating). |
| 8 | doc: companion-repo dangling ref | **Fixed** — removed; conclusion inlined. |
| 9 | duplicate-ISSUED divergence | **Fixed** — aggregator parity (row kept, issue ts refreshed); test mirrors the retry sequence. |
| 10 | lazy import / unclosed handle / double sort | **Fixed** — top-level imports (`token_metrics` is stdlib-only at module level), context-managed `--compare` read, single sort shared by `es_blocks` + `_cross_check`. |
| 11 | dead dict-shape branch | **Fixed** — removed. |
| 12 | empty series omits ES block | **Fixed** — emits `n=0, sufficient:false` blocks on COMPLETE; documented. |
| 13 | ES block bypasses `_scrub_nonfinite` | **Fixed** — estimate/empirical scrubbed like every other numeric field. |
| 14 | templates lost warmup comment | **Fixed at the root** — `regenerate_templates.py` comment keys are now dotted paths (indentation-aware injector); warmup + ES comments coexist, and previously-collided comments (`name`, `samples`, `eval_method`, `drain`) reappear. `--check` mode passes. |
| 15 | script test gaps | **Fixed** — malformed-input, tool-call-warn, `_cross_check` match/mismatch, and `main()` (json output + arg validation) covered. |

Gate: **1348 unit tests pass** (+15 new), integration CLI tests pass, templates `--check` clean, ruff + ruff-format clean on all touched files.

nvzhihanj:
## Design follow-up (`706f7b5`)

Three refinements from continued review:

1. **ES percentile targets are no longer a hardcoded set** — they derive from each series' own report percentile grid, filtered to `p ≥ ES_MIN_PERCENTILE` (0.5, since an ES estimate is a *tail* certification and below-median grid entries aren't meaningful gates). One source of truth: whatever percentiles a series reports, ES covers — with the default grid that's p50/p75/p80/p90/p95/p97/p99/p99.9, and the block's `empirical` matches the grid value for every entry by construction. `EarlyStoppingSpec(percentiles=None)` means "derive"; explicit tuples remain for tests/offline analysis.
2. **`ES_TARGET_METRICS` removed** — a name-based frozenset in a distant module was fragile (a new latency metric would silently miss ES). Series now opt in with `register_series(..., tail_latency=True)` at the metric's definition site in the aggregator.
3. **Math helpers fully documented** — `_betacf`/`_betai` cite Numerical Recipes 3rd ed. §6.4 (modified Lentz continued fraction) with every magic number explained (`eps=3e-16` ≈ 1.4× machine epsilon, `fpmin=1e-300` Lentz near-zero guard, 400-iteration safety cap, the symmetry-split threshold); `_odds`/`find_min_passing`/`_discard_count` cite the exact LoadGen counterparts (`early_stopping.cc:47-58`, `:62-114`; `results.cc:162-226`).

Gate: 1348 unit tests pass; ruff/format clean on touched files.

nvzhihanj:
Pre-commit gate is now fully green on the branch (`b0315e1`): all 13 hooks pass — mypy (src+tests, 287 files), ruff/ruff-format (pinned hook versions), prettier, license headers, `regenerate-templates --check`, uv lock check. Docs updated across the repo: `--early-stopping` in docs/CLI_QUICK_REFERENCE.md, README metrics feature row, and AGENTS.md (code tree + aggregator early-stopping bullet incl. the `tail_latency` registration flag and grid-derived percentiles).

scripts/es_from_events.py
```diff
@@ -0,0 +1,267 @@
...
+    ap.add_argument("--json", dest="json_out", help="write blocks to this JSON file")
+    args = ap.parse_args(argv)
+
+    percentiles = [float(x) for x in args.percentiles.split(",")]
+    counter = _make_counter(args.tokenizer) if args.tokenizer else None
+    if counter is None:
+        print("NOTE: no --tokenizer given -> TPOT skipped", file=sys.stderr)
+
+    series = compute_series(args.events, count_tokens=counter, progress=True)
+    summary = json.load(open(args.compare)) if args.compare else None
```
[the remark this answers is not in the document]
nvzhihanj: Fixed in 166f39a — `--compare` is read via a `with open(...)` context manager now.

src/inference_endpoint/config/schema.py
nvzhihanj: Fixed in 166f39a — `@cyclopts.Parameter(name="*")` restored on `Settings`, `EarlyStoppingConfig` un-flattened. `tests/unit/commands/test_cli_flag_surface.py` pins the flag namespace and `tests/integration/commands/test_cli.py` passes again.

src/inference_endpoint/metrics/early_stopping.py
nvzhihanj: Fixed in 166f39a — `_validate_domain` rejects p/c outside (0,1) and d ≥ p with `ValueError`; the script validates `--percentiles`/`--confidence` up front. The hang was reproduced by the new test before the fix.

src/inference_endpoint/metrics/early_stopping.py
nvzhihanj: Fixed in 166f39a — `empirical` now uses `floor(p·(n−1))` (grid parity, `method="lower"`); `test_block_empirical_equals_grid_value` pins block == grid.

src/inference_endpoint/metrics/report.py
nvzhihanj: Fixed in 166f39a — `_display_metric` renders an "Early-stopping estimates" section per metric; insufficient percentiles say so explicitly.

docs/early_stopping.md
nvzhihanj: Fixed in 166f39a — reworded: estimates are computed for any scenario, just not gating metrics for throughput-bound runs.

docs/early_stopping.md
nvzhihanj: Fixed in 166f39a — companion-repo reference removed; the gap-analysis conclusion is inlined.

scripts/es_from_events.py
nvzhihanj: Fixed in 166f39a — malformed lines are counted + WARNed, non-dict JSON and missing sample_uuid/timestamp_ns skip-and-count. One sub-claim did not survive verification: msgspec.DecodeError is NOT a ValueError subclass, so the two-exception tuple stays (with a comment) — the new junk-line test proved it.

scripts/es_from_events.py
nvzhihanj: Fixed in 166f39a — duplicate ISSUED keeps the row (recv_first preserved) and refreshes only the issue timestamp, matching `MetricsTable._create_row`; covered by `test_duplicate_issued_mirrors_aggregator`.

scripts/es_from_events.py
nvzhihanj: Fixed in 166f39a — top-level imports (token_metrics is stdlib-only at module level), context-managed `--compare` read, single sort shared by `es_blocks` + `_cross_check` (finished in e71ddd0 alongside the `SERIES_TO_SUMMARY_FIELD` de-hardcoding).

scripts/es_from_events.py
nvzhihanj: Fixed in 166f39a — dead branch removed. e71ddd0 later added a type guard (`isinstance(inband, list)`) purely as corrupt-input robustness, with a current-state comment.

src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
nvzhihanj: Fixed in 166f39a — the empty arm emits self-describing n=0/sufficient:false blocks on COMPLETE when ES is enabled; `test_empty_target_series_still_reports_blocks` pins it.

src/inference_endpoint/async_utils/services/metrics_aggregator/snapshot.py
nvzhihanj: Fixed in 166f39a — estimate/empirical pass through `_scrub_nonfinite` like every other numeric field.

src/inference_endpoint/config/templates/concurrency_template_full.yaml
nvzhihanj: Fixed at the root in 166f39a — `regenerate_templates.py` comment keys are dotted paths with an indentation-aware injector; the warmup comment is restored, ES gets its own, and previously-collided comments (name/samples/eval_method/drain) reappear.

tests/unit/async_utils/services/metrics_aggregator/test_early_stopping_integration.py
nvzhihanj: Fixed in 166f39a — INTERRUPTED pinned ES-free, empty-series blocks, all three target series (incl. float tpot_ns) asserted, and empirical-vs-grid pinned; 706f7b5 extended these to the grid-derived percentile set and the tail_latency registration flag.

tests/unit/scripts/test_early_stopping_estimate_from_events.py
nvzhihanj: Fixed in 166f39a — malformed-input, tool-call-warn, `_cross_check` match/mismatch, and `main()` (json output + arg validation) covered; e71ddd0 added the product-mapping identity and non-list-inband guard tests.

nvzhihanj:
Report/UX follow-up (`e3e8d31`): `result_summary.json` now carries a compact `early_stopping_percentile` map per metric — keys mirror the `percentiles` grid (≥ p50), values are the estimate or `null` when insufficient; the rich per-percentile detail (empirical/n/min_queries/discarded) moved to aggregator INFO logging and the post-hoc script. `--no-early-stopping` removed (`negative=()`; dead surface for a default-off feature, pinned by a CLI-surface test). The post-hoc script is renamed to `scripts/early_stopping_estimate_from_events.py` and now decodes lines with the product's typed `EventRecord` (JSONL-writer mirror decoder) — no hand-parsed wire format remains. Gate: 1351 unit tests, 13/13 pre-commit hooks.

nvzhihanj:
Default flipped to ON (`e4fbf3b`) — flagged in the description as an explicit review question. Rationale: the feature is cold-path-only (computed once at COMPLETE from data the aggregator already keeps; the exact path now sorts that array once **in place** and shares it between the percentile grid, histogram, and ES estimates — so default-on adds no transient copy and roughly no finalize cost), the output field is additive, and the users who most need honest tail estimates are exactly those who wouldn't know to enable a flag. `enabled: false` / `--no-early-stopping` remains as the single opt-out for strict summary-schema consumers. A repeated-COMPLETE-snapshot test guards the in-place sort. Gate: 1352 unit tests, 13/13 pre-commit hooks.

nvzhihanj:
## Review Council — round 2 (Codex + Claude, thorough)

11 findings (2 duplicates across reviewers), all verified against HEAD and fixed in `a0ad253` + `e5f6d48`:

| Severity | Finding | Fix |
|---|---|---|
| medium | grid entry `100.0` → ES target p1.0 → `ValueError` inside `publish_final` **before** its try/except — entire final snapshot lost (default-on!) | `es_targets_from_grid` excludes ≥100; crash-repro integration test |
| medium | empty enabled series: the all-null map was built in the snapshot then dropped by `_series_to_metric_dict` (`count==0 → {}`) — summary contradicted docs [both reviewers] | empty-rollup dict keeps the map; display hardened |
| medium | `execute.py` `--early-stopping` plumbing had zero test coverage — a refactor could silently disable the default-on feature in every real run | e2e assert: `early_stopping_percentiles` in a real run's `result_summary.json` |
| medium/P3 | `discarded` reported the order-statistic offset `t`; the estimate IS the t-th highest, so only `t−1` are discarded (floor case: 0) | reports `t−1`; docstrings + anchors updated |
| low | int grids (`(99, 90, 50)`) produced ES keys (`"99.0"`) diverging from grid keys (`"99"`) — broke the 1:1 overlay contract | keys are `str()` of original grid values |
| low | `EventType.decode_hook` raises `NotImplementedError` for non-string event_type; msgspec doesn't wrap it → one line crashed the post-hoc tool | counted as malformed; fixture line added |
| low | duplicate `RECV_FIRST` (retry) diverged from the aggregator (first-only vs re-fire + window refresh) | aggregator parity; doubled `flush()` removed |
| low | `_betacf` 400-iteration cap comment was empirically false at n≈1e7 (cap IS reachable; measured ≤3e-8 error vs scipy) | comment corrected; plateaued-return documented as deliberate on the terminal path |
| low | docs drift: layering diagram `enabled=False`, stale script path in a comment, over-broad "present unless opted out" | all corrected |

Gate: 1330 unit tests (lean consolidated suite), e2e + binding integration tests, 13/13 pre-commit hooks.

scripts/early_stopping_estimate_from_events.py
```diff
@@ -0,0 +1,362 @@
...
+        "--tokenizer", help="HF model dir (or tokenizer.json) — enables TPOT"
+    )
+    ap.add_argument(
+        "--percentiles",
+        default=",".join(
+            str(f) for f in es_targets_from_grid(DEFAULT_PERCENTILES).values()
+        ),
+        help="stdout-analysis override (fractions); --json always uses the summary grid",
+    )
+    ap.add_argument("--confidence", type=float, default=CONFIDENCE)
```
arekay-nv: missing help string.

scripts/early_stopping_estimate_from_events.py
```diff
@@ -0,0 +1,362 @@
...
+        with open(args.summary) as f:
+            summary = json.load(f)
+    augmented = json.loads(json.dumps(summary)) if summary is not None else {}
+
+    rows = []
+    for name, values in series.items():
+        if not values:
+            continue
+        values.sort()
+        field = SERIES_TO_SUMMARY_FIELD[name]
```
arekay-nv: minor - dict lookup can fail here - easier to log+skip entry.

scripts/early_stopping_estimate_from_events.py
```diff
@@ -0,0 +1,362 @@
...
+                    else:
+                        rows[uuid] = [ts, None]
+                elif et is SampleEventType.RECV_FIRST:
+                    row = rows.get(uuid)
+                    if row is not None:
+                        # aggregator parity: every RECV_FIRST re-fires the ttft
+                        # trigger and refreshes the TPOT window start (retried
+                        # streaming attempts re-emit their first chunk).
+                        row[1] = ts
+                        ttft.append(ts - row[0])
```
arekay-nv, line 172: else should we increment `malformed`?

src/inference_endpoint/metrics/early_stopping.py
```diff
@@ -0,0 +1,327 @@
...
+TOLERANCE: Final[float] = 0.0
+
+# ES runs for every percentile the series' report grid shows at or above the median. The
+# estimate is a tail certification (a conservative upper confidence bound), so below-median
+# grid entries (p25/p10/...) are not meaningful gates and are skipped. Deriving from the grid
+# keeps one source of truth: whatever percentiles a series reports, ES covers.
+ES_MIN_PERCENTILE: Final[float] = 0.5
+
+
+def es_targets_from_grid(grid_percentiles) -> dict[str, float]:
```
arekay-nv: type annotation for `grid_percentiles`

src/inference_endpoint/metrics/early_stopping.py
```diff
@@ -0,0 +1,327 @@
...
+    the entries as numbers (registry grids) or strings (summary-JSON keys — an
+    int grid entry ``99`` keys as ``"99"`` either way). Entries below the median
+    are skipped (an ES estimate is a tail certification) and ``>= 100`` is
+    excluded — p1.0 is outside the binomial test's domain and would otherwise
+    crash the terminal snapshot. Fractions are rounded to 6 decimals: lossless
+    for grid values while avoiding float-division artifacts (99.9/100 != 0.999).
+    """
+    targets: dict[str, float] = {}
+    for p in grid_percentiles:
+        fraction = round(float(p) / 100.0, 6)
```
arekay-nv, line 69: Should this use `grid_percentile_key` for consistency? This is using `6` versus `4` for the key below.

arekay-nv (approved):
Looks good - thanks for the quick turnaround!

scripts/early_stopping_estimate_from_events.py
nvzhihanj: Fixed in 147ce94 — help string added.

scripts/early_stopping_estimate_from_events.py
nvzhihanj: Fixed in 147ce94 — `.get()` with a WARN-and-skip: today `compute_series` only emits the three mapped series, but if it ever grows one before the mapping does, an analysis run now degrades loudly instead of crashing.

scripts/early_stopping_estimate_from_events.py
nvzhihanj: Not malformed input, so not counted: a RECV_FIRST (or COMPLETE) without a row is the expected path for samples issued OUTSIDE the tracking window (e.g. warmup) — the aggregator likewise only creates rows while tracking is on, so skipping silently here is exactly aggregator parity. Added a comment in 147ce94 saying so, since the code didn't explain itself.

src/inference_endpoint/metrics/early_stopping.py
nvzhihanj: Fixed in 147ce94 — `grid_percentiles: Iterable[float | str]` (strings are accepted so summary-JSON grid keys pass through with their exact spelling).

src/inference_endpoint/metrics/early_stopping.py
nvzhihanj: Deliberately not `grid_percentile_key` here: that helper reconstructs float-style keys from the fraction ("99.0"), which would break the 1:1 overlay for int grids whose `percentiles` keys render as "99" — exactly the key-fidelity bug fixed in review round 2 (there's a pinned test: `es_targets_from_grid((99, 90, 50))` keys as `"99"/"90"/"50"`). So this path must key by `str()` of the ORIGINAL grid entry. On 6 vs 4: they're the same precision in different spaces — 6 decimals on the 0-1 fraction ≡ 4 decimals on the 0-100 percent value. Added comments in 147ce94 making both points at the call sites.

nvzhihanj: Eh, this (using both decimal and full number) is kinda stupid.

src/inference_endpoint/metrics/report.py
```diff
@@ -49,7 +80,10 @@ def _series_to_metric_dict(stat: dict[str, Any]) -> dict[str, Any]:
     """
     count = stat.get("count", 0)
     if count == 0:
-        return {}
+        # No data -> no rollups, but an enabled-ES empty series still self-describes
+        # (all-null map) in the summary instead of looking feature-off.
+        esp = stat.get("early_stopping_percentiles")
+        return {"early_stopping_percentiles": esp} if esp is not None else {}
```
arekay-nv, line 86: [Council: Codex+Claude, verified] **high / data-integrity**: This empty-series preservation branch (added in 166f39a) is **unreachable through the actual report path**, so the 166f39a empty-series fix is defeated on the consumer side. `Report.from_snapshot._series_dict` short-circuits `count==0` and returns `{}` *before* it ever calls `_series_to_metric_dict`:
```python
# report.py:276-280
def _series_dict(key: str) -> dict[str, Any]:
    stat = series.get(key)
    if stat is None or stat.get("count", 0) == 0:
        return {}                       # <- drops the all-null ES map here
    return _series_to_metric_dict(stat)
```
So a clean COMPLETE run where a `tail_latency=True` series recorded nothing (every request errored -> no `sample_latency_ns`, or TPOT on a non-streaming run) still emits `ttft/tpot/latency: {}` in `result_summary.json` -- indistinguishable from feature-off, exactly what 166f39a and `docs/early_stopping.md` say must not happen. The registry now emits the n=0 block (good), but the report consumer throws it away. The integration test gives false confidence: it asserts the `_series_to_metric_dict` helper directly, not `Report.from_snapshot`. Verified empirically: helper -> `{'early_stopping_percentiles': {...null...}}`, but `Report.from_snapshot(snap).ttft` -> `{}`. **Fix**: route the count==0 ES-preservation through `_series_dict` (or delete the now-dead branch here if the empty map isn't wanted).

src/inference_endpoint/metrics/early_stopping.py
```diff
@@ -0,0 +1,327 @@
...
+        as further terms cannot change the result.
+      - ``fpmin = 1e-300``: substituted for near-zero intermediate denominators (Lentz's
+        guard against division blow-up); any positive value far below the smallest
+        normal double (~2.2e-308) times a typical term works.
+      - ``400``: iteration cap. For the boundary searches at very large n (a ~ b ~ n/2
+        with x within O(1/sqrt(n)) of the mean — the p50/p75 targets at tens of millions
+        of samples) Lentz needs ~O(sqrt(a)) iterations and CAN hit this cap; the running
+        product has plateaued by then (measured <= 3e-8 relative error vs
+        scipy.special.betainc at n=1e7, far below the 0.01 threshold the result is
+        compared against), so the truncated value is returned rather than raising —
```
arekay-nv, line 115: [Council: Claude, independently reproduced] **medium / documentation**: This docstring cites a load-bearing accuracy figure that is wrong by ~3 orders of magnitude. It claims `<= 3e-8 relative error vs scipy.special.betainc at n=1e7`, but at exactly that point (a=b=5e6, x=0.5, where I_0.5(a,a)=0.5 is exact ground truth) `_betai` returns 0.50004 -- a relative error of **8.0e-5**, ~2670x the stated 3e-8 (scipy confirms 0.5). It keeps growing (2.1e-2 at a=b=2.5e7, 1.9e-1 at 1e8), so the product has **not** "plateaued". The code is nonetheless correct: the truncation is safe because the real decision boundary is odds~=0.01, never the midpoint 0.5, and `find_min_passing` returns identical answers with cap=400 vs 20000 (refs 459/44/661 match LoadGen). But the fabricated measurement violates AGENTS.md "comments must describe current state accurately" and would mislead anyone reusing `_betai` where the odds~=0.5 value matters. Replace with the real bound + the correct rationale (safety comes from the boundary, not midpoint accuracy).

src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
```diff
@@ -315,6 +369,11 @@ def _exact_stat(self) -> SeriesStat:
             for i in range(len(edges) - 1)
         ]
 
+        # Early-stopping estimates (COMPLETE path only): conservative confidence-backed
+        # bound per configured percentile, all off one sorted raw array. Cold path —
+        # the sort is one-time at run end and each estimate is a few beta evaluations.
+        early_stopping_percentiles = self._es_estimates(arr)
```
arekay-nv, line 375: [Council: Grok, verified] **medium / error-handling**: The ES estimate runs unguarded on the finalize path, where an exception loses the **entire** final report. `publish_final` sets `self._finalized = True` (publisher.py:198) and *then* calls `registry.build_snapshot(...)` (publisher.py:211), which invokes this `_es_estimates(arr)`. The `try/except` at publisher.py:219 only wraps the JSON write *after* `build_snapshot`, not `build_snapshot` itself. So any ES exception -- e.g. an `EarlyStoppingSpec.percentiles` override containing >=1.0 (which `_validate_domain` now correctly *raises* on, unlike the grid path which pre-filters >=1.0), or future grid corruption -- propagates out of `publish_final` after `_finalized=True`; the idempotency guard then makes every retry early-return, so neither `final_snapshot.json` nor the pub/sub frame is produced, and the exact percentile/histogram stats are lost too even though they computed fine. **Fix**: make ES best-effort -- wrap `_es_estimates` in try/except, log, set `early_stopping_percentiles=None`, still emit the exact `SeriesStat`. Default-config runs can't trigger this today, so this is defense-in-depth on a path the codebase otherwise guards very carefully.

src/inference_endpoint/metrics/early_stopping.py
```diff
@@ -0,0 +1,327 @@
...
+            lo = mid + 1
+    return lo
+
+
+def _discard_count(n: int, p: float, d: float, c: float) -> int:
+    """Largest ``t >= 1`` with ``n >= find_min_passing(t) + t``; 0 below the floor.
+
+    This is the LoadGen SingleStream estimate construction (``results.cc:162-226``):
+    discard the ``t`` highest samples such that the run of ``n`` queries would still
+    pass the binomial test with those as its over-latency set — the (t+1)-th highest
```
arekay-nv, line 229: [Council: Grok, verified] **low / documentation**: `_discard_count`'s docstring contradicts its own return contract -- an off-by-one trap. Line 225 defines the return as `t` in `n >= find_min_passing(t) + t`; line 229 then says "the (t+1)-th highest value is ... an upper bound", i.e. discard `t` and report the (t+1)-th. But the implementation reports the **t-th** highest (`sorted[n - t]`) with `t-1` discarded above (pinned by the unit test: t=77 -> `arr[10000-77]`, discarded=76). The same symbol `t` carries two meanings in one docstring. Reword line 229 to "the t-th highest value / t-1 discarded above" to match the code.

src/inference_endpoint/metrics/early_stopping.py
```diff
@@ -0,0 +1,327 @@
...
+    ``hi`` until it passes) followed by binary search finds the boundary exactly.
+    """
+    _validate_domain(p, d, c)
+    target = 1.0 - c
+    lo, hi = 1, 2
+    while _odds(hi, t, p, d) > target:
+        hi *= 2
+    while lo < hi:
+        mid = (lo + hi) // 2
+        if _odds(mid, t, p, d) <= target:
```
arekay-nv, line 217: [Council: Claude, verified] **low / bug**: (LoadGen parity) The binary search accepts `h` on `_odds(mid) <= target` (non-strict). The cited upstream (`loadgen/early_stopping.cc:70`) is strict -- `if (prob < 1 - c) { count = step; } else { min_h = h + 1; }` -- i.e. LoadGen treats `odds == 1-c` as *not* passing and requires `h+1`. At the exact boundary the port returns an `h` one smaller than LoadGen. Practically unreachable in IEEE-754 (all refs 459/44/661/63 match LoadGen exactly), so zero observable impact -- but for a PR whose stated value is faithful parity, either match the strict `<` or note the intentional boundary difference in the docstring.

scripts/early_stopping_estimate_from_events.py
```diff
@@ -0,0 +1,362 @@
...
+        unit, div = _UNITS[field]
+        print(f"\n{field} (from {name}, n={len(values)}):")
+        for key, r in results.items():
+            line = (
+                f"  p{key}: sufficient={r.estimate is not None} "
+                f"min_queries={r.min_queries} discarded={r.discarded}  "
+                f"empirical={_fmt(r.empirical, unit, div)}  "
+                f"estimate={_fmt(r.estimate, unit, div)}"
+            )
+            if r.estimate is not None and r.empirical:
```
arekay-nv, line 319: [Council: Grok, verified] **low / bug**: Gap reporting gates on the truthiness of `r.empirical`, so a legitimate `0.0` empirical silently drops the gap line (same pattern at line 344):
```python
if r.estimate is not None and r.empirical:      # 0.0 is falsy -> skipped
    gap = r.estimate - r.empirical
    line += f"  gap=+... (+{100 * gap / r.empirical:.2f}%)"
```
Use `r.empirical is not None` for the gap, and guard the percentage separately on `r.empirical != 0` (to avoid ZeroDivision) rather than conflating "zero" with "missing".

src/inference_endpoint/metrics/early_stopping.py
```diff
@@ -0,0 +1,327 @@
...
+    while lo < hi - 1:
+        mid = (lo + hi) // 2
+        if find_min_passing(mid, p, d, c) + mid <= n:
+            lo = mid
+        else:
+            hi = mid
+    return lo
+
+
+class EarlyStoppingResult:
```
arekay-nv, line 248: [Council: Claude, verified] **low / design**: `EarlyStoppingResult` is hand-rolled (manual `__slots__` + `__init__` + `as_dict`), while its sibling `EarlyStoppingSpec` two definitions up (line 83) is `@dataclass(frozen=True, slots=True)` -- also the repo convention (AGENTS.md: "Use dataclass(slots=True) for frequently instantiated classes"). A frozen dataclass gives the same slots/immutability with auto-generated `__init__`/`__eq__`/`__repr__` and less boilerplate. Cold-path, so purely consistency/maintainability.

tests/unit/metrics/test_early_stopping.py
```diff
@@ -0,0 +1,146 @@
...
+    ]  # int keys kept
+    assert list(es_targets_from_grid(("99.0", "50.0"))) == [
+        "99.0",
+        "50.0",
+    ]  # str keys kept
+    # explicit-spec overrides still key float-style via grid_percentile_key
+    assert grid_percentile_key(0.999) == "99.9"
+
+
+def test_estimate_reference_values_and_conservatism():
```
arekay-nv, line 85: [Council: Grok+Claude, verified] **low / testing**: A few numerical branches of the port remain unexercised (prior rounds covered wiring/empirical/grid well): (1) `_betai`'s reflection branch `x >= (a+1)/(a+b+2)` -> `1 - bt*_betacf(b,a,1-x)/b` (early_stopping.py:175) is never asserted against a reference; (2) the `_betacf` 400-iteration-cap branch (line 126 loop exiting without `break`) has no test driving the a~=b~=n/2, x=0.5 regime the docstring reasons about; (3) the `t==1`/`discarded==0` no-discard boundary (n~=662-700 for p99, estimate == max) -- `test_sufficiency_floor_boundary` only checks `estimate is not None`. Cheap to pin now that the math is default-on.

arekay-nv (commented):
## Review Council — Multi-AI Code Review (round 3)

Reviewed by **Codex** (gpt-5.5) + **Grok** (cursor-grok-4.5-high) + **Claude**, depth **thorough**. All line numbers and claims verified against HEAD; findings already raised/fixed in prior rounds were dropped. Full tiered summary in the follow-up comment.

arekay-nv:
## Review Council — Multi-AI Code Review (round 3)

Reviewed by **Codex** (gpt-5.5, xhigh) + **Grok** (cursor-grok-4.5-high) + **Claude** · depth **thorough**

Raw: Grok 5, Codex 2, Claude 5. After verifying every line/claim against HEAD and dedup'ing against the 45 existing comments from prior rounds, **8 non-duplicative findings** survived. The two most actionable are new and were **not** caught in earlier rounds.

### 🔴 Must Fix

| # | File:Line | Category | Reviewers | Summary |
|---|-----------|----------|-----------|---------|
| 1 | `metrics/report.py:86` | data-integrity | Codex + Claude | **The 166f39a empty-series fix is dead code.** `_series_dict:278` short-circuits `count==0`→`{}` *before* `_series_to_metric_dict`, so an enabled-but-empty tail series still emits `{}` in `result_summary.json` (looks feature-off). Producer emits the n=0 block; the report consumer drops it. Test asserts the helper, not `Report.from_snapshot`. Verified empirically. |

### 🟡 Should Fix

| # | File:Line | Category | Reviewers | Summary |
|---|-----------|----------|-----------|---------|
| 2 | `metrics/early_stopping.py:115` | documentation | Claude | `_betacf` docstring's "≤ 3e-8 rel error vs scipy at n=1e7" is false — actual **8.0e-5** (~2670× off), reproduced independently. Code is **safe** (boundary is odds≈0.01, not midpoint 0.5; refs match LoadGen), but the fabricated figure violates the "accurate comments" rule. |
| 3 | `metrics_aggregator/registry.py:375` | error-handling | Grok | ES runs unguarded on the finalize path: `publish_final` sets `_finalized=True` (publisher.py:198) *then* calls `build_snapshot` (211); the `try/except` at 219 only wraps the file-write. An ES exception (e.g. a ≥1.0 override, which now *raises*) loses the whole final report irrecoverably. Make ES best-effort. Defense-in-depth (not triggerable by default config). |

### 🔵 Consider

| # | File:Line | Category | Reviewers | Summary |
|---|-----------|----------|-----------|---------|
| 4 | `metrics/early_stopping.py:229` | documentation | Grok | `_discard_count` docstring says "(t+1)-th highest" but impl reports the **t-th** (`sorted[n-t]`, discarded=t−1). Same `t` used two ways — off-by-one trap. |
| 5 | `metrics/early_stopping.py:217` | bug (parity) | Claude | Binary search uses non-strict `<=`; LoadGen `early_stopping.cc:70` is strict `<`. Boundary `odds==1-c` returns h one smaller. Unreachable in IEEE-754 (refs match); parity note only. |
| 6 | `scripts/early_stopping_estimate_from_events.py:319` | bug | Grok | Gap line gated on truthy `r.empirical`, dropping a legitimate `0.0` (also line 344). Use `is not None`; guard the `%` separately. |
| 7 | `metrics/early_stopping.py:248` | design | Claude | `EarlyStoppingResult` hand-rolled while sibling `EarlyStoppingSpec` (line 83) is a `@dataclass(frozen=True, slots=True)` — repo convention. Cold-path nit. |
| 8 | `tests/unit/metrics/test_early_stopping.py:85` | testing | Grok + Claude | Untested port branches: `_betai` reflection (line 175), `_betacf` 400-iter cap (line 126), `t==1`/`discarded==0` boundary. |

### Dropped after verification
- **Codex [P2] `scripts/Dockerfile.dev:50` UV_NO_SYNC** — file is **not in this PR's diff**; Codex's worktree picked up unrelated history. Out of scope.
- **Override-percentiles ≥1.0 filter (registry.py:245)** — the domain-validation topic was already raised (early_stopping.py:218) and fixed in `166f39a` (`_validate_domain`); residual folded into finding #3.

> ⚠️ **Commit hygiene**: 16 commits including 3 apparent fixups. Consider squashing before merge.

<sub>All findings posted as inline comments on their exact lines. Event: COMMENT (no approve/request-changes).</sub>

src/inference_endpoint/metrics/early_stopping.py
nvzhihanj: Can patch in a follow up, doesn't seem critical

src/inference_endpoint/metrics/report.py
nvzhihanj: Valid — the round-2 fix was indeed dead on the consumer path. Fixed in #423 (06f3b2b): `_series_dict` now delegates unconditionally to `_series_to_metric_dict`, and the regression test pins `Report.from_snapshot` itself.

src/inference_endpoint/metrics/early_stopping.py
nvzhihanj: Valid — the figure was wrong and the plateau claim didn't hold. Rewritten in #423 (06f3b2b) with the correct rationale (safety from the odds≈0.01 decision boundary; midpoint accuracy explicitly not claimed) and a cap-regime test.

src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
nvzhihanj: Valid — fixed in #423 (06f3b2b): `_es_estimates` is best-effort (catch, log, omit the map), so no ES exception can escape `publish_final` after the finalized guard; pinned by a test with an out-of-domain override.

src/inference_endpoint/metrics/early_stopping.py
nvzhihanj: Valid — docstring reworded in #423 (06f3b2b): the t-th highest value is the bound, t−1 discarded above it.

src/inference_endpoint/metrics/early_stopping.py
nvzhihanj: Valid parity nit — #423 (06f3b2b) makes the boundary strict (`odds < 1 − c`), matching loadgen early_stopping.cc:70; reference anchors unchanged.

scripts/early_stopping_estimate_from_events.py
nvzhihanj: Valid — fixed in #423 (06f3b2b): gap gates on `is not None`, percentage guarded separately against zero.

src/inference_endpoint/metrics/early_stopping.py
nvzhihanj: Valid — `EarlyStoppingResult` is a frozen slots dataclass in #423 (06f3b2b).

tests/unit/metrics/test_early_stopping.py
nvzhihanj: Valid — #423 (06f3b2b) pins both `_betai` branches against closed forms, the iteration-cap regime, and the t=1 no-discard floor boundary (`estimate == max`, `discarded == 0`).

PR #421, opened 2026-07-17
refactor(benchmark): split execute.py into execute/profiling/accuracy/pipeline

arekay-nv:
# Refactor: split `commands/benchmark/execute.py` into cohesive modules

**Branch:** `refactor/execute-split` · **Scope:** strictly behavior-preserving · **Status:** ready for review (3 review rounds complete, all findings addressed)

## Why

`src/inference_endpoint/commands/benchmark/execute.py` had grown to **1691 lines** and
concentrated almost all CLI benchmark orchestration in one file. Two functions dominated:
`_run_benchmark_async` (~403 lines) interleaved ZMQ setup, two subprocess launches, HTTP-client
construction, agentic wiring, profiling triggers, the session run, and a ~110-line `finally`
doing five unrelated teardown jobs; `_score_accuracy` (~172 lines) mixed tokenizer loading, uuid
bounding, scoring, response counts, and OSL. Duplicated tmpfs-salvage / `AccuracyConfiguration`
construction / profile start-stop blocks and best-effort `except`-sprawl compounded it.

The goal was to split the file along its natural seams **without changing any observable
behavior** — same exception types (and therefore exit codes), same on-disk artifacts, same ZMQ
connect-before-bind and teardown ordering, same QPS/report semantics, same tmpfs ownership.

## What changed

`execute.py` split into four modules under `src/inference_endpoint/commands/benchmark/`:

| Module | Lines | Responsibility |
| --- | --- | --- |
| `execute.py` | 1691 → **1076** | Thin orchestrator: `setup_benchmark` / `run_benchmark_async` / `finalize_benchmark` / `run_benchmark`, `_run_benchmark_async`, `_build_phases`, `_load_datasets`, `_resolve_accuracy_components`, `_PerfPhaseTimeout`, `BenchmarkContext`/`BenchmarkResult`/`ResponseCollector`, and new helpers (`_create_issuer`, `_build_agentic_strategy`, `_wire_on_sample_complete`, `_write_report_artifacts`, `_summarize_and_log_metrics`) |
| `profiling.py` (new) | 222 | Profiler-trigger protocol (`_derive_profile_urls`, `_post_profile`, `_render_profile_status`, `_write_profiling_section`) + new **`ProfileController`** (owns URL derivation + start/stop/payload lifecycle) |
| `accuracy.py` (new) | 405 | `AccuracyConfiguration`, `_phase_osl_stats`, `_phase_response_counts`, `_accuracy_uuid_bound`, `_score_accuracy`, new `_load_osl_backend`, new `write_accuracy_results` |
| `pipeline.py` (new) | 345 | new **`MetricsPipeline`** (ZMQ + publisher + subscriber + aggregator/event-logger subprocess lifecycle) + `_build_aggregator_args`, `_build_event_logger_args`, `_build_report_from_snapshot`, `_load_final_snapshot_from_disk` |

### Two new abstractions

- **`ProfileController`** (profiling.py) — collapses the profile URL pre-derivation, the
  `/start_profile` (PERFORMANCE-phase-only) fire, the `/stop_profile` (only for `status==200`
  starts) fire, and the `{engine, starts, stops}` payload build into one object. Disabled and
  inert when `engine is None`; raises up-front (fail-before-run) when an engine is set but
  endpoints are empty.
- **`MetricsPipeline`** (pipeline.py) — owns the ZMQ context, event publisher, snapshot
  subscriber, and the two service subprocesses via **explicit lifecycle methods** — deliberately
  *not* an async context manager (an early design that a review rejected). The run has three
  distinct teardown paths a single `__aexit__` can't express cleanly:
  - `start()` — bring up, unwinding partial resources if service launch fails (no ZMQ leak);
  - `drain_and_build_report()` — graceful drain (publisher close → wait for services → source the
    final snapshot → build the Report), run on **both** clean-finish and session-failure;
  - `abort()` — connect-failure fast path: kill services without a graceful drain.
  `close()` exits the ZMQ scope idempotently.

### Import contracts preserved

Everything `commands/audit.py` imports from `execute` (`BenchmarkResult`, `TestMode`,
`_salvage_tmpfs`, `finalize_benchmark`, `run_benchmark_async`, `setup_benchmark`) and `cli.py`'s
imports (`resolve_report_dir`, `run_benchmark`) stay in `execute` — **`audit.py`, `cli.py`, the
package `__init__`, and the TEST04 audit tests are unchanged.** Runtime import direction is
one-way (`execute → {profiling, accuracy, pipeline}`); the sibling modules reference
`BenchmarkContext` only under `TYPE_CHECKING`, so there is no cycle.

Three unit test files were repointed to the new module paths (imports / `mock.patch` targets
only — no assertion changes): `test_benchmark.py`, `test_score_accuracy.py`,
`test_benchmark_final_snapshot.py`.

## Review process (3 rounds, Codex + Claude in parallel each round)

Every round ran two independent reviewers against the diff vs `main`; each finding was fixed and
re-verified before the next round.

| # | Finding | Sev | Reviewer | Fix |
| --- | --- | --- | --- | --- |
| F1 | Publisher double-closed on drain path (safe — `close()` is idempotent — but violated the null-after-drain invariant) | Low | Claude R1 | Null `publisher`/`subscriber` after closing in `drain_and_build_report` |
| F2 | `ProfileController` built before the run `try`, so a misconfig `ValueError` bypassed tmpfs/pbar cleanup | Low | Claude R1 | Construct it as the first statement inside the `try` (also closes a latent service-subprocess leak that existed in the original) |
| F3 | tmpfs-salvage vs pbar/zmq ordering on the failure path | Info | Claude R1 | Confirmed benign — no change |
| F4 | tmpfs/event-log/metrics `mkdir`s ran before the cleanup `try`; a mkdir failure after `tmpfs_dir` existed would leak it | Med | Codex R1 | Compute paths before the `try`, move the `mkdir`s inside it |
| F5 | Flat `finally: pbar.close(); pipe.close()` — if `pbar.close()` raised, ZMQ leaked and tmpfs wasn't salvaged (the original guaranteed both via the ZMQ `with`) | Med | Codex R2 | Nested `try: <body> finally: (try: pbar.close() finally: pipe.close())` wrapped by the outer `except BaseException` salvage |
| — | Stale `execute_mod` comment/var name in `test_score_accuracy.py` | Nit | Claude R3 | Renamed to `scoring_mod`, comment corrected |

**Round 3 verdict (both reviewers): clean / ready to commit.** Both traced the ZMQ scope exiting
exactly once on all five teardown paths (clean success, session `ExecutionError`, connect
`SetupError`→`abort()`, `launcher.launch` raising, mid-run `KeyboardInterrupt`), confirmed
`pipe.close()` survives a `pbar.close()` failure, that cleanup exceptions route to tmpfs salvage,
and that no `result`/`profiler`/`publisher`/`http_client` is possibly-unbound at any reachable use.

## Verification

> This was developed on **macOS**, where a chunk of the suite fails for
> platform reasons unrelated to this change (`pin_loadgen()` raises
> `UnsupportedPlatformError` on darwin; the metrics-aggregator/`sched_*` paths are
> Linux-only). To separate refactor regressions from environment noise, every suite
> was run **both with the change and against a clean `main` tree** (`git stash`) — the
> pass/fail profile is **identical**, so the refactor adds **zero** failures.

- **Affected unit tests** (`test_benchmark`, `test_score_accuracy`,
  `test_benchmark_final_snapshot`, compliance `test_output_caching`): **233 passed** —
  these all run on macOS and cover the moved code directly.
- **Full unit suite**: `main` 58 failed / 1542 passed → this branch **58 failed / 1542
  passed** (identical). All 58 failures are in `metrics_aggregator/` (44) and
  `endpoint_client/test_cpu_affinity.py` (14) — untouched modules that fail on macOS
  only; none is in `commands/benchmark/`.
- **Integration `commands/`** (drives `run_benchmark` end-to-end): `main` 29 failed / 8
  passed → this branch **29 failed / 8 passed** (identical). The failures are the
  `pin_loadgen()`-requires-Linux path, which the original `setup_benchmark` invoked the
  same way.
- **End-to-end smoke** (echo server, `--no-cpu-affinity`): the full refactored path runs
  clean and writes every artifact — `config.yaml`, `events.jsonl`, `sample_idx_map.json`,
  `report.txt`, `metrics/final_snapshot.json`, `performance/result_summary.json`
  (`qps≈10630`, `n_samples_issued=7000`, `7000/7000 successful`, `state=complete`,
  `complete=True`). Confirms `MetricsPipeline` bring-up + drain (`final_snapshot.json`
  sourcing → Report), tmpfs event-log salvage, and clean teardown.
- **mypy**: clean on all four files. (The pre-commit `mypy` hook reports 3 errors in
  `cpu_affinity.py` / `token_metrics.py` — the same macOS-only `sched_setaffinity` /
  `sched_getaffinity` false positives in *untouched* files; they pass in CI/Linux, and are
  why the hook needs `--no-verify` locally on macOS.)
- **ruff / ruff-format / license headers / whitespace**: all pass.

> **Note on CI:** these platform failures are macOS-only. The full suite should be run on
> Linux (CI) to confirm a green board; locally on macOS the baseline-diff above is the
> meaningful signal.

### Reproduce

```bash
# The moved code, all macOS-runnable:
uv run pytest tests/unit/commands/test_benchmark.py tests/unit/commands/test_score_accuracy.py \
  tests/unit/commands/test_benchmark_final_snapshot.py tests/unit/compliance/test_output_caching.py  # 233 pass
# Full suite on Linux/CI for a green board; on macOS compare against a stashed `main`
# tree instead (identical failure profile = no regression):
uv run pytest                                              # Linux/CI
uv run pre-commit run --files src/inference_endpoint/commands/benchmark/{execute,profiling,accuracy,pipeline}.py
# echo-server smoke — report dir gets config.yaml, events.jsonl, sample_idx_map.json,
# metrics/final_snapshot.json, performance/result_summary.json, report.txt.
# On macOS pass --no-cpu-affinity (pin_loadgen is Linux-only); dummy_1k.jsonl needs parser.prompt.
uv run python -m inference_endpoint.testing.echo_server --port 8765 &
uv run inference-endpoint benchmark offline --endpoints http://localhost:8765 --model test-model \
  --dataset "tests/assets/datasets/dummy_1k.jsonl,samples=40,parser.prompt=text_input" \
  --workers 2 --no-cpu-affinity --max-connections 16
```

## What to focus on in review

1. **`MetricsPipeline` teardown** (pipeline.py) and its three call sites in
   `execute._run_benchmark_async` — the ZMQ-scope-exactly-once invariant across all five paths.
2. **The nested cleanup structure** in `_run_benchmark_async` (F5) — that `pipe.close()` always
   runs and cleanup failures still salvage tmpfs.
3. **`accuracy._score_accuracy` / `write_accuracy_results`** equivalence to the original inline
   scoring + `accuracy_results.json` block (OSL fast-backend gating, lazy `uuid_to_text`, numpy
   coercion, finalize write order: report artifacts *before* accuracy results).

---


nv-alicheng (commented):

Reviewed by: **Claude** + **Code-Quality** | Depth: **thorough**

See the summary comment for the tiered breakdown. `codex` was unavailable in this environment, so this is a Claude + Code-Quality review.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -19,66 +19,56 @@
...
-    ServiceLauncher,
-)
-from inference_endpoint.async_utils.services.metrics_aggregator.snapshot import (
-    snapshot_to_dict,
-)
-from inference_endpoint.async_utils.services.metrics_aggregator.subscriber import (
-    MetricsSnapshotSubscriber,
+from inference_endpoint.commands.benchmark.accuracy import (
+    AccuracyConfiguration,
+    _effective_external_sample_count,
```
nv-alicheng: [Code-Quality] **medium** (code-quality): The split promoted `_effective_external_sample_count` and `_score_accuracy` into cross-module API, but they keep their `_`-prefixed private names — so `execute.py` now imports across a package boundary into `accuracy.py`'s privates. The `_` prefix now lies about the visibility contract (contrast the sibling `write_accuracy_results` on the next line, correctly public). **Root cause (shared with the `_write_profiling_section` import below):** a symbol that becomes cross-module API in a split should be renamed public. Promote these to public names.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -19,66 +19,56 @@
...
+    _score_accuracy,
+    write_accuracy_results,
 )
-from inference_endpoint.async_utils.services.metrics_aggregator.token_metrics import (
-    encode_lengths,
-    load_reference_backend,
+from inference_endpoint.commands.benchmark.pipeline import MetricsPipeline
+from inference_endpoint.commands.benchmark.profiling import (
+    ProfileController,
+    _write_profiling_section,
```
nv-alicheng: [Code-Quality] **medium** (code-quality): `_write_profiling_section` is imported from `profiling.py` and called at `execute.py:1027` to render the report.txt section — it is de facto public API of the profiling module, yet still `_`-prefixed. Drop the leading underscore (same root cause as the accuracy-privates import above).

src/inference_endpoint/commands/benchmark/profiling.py
```diff
@@ -0,0 +1,222 @@
...
+            rec = _post_profile(url)
+            if rec["status"] == 200:
+                logger.info("Profile start: %s -> 200 OK", url)
+            else:
+                logger.warning(
+                    "Profile start: %s -> %s", url, rec["error"] or rec["status"]
+                )
+            self._starts.append(rec)
+
+    def stop(self, completed_normally: bool) -> None:
```
nv-alicheng, line 185: [Claude] **medium** (testing): `ProfileController.start`/`stop` have no coverage. Every `_run_benchmark_async` test in `test_benchmark.py` raises before `session.run` completes, so `profiler.start()` (execute.py:897) and `profiler.stop()` (execute.py:912) never execute. The refactor moved real branching into `stop()` — the non-200 / `i >= len(self._stop_urls)` skip (line 195) and the `stop_reason` abort-vs-phase_end selection (line 193) — none of it exercised. Add a direct `ProfileController` unit test covering the start→stop index mapping.

src/inference_endpoint/commands/benchmark/profiling.py
```diff
@@ -0,0 +1,222 @@
...
+        self._starts: list[dict[str, Any]] = []
+        self._stops: list[dict[str, Any]] = []
+        if engine is not None:
+            # Preserve the truthiness fallback: an explicit empty urls override
+            # falls through to endpoint_config.endpoints.
+            profile_endpoints = urls_override or endpoints
+            self._start_urls = _derive_profile_urls(profile_endpoints, engine, "start")
+            self._stop_urls = _derive_profile_urls(profile_endpoints, engine, "stop")
+
+    @property
```
nv-alicheng: [Claude] **low** (design): `ProfileController.enabled` (`@property -> self._engine is not None`) is dead code introduced by the split — a repo-wide grep finds no reference. The pre-refactor inline code had no such accessor. Either wire it into the orchestrator (gate `start()`/`stop()` on it) or drop it; an unused public accessor invites callers onto an untested path.

src/inference_endpoint/commands/benchmark/pipeline.py
```diff
@@ -0,0 +1,356 @@
...
+        run_config["dataloader_random_seed"] = runtime_cfg.dataloader_random_seed
+        report = Report.from_snapshot(
+            snap_dict,
+            run_config=run_config,
+            use_legacy_loadgen_qps_metrics=(
+                load_pattern.type == LoadPatternType.POISSON
+                and load_pattern.use_legacy_loadgen_qps_metrics
+            ),
+        )
+        if not report.complete:
```
nv-alicheng, line 175: [Claude] **low** (testing): `_build_report_from_snapshot` moved two warning branches out of the monolith that are now untested: the incomplete-report warning (`if not report.complete:`, 170-175) and the legacy-loadgen-QPS deprecation warning (176-183), plus the swallow-to-None `except` at 185. `test_benchmark_final_snapshot.py` only tests `_load_final_snapshot_from_disk`; `_build_report_from_snapshot` is never called from tests.

src/inference_endpoint/commands/benchmark/pipeline.py
```diff
@@ -0,0 +1,356 @@
...
+        self.publisher.close()
+        logger.info("Waiting for services to finish processing...")
+        await asyncio.to_thread(self._launcher.wait_for_exit, None)
+
+        snap_dict = _load_final_snapshot_from_disk(
+            self._metrics_output_dir / "final_snapshot.json"
+        )
+        if snap_dict is not None:
+            logger.info("Built report from final_snapshot.json")
+        elif self.subscriber is not None and self.subscriber.latest is not None:
```
nv-alicheng, line 347: [Claude] **low** (testing): `drain_and_build_report`'s subscriber-fallback branch (298-307: on-disk `final_snapshot.json`, else last pub/sub snapshot via `snapshot_to_dict(self.subscriber.latest)`, else error log) is untested — no direct `MetricsPipeline` unit test exists, and all `_run_benchmark_async` tests fail before reaching here. The disk-missing fallback is exactly the SIGKILL/OOM recovery path that matters most.

src/inference_endpoint/commands/benchmark/accuracy.py
```diff
@@ -0,0 +1,434 @@
...
+                    osl = _phase_osl_stats(
+                        scorer_instance.sample_index_map, uuid_to_text, osl_backend
+                    )
+                    if osl is not None:
+                        # Same shape/key as the perf report output_sequence_lengths.
+                        entry["output_sequence_lengths"] = osl
+                        # Wall-clock of just this phase's tokenization (seconds);
+                        # summed across datasets for the accuracy report's total.
+                        entry["osl_tokenize_s"] = round(time.perf_counter() - t0, 3)
+            except Exception as e:  # noqa: BLE001 - optional blocks; never fail scoring
```
nv-alicheng, line 370: [Claude] **low** (testing): In the moved `_score_accuracy`, the `get_raw_outputs` build-once branch (342-355) and the broad `except Exception` that drops response-counts/OSL without failing scoring (370-375) are load-bearing. `test_score_accuracy.py` drives `_score_accuracy` with fake backends but no test forces `get_raw_outputs`/tokenize to raise and asserts the entry is still appended with `score` present and the counts/OSL keys simply absent — the docstring's 'a read failure only drops these blocks, never fails scoring' invariant is unverified.

src/inference_endpoint/commands/benchmark/profiling.py
```diff
@@ -0,0 +1,222 @@
...
+
+# (start_path, stop_path) for each supported inference engine's profiling
+# protocol. Add a row when introducing a new ProfilerEngine variant.
+_PROFILE_PATHS: dict[ProfilerEngine, tuple[str, str]] = {
+    ProfilerEngine.VLLM: ("/start_profile", "/stop_profile"),
+}
+
+
+def _derive_profile_urls(
+    endpoints: list[str], engine: ProfilerEngine, action: str
```
nv-alicheng: [Code-Quality] **low** (code-quality): `_derive_profile_urls(..., action: str)` is stringly-typed; only `"start"`/`"stop"` are valid and it branches `path = start_path if action == "start" else stop_path`. Use `Literal["start", "stop"]` (or an enum) so an out-of-range string can't silently select the stop path.

src/inference_endpoint/commands/benchmark/profiling.py
```diff
@@ -0,0 +1,222 @@
...
+    record: dict[str, Any] = {
+        "url": url,
+        "sent_at_ns": time.monotonic_ns(),
+        "sent_at_iso": datetime.now().isoformat(timespec="milliseconds"),
+        "status": None,
+        "error": None,
+    }
+    req = urllib_request.Request(url, method="POST", data=b"")
+    try:
+        with urllib_request.urlopen(req, timeout=2) as resp:
```
nv-alicheng: [Code-Quality] **low** (code-quality): `urlopen(req, timeout=2)` inlines a magic constant for the profile-trigger HTTP timeout — a meaningful tunable buried in a helper. Name it (e.g. `_PROFILE_POST_TIMEOUT_S = 2`).

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -760,110 +702,92 @@ def cancel(self) -> None:
...
+        dataset_metadata=ctx.dataloader.conversation_metadata,
+        agentic_inference_config=agentic_cfg,
+        target_concurrency=ctx.config.settings.load_pattern.target_concurrency,
     )
 
 
+def _wire_on_sample_complete(
+    collector: ResponseCollector,
+    agentic_inference_strategy: AgenticInferenceStrategy | None,
+    publisher: Any,
```
nv-alicheng: [Code-Quality] **low** (code-quality): `_wire_on_sample_complete(..., publisher: Any)` — the argument is always `pipe.publisher`, an `EventPublisherService` (asserted non-None at the call site). Annotate precisely (import under `TYPE_CHECKING`) instead of `Any`.

src/inference_endpoint/commands/benchmark/accuracy.py
```diff
@@ -0,0 +1,434 @@
...
+from inference_endpoint.utils.atomic_write import atomic_write_bytes
+
+if TYPE_CHECKING:
+    from inference_endpoint.commands.benchmark.execute import BenchmarkContext
+    from inference_endpoint.config.schema import EndpointConfig, ModelParams
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
```
nv-alicheng: [Code-Quality] **low** (code-quality): `AccuracyConfiguration` is a read-only config carrier — built once per dataset and never mutated. Declare it `@dataclass(frozen=True)` to encode the build-once/read-many invariant in the data model.

nv-alicheng:

Reviewed by: **Claude** + **Code-Quality** | Depth: **thorough**
(`codex` CLI was unavailable in this environment — Claude + Code-Quality only.)

Found **12** issues across 4 files. The refactor is **largely faithful** — imports, call sites, operation ordering, and error paths are preserved across the split. No critical/high defects; findings are testing gaps, maintainability debt on the new module boundaries, and one dead accessor. 11 posted inline; 1 (execute.py:1082) is summary-only (outside the diff hunk).

_None._

### 🟡 Should Fix (medium)

| 1 | `benchmark/execute.py` | 55 | code-quality | Code-Quality | `_effective_external_sample_count`/`_score_accuracy` are now cross-module API but kept `_`-private — rename public (shared root cause with #2) |
| 2 | `benchmark/execute.py` | 62 | code-quality | Code-Quality | `_write_profiling_section` is de facto public (called at execute.py:1027) — drop the underscore |
| 3 | `benchmark/profiling.py` | 186 | testing | Claude | `ProfileController.start`/`stop` never execute in the suite (all `_run_benchmark_async` tests raise first); stop() branching untested |

### 🔵 Consider (low)

| 4 | `benchmark/profiling.py` | 170 | design | Claude | `ProfileController.enabled` is dead code (no references) — wire in or drop |
| 5 | `benchmark/pipeline.py` | 170 | testing | Claude | `_build_report_from_snapshot` warning branches + swallow-to-None except untested |
| 6 | `benchmark/pipeline.py` | 300 | testing | Claude | `drain_and_build_report` subscriber-fallback (SIGKILL/OOM recovery path) untested |
| 7 | `benchmark/accuracy.py` | 370 | testing | Claude | `_score_accuracy` broad-except "drop counts, never fail scoring" invariant unverified |
| 8 | `benchmark/profiling.py` | 47 | code-quality | Code-Quality | `_derive_profile_urls(action: str)` stringly-typed — use `Literal["start","stop"]` |
| 9 | `benchmark/profiling.py` | 82 | code-quality | Code-Quality | `urlopen(timeout=2)` magic constant — name it |
| 10 | `benchmark/execute.py` | 767 | code-quality | Code-Quality | `_wire_on_sample_complete(publisher: Any)` — annotate as `EventPublisherService` |
| 11 | `benchmark/accuracy.py` | 58 | code-quality | Code-Quality | `AccuracyConfiguration` never mutated — `@dataclass(frozen=True)` |
| 12 | `benchmark/execute.py` | 1082 | error-handling | Claude | `profiling.json` write is non-atomic/unguarded and first in `finalize_benchmark` — an `OSError` sinks the already-built perf report. Moved verbatim (not a regression). _(outside diff hunk — not postable inline)_ |

**Cross-cutting note:** #1/#2 share a root cause — the split left helpers that are now cross-module API with private (`_`) names. Renaming them public is the durable fix; the rename spans importer + definition + call sites, so no inline suggestion was attached.

> Existing `gemini-code-assist` bot comments (publisher-leak on abnormal `close()`, `_salvage_tmpfs` masking the original exception, `subscriber.close()` unguarded in drain) were **not duplicated** — they cover different lines/concerns and remain valid.

src/inference_endpoint/commands/benchmark/pipeline.py
```diff
@@ -0,0 +1,356 @@
...
+        Connects the snapshot subscriber BEFORE launching the aggregator that binds
+        the matching PUB socket (ZMQ tolerates connect-before-bind on IPC; starting
+        the SUB reader early lets the subscription handshake complete during the
+        ~1-2s subprocess-launch window, avoiding the slow-joiner risk of dropping
+        early live ticks — or, worst case, missing COMPLETE). On any failure the
+        partially-acquired resources are unwound so nothing (notably the ZMQ
+        context) leaks.
+        """
+        self._zmq_cm = ManagedZMQContext.scoped(io_threads=2)
+        zmq_ctx = self._zmq_cm.__enter__()
```
nv-alicheng: code smell - a context manager's `__enter__` and `__exit__` shouldn't be called explicitly unless by another context manager's enter and exit methods.

Something tells me there's a better way to structure the call-chain in MetricsPipeline and the consumer of MetricsPipeline to avoid this and use context managers correctly to handle cleanup.

hvagadia: Concrete failure mode here: `MetricsPipeline._teardown()` sets `_closed = True` before any teardown step succeeds, while its guards catch only `Exception`. A `KeyboardInterrupt`/`SystemExit` from `publisher.close()`, `kill_all()`, `subscriber.close()`, or the explicit ZMQ `__exit__()` stops the remaining teardown; a later `abort()`/`close()` then immediately returns because `_closed` is already true. That can leave service children or the ZMQ context alive. Structuring this as context-manager/finally-owned cleanup should also ensure `_closed` is committed only after the remaining teardown has run, or otherwise guarantee every remaining step runs despite a `BaseException`.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1141,134 +907,50 @@ def _on_phase_start(phase: PhaseConfig) -> None:
...
+                # a failed run still has a terminal snapshot worth draining).
+                report = await pipe.drain_and_build_report()
+                drained = True
+        finally:
+            # Cleanup runs on every path. pipe.close() (ZMQ scope exit) must run
+            # even if pbar.close() raises, and a failure here propagates to the
+            # outer except → tmpfs salvage — matching the monolith's guarantees
+            # (pbar.close was inside the ZMQ `with`, whose __exit__ always ran).
+            try:
                 pbar.close()
```
hvagadia: [Codex] **medium** (error-handling): This unconditional `pbar.close()` can replace the exception already being unwound. For failures before `session.run` (for example service launch, profiler validation, or other setup errors), if a tqdm write/flush raises something like `BrokenPipeError`, the CLI reports that cleanup error instead of the original failure, potentially changing the diagnostic and exit classification. Please make progress-bar close best-effort, or otherwise preserve the active exception while still guaranteeing `pipe.abort()`/`pipe.close()` run.

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1065,134 +843,40 @@ def _on_phase_start(phase: PhaseConfig) -> None:
...
+            finally:
+                pipe.close()
     except BaseException:
-        # tmpfs_dir may still be None if the exception hit before it was
-        # created (e.g. ZMQ context setup), in which case there is nothing
-        # to clean up.
-        if tmpfs_dir is not None and tmpfs_dir.exists():
+        if tmpfs_dir.exists():
             _salvage_tmpfs(ctx.report_dir, tmpfs_dir)
             shutil.rmtree(tmpfs_dir, ignore_errors=True)
```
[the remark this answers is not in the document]
hvagadia: [Codex] **medium** (data-integrity): This prevents the salvage error from masking the original exception, but the unconditional `rmtree` immediately after the `except` now deletes tmpfs even when `_salvage_tmpfs()` failed. If `copy2` failed because the report filesystem is full or unavailable, tmpfs may hold the only `events.jsonl`, so the run becomes unrecoverable. Please remove the tmpfs directory only after salvage succeeds; on failure, preserve it and log its path. The same pattern appears in `run_benchmark()` around lines 1181-1185.

src/inference_endpoint/commands/benchmark/pipeline.py
```diff
@@ -0,0 +1,384 @@
...
+        # released by the ExitStack in __aexit__.
+        self.publisher = None
+        return report
+
+    def _kill_services(self) -> None:
+        """Best-effort service termination owned by the pipeline ExitStack."""
+        if self._launcher is None:
+            return
+        try:
+            self._launcher.kill_all()
```
hvagadia: [Codex] **medium** (data-integrity): This teardown does not preserve the interrupted-snapshot semantics the metrics service expects. `ServiceLauncher.kill_all()` calls `Popen.kill()`, which sends SIGKILL on POSIX, while the metrics aggregator installs a SIGTERM handler to persist an `INTERRUPTED` `final_snapshot.json`. Cancellation or a drain failure can therefore discard the current partial metrics snapshot. Also, `kill_all()` iterates the children without per-process `BaseException` handling, so a second Ctrl-C can stop the loop before every child is handled. Please send SIGTERM to every child, use a bounded wait with SIGKILL only as fallback, and re-raise an interrupt only after every child has been attempted.

hvagadia (approved):

hvagadia:
LGTM after open comments are resolved.

PR #423, opened 2026-07-17
refactor(metrics): unify the percentile convention; address #417 round-3 review

nvzhihanj:
Follow-up to #417 (early-stopping percentile estimates). Rebased onto `main` after #441 / #442 landed; subsequently updated from `main` again for #446, so the branch carries one merge commit. Squash-merge recommended — the docs-only and review-follow-up commits only read as a unit.

## 1. `bcb4ebe` — address the post-merge round-3 review of #417

All eight findings from the round-3 review council on #417 (each verified against HEAD before fixing):

- **high / data-integrity**: the #417 empty-series preservation was dead code on the real consumer path — `Report.from_snapshot._series_dict` short-circuited `count==0` to `{}` before `_series_to_metric_dict` could keep the all-null `early_stopping_percentiles` map. Now delegates unconditionally, and the regression test pins `Report.from_snapshot` itself rather than the helper (the previous test's false-confidence trap).
- **medium / error-handling**: ES is now best-effort on the finalize path — `_es_estimates` catches, logs, and omits the map instead of letting an exception escape `publish_final` after the finalized guard is set (which would cost the entire final report).
- **medium / documentation**: the `_betacf` iteration-cap comment cited a fabricated `3e-8` accuracy figure; replaced with the correct rationale (safety comes from the odds ≈ 0.01 decision boundary — midpoint accuracy in the cap regime is explicitly *not* claimed, and is now pinned by a test).
- **low / parity**: `find_min_passing`'s boundary is now strict (`odds < 1 − c`), byte-matching `loadgen/early_stopping.cc:70`.
- **low**: `_discard_count` docstring off-by-one wording; script gap lines gated on `is not None` (a legitimate `0.0` empirical was silently dropped) with zero-division guarded separately; `EarlyStoppingResult` converted to a frozen slots dataclass like its sibling spec.
- **testing**: `_betai` both branches pinned against closed forms (`I_x(a,1) = x^a`, `I_x(1,b) = 1−(1−x)^b`), the iteration-cap regime pinned sane, and the `t=1` no-discard floor boundary pinned (`estimate == max`, `discarded == 0`).

## 2. `49998aa` — one percentile convention: grid (0–100) at every surface

The codebase mixed `99.9` (grid, `np.percentile`, summary keys) and `0.999` (the whole ES surface), stitched together by **two different rounding hacks** (`round(p/100, 6)` and `round(p*100, 4)`) and a key-reconstruction helper. Now:

- `EarlyStoppingSpec.percentiles`, `es_percentile_estimate(percentile=…)`, `EarlyStoppingResult.percentile`, `ES_MIN_PERCENTILE` (= 50.0), `es_targets_from_grid` values, and the script's `--percentiles` all speak 0–100.
- `es_targets_from_grid` does **no conversion and no rounding at all** — keys are `str()` of the original grid entries, values the entries themselves.
- `grid_percentile_key` is deleted; both rounding constants are gone.
- The LoadGen-parity kernel (`find_min_passing`/`_odds`/`_discard_count`) keeps LoadGen's own fraction domain; the **single** conversion lives inside `es_percentile_estimate`, unrounded — the same `/100` that `np.percentile` applies internally, so `empirical`-vs-grid parity is exact by construction.
- Estimates are bit-identical before/after (interface-only change). This also closes the round-3 question about the 6-vs-4 rounding inconsistency — both roundings no longer exist.

## 3. `fd54268` — docs: minimum samples per percentile

Cheat-sheet table of the sufficiency floor (`find_min_passing(1, p) + 1` at c = 0.99): p50 → 11 samples, p90 → 64, p99 → 662, p99.9 → 6,636, p99.99 → 66,381. Rule of thumb: floor ≈ 6.64 / (1 − p/100) — one more "9" costs 10× the samples.

## 4. `7e4cf60` — reject fraction-style percentiles; dedupe CLI targets

With the convention unified, `--percentiles 0.99` would silently mean "p0.99%" rather than p99. The script now rejects any value below 1 with the intended value in the message (`did you mean [99.0]?`), and dedupes targets on the parsed value so `99,99.0` is one computation under one canonical key.

## 5. `c773bb0` — docs: how the estimate is computed

Pseudo-code for the two monotone searches plus a worked example (n = 10,000, p99 → floor 662, budget t = 77, estimate is the 77th-highest sample), and the statistical framing of the binomial tail test the beta function evaluates.

## 6. `90b1f3d` — review follow-ups

- `test_sufficiency_floor_boundary` names 662 as the p99 minimum and pins both sides of it (661 → no estimate; 662 → estimate == max, nothing discarded), so the floor cannot drift without failing the test.
- The empty-series regression test states the invariant it protects instead of narrating the review that found it; the remaining `pre-#423` convention-history references are gone from comments and the CLI error text (the user-facing "did you mean" hint is unchanged).

## 7. `6b39f02` — review follow-ups (round 2)

- `es_percentile_estimate` rejects percentiles below 1 rather than computing them. The `(0, 100)` domain admitted a fraction-style `0.99`, which certified the sub-1% percentile p0.99 — and since that target's sufficiency floor is only 3 samples, it produced a confident-looking estimate rather than a `null`. The grid path filters at 50 and the script already rejected it, but `EarlyStoppingSpec(percentiles=(0.99,))` reached the kernel unchecked. Guarding the one function all three surfaces funnel through makes them agree without duplicating the rule; a tripped override degrades through the existing best-effort catch exactly as `150.0` does.
- The explicit-override path dedupes on the parsed value, so `(99, 99.9, 99.0)` yields `["99.9", "99"]` instead of three keys computing p99 twice — the rule the script already applies to its `--percentiles` fallback. The first spelling still wins the key.
- `find_min_passing`'s summary line said `<= 1 - c` while the implementation and the rest of its docstring use the strict `<` this PR pinned for LoadGen parity.

Both new behaviours are pinned by tests that fail on the pre-fix code.

Gate: 1562 unit tests, 13/13 pre-commit hooks, LoadGen reference anchors (459/44/661) unchanged.

### DIFF
--- a/src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
+++ b/src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
@@ -45,7 +45,6 @@ from inference_endpoint.metrics.early_stopping import (
     EarlyStoppingSpec,
     es_percentile_estimate,
     es_targets_from_grid,
-    grid_percentile_key,
 )
 
 from .snapshot import (
@@ -257,10 +256,7 @@ class SeriesSampler(MetricSampler):
         self, sorted_values, spec: EarlyStoppingSpec
     ) -> dict[str, float | None]:
         if spec.percentiles is not None:  # explicit override: tests / offline analysis
-            targets = {
-                grid_percentile_key(f): f
-                for f in sorted(spec.percentiles, reverse=True)
-            }
+            targets = {str(v): float(v) for v in sorted(spec.percentiles, reverse=True)}
         else:
             targets = es_targets_from_grid(self._percentiles)
         results = {
--- a/src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
+++ b/src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
@@ -242,6 +242,20 @@ class SeriesSampler(MetricSampler):
         if self._es_spec is None:
             return None
         spec = self._es_spec
+        try:
+            return self._compute_es_estimates(sorted_values, spec)
+        except Exception:
+            # Best-effort by design: this runs inside publish_final's
+            # build_snapshot, where an exception would cost the ENTIRE final
+            # report (the publisher's finalized guard makes retries no-ops).
+            logger.exception(
+                "%s: early-stopping computation failed; omitting the map", self.name
+            )
+            return None
+
+    def _compute_es_estimates(
+        self, sorted_values, spec: EarlyStoppingSpec
+    ) -> dict[str, float | None]:
         if spec.percentiles is not None:  # explicit override: tests / offline analysis
             targets = {
                 grid_percentile_key(f): f
--- a/src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
+++ b/src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
@@ -256,7 +256,13 @@ class SeriesSampler(MetricSampler):
         self, sorted_values, spec: EarlyStoppingSpec
     ) -> dict[str, float | None]:
         if spec.percentiles is not None:  # explicit override: tests / offline analysis
-            targets = {str(v): float(v) for v in sorted(spec.percentiles, reverse=True)}
+            # Descending, deduped on the parsed value so 99 and 99.0 are one target
+            # (the rule the post-hoc script applies to its --percentiles fallback);
+            # the first spelling wins the key, so an int override still keys "99".
+            seen: dict[float, str] = {}
+            for p in sorted(spec.percentiles, reverse=True):
+                seen.setdefault(float(p), str(p))
+            targets = {key: value for value, key in seen.items()}
         else:
             targets = es_targets_from_grid(self._percentiles)
         results = {
--- a/src/inference_endpoint/metrics/early_stopping.py
+++ b/src/inference_endpoint/metrics/early_stopping.py
@@ -284,10 +284,12 @@ def es_percentile_estimate(
 ) -> EarlyStoppingResult:
     """Conservative early-stopping estimate over an ascending-sorted latency series.
 
-    ``percentile`` uses the grid convention (0-100, e.g. ``99.9``); the single
-    conversion to the kernel's fraction domain happens here, unrounded — the same
-    ``/100`` ``np.percentile`` applies internally, so the ``empirical`` value can
-    never diverge from the report grid's ``method="lower"`` entry.
+    ``percentile`` uses the grid convention (0-100, e.g. ``99.9``) — a value below
+    1 almost certainly means a caller still passing the pre-#423 fraction style
+    and will be interpreted as a sub-1% percentile. The single conversion to the
+    kernel's fraction domain happens here, unrounded — the same ``/100``
+    ``np.percentile`` applies internally, so the ``empirical`` value can never
+    diverge from the report grid's ``method="lower"`` entry.
     """
     if not 0.0 < percentile < 100.0:
         raise ValueError(f"percentile must be in (0, 100), got {percentile}")
--- a/src/inference_endpoint/metrics/early_stopping.py
+++ b/src/inference_endpoint/metrics/early_stopping.py
@@ -113,13 +113,15 @@ def _betacf(a: float, b: float, x: float) -> float:
         guard against division blow-up); any positive value far below the smallest
         normal double (~2.2e-308) times a typical term works.
       - ``400``: iteration cap. For the boundary searches at very large n (a ~ b ~ n/2
-        with x within O(1/sqrt(n)) of the mean — the p50/p75 targets at tens of millions
-        of samples) Lentz needs ~O(sqrt(a)) iterations and CAN hit this cap; the running
-        product has plateaued by then (measured <= 3e-8 relative error vs
-        scipy.special.betainc at n=1e7, far below the 0.01 threshold the result is
-        compared against), so the truncated value is returned rather than raising —
-        this runs on the terminal-snapshot path, where an exception would cost the
-        whole final report.
+        with x near the mean — the p50/p75 targets at tens of millions of samples)
+        Lentz needs ~O(sqrt(a)) iterations and CAN hit this cap, returning a truncated
+        value whose error near the distribution midpoint grows with n (order 1e-4 at
+        a = b = 5e6 and worse beyond — do NOT reuse ``_betai`` standalone where
+        midpoint accuracy matters). Truncation is safe HERE because every decision
+        compares odds against 1 - c (~0.01), far from the slow-converging midpoint:
+        ``find_min_passing`` returns identical answers with cap 400 vs 20000 on the
+        LoadGen references (459/44/661). Returned rather than raised — this runs on
+        the terminal-snapshot path, where an exception would cost the final report.
     """
     eps, fpmin = 3e-16, 1e-300
     qab, qap, qam = a + b, a + 1.0, a - 1.0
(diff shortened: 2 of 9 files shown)


nvzhihanj:
## Review round (Codex + Claude, both on `eb7fc7b`)

**Codex** (`codex review --base origin/main`): no findings — it ran the ES unit suites and probed the float/rounding behavior of the convention switch directly.

**Claude** reported 3 findings; each was verified against HEAD before fixing, all addressed in `50e02bf`:

1. **medium / api-contract — silent fraction-style footgun.** After the convention switch, a pre-#423 invocation like `--percentiles "0.99"` is *valid input* meaning the sub-1% percentile p0.99 — it would produce a plausible-looking `sufficient=true` block instead of failing. The script now hard-errors on values `< 1` with a did-you-mean message (`0.99 → did you mean 99.0?`), and `es_percentile_estimate`'s docstring calls out the hazard. Sub-1% targets have no use here (ES is a tail certification; the product path filters at `ES_MIN_PERCENTILE = 50`), so rejecting is strictly safer than guessing. Test added: fraction-style input raises `SystemExit`.
2. **low / correctness — CLI fallback targets didn't dedupe on value.** `--percentiles "99,99.0"` produced two identical computations under two keys (`"99"`, `"99.0"`). Fallback keys are now canonical `str(float(...))`, deduped by parsed value, descending to match the grid order. Test added. (The product path is unaffected — grid entries are already canonical.)
3. **low / documentation — stale comment.** A `snapshot.py` comment still referenced `grid_percentile_key`, deleted by this PR. Reworded.

Claude also verified the core claims of the refactor negatively: the strict-vs-lenient boundary behavior of `find_min_passing` is identical across the domain change, and the single unrounded `/100` matches `np.percentile`'s internal division on all grid values tested up to 1e6.

Gate on `50e02bf`: 1341 unit tests pass, pre-commit green.

tests/unit/metrics/test_early_stopping.py
```diff
@@ -95,35 +83,52 @@ def test_estimate_reference_values_and_conservatism():
 
 def test_sufficiency_floor_boundary():
     # Below the floor the estimate must be None (never a fabricated bound), the
```
arekay-nv: Add a comment here to mention that 662 is the min samples required. Below it would be insufficient.

tests/unit/async_utils/services/metrics_aggregator/test_early_stopping_integration.py
```diff
@@ -140,6 +140,28 @@ def test_custom_grid_with_p100_and_int_keys_is_safe():
     assert list(esp) == ["99", "50"]  # 100.0 excluded; int keys + grid order kept
 
 
+def test_report_from_snapshot_preserves_empty_series_map():
+    # The HIGH round-3 finding: _series_to_metric_dict preserved the map but
```
arekay-nv: Probably want to remove the reference to the review itself `The HIGH round-3 finding:`

arekay-nv (approved):
Thanks for the followup.

tests/unit/metrics/test_early_stopping.py
nvzhihanj: Done in 90b1f3d8. The comment now names 662 explicitly and says where it comes from (`min_queries = find_min_passing(1, p99) + 1`, cross-referenced to the per-percentile cheat sheet in `docs/early_stopping.md`).

I also tightened the boundary so the number is pinned rather than only described — the test previously used 600 and 663, which straddle the floor loosely:

- `n = 661` → no estimate (insufficient), and `min_queries == 662` asserted directly
- `n = 662` → estimate is the maximum observed sample, nothing discarded

So if the floor ever drifts, this fails instead of leaving a stale comment behind.

tests/unit/async_utils/services/metrics_aggregator/test_early_stopping_integration.py
nvzhihanj: Done in 90b1f3d8. The comment now states the invariant instead of the review that found it:

> An enabled-but-empty series must still carry its all-null `early_stopping_percentiles` map into the summary. Pin the REAL consumer path (`Report.from_snapshot`): a `count==0` short-circuit there drops the map even when `_series_to_metric_dict` preserves it, so exercising the helper alone would prove nothing.

Good catch — and applying the same rule, I dropped the remaining `pre-#423` convention-history references from `early_stopping.py`, the script comment, the CLI error text, and one test comment. The user-facing "did you mean 99.0?" migration hint is unchanged; only the "this changed in PR N" framing is gone.

src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
```diff
@@ -242,11 +241,22 @@ def _es_estimates(self, sorted_values) -> dict[str, float | None] | None:
...
+
+    def _compute_es_estimates(
+        self, sorted_values, spec: EarlyStoppingSpec
+    ) -> dict[str, float | None]:
         if spec.percentiles is not None:  # explicit override: tests / offline analysis
-            targets = {
-                grid_percentile_key(f): f
-                for f in sorted(spec.percentiles, reverse=True)
-            }
+            targets = {str(v): float(v) for v in sorted(spec.percentiles, reverse=True)}
```
viraatc: Should we mirror the script's fraction-style guard here on the explicit-override path? As written, a caller passing the old convention (`EarlyStoppingSpec(percentiles=(0.99,))`) stays inside `es_percentile_estimate`'s `(0, 100)` domain and quietly computes a sub-1% percentile keyed `"0.99"`, rather than raising into the `_es_estimates` best-effort catch the way an out-of-domain `150.0` does. A small `< 1.0` `ValueError` here would make the registry, script, and kernel surfaces consistent. Non-blocking.

src/inference_endpoint/metrics/early_stopping.py
```diff
@@ -209,17 +204,18 @@ def find_min_passing(
     LoadGen's ``MinPassingQueriesFinder`` (``loadgen/early_stopping.cc:62-114``): the
     smallest number of under-latency queries that, alongside ``t`` over-latency queries,
     lets you conclude at confidence ``c`` that the true p-percentile meets the bound.
-    ``_odds`` is monotonically decreasing in ``h``, so exponential bracketing (double
-    ``hi`` until it passes) followed by binary search finds the boundary exactly.
+    The pass test is strict (``odds < 1 - c``), matching LoadGen's boundary handling
```
viraatc, line 207: Small docstring nit: the summary line above (line 202) still reads `_odds(h, t, p, d) <= 1 - c`, while the implementation and this sentence use the strict `<`. Suggest updating it to:

```
"""Minimum ``h`` such that ``_odds(h, t, p, d) < 1 - c``.
```

so the one-liner matches the boundary parity this PR pinned.

src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
```diff
@@ -242,11 +241,22 @@ def _es_estimates(self, sorted_values) -> dict[str, float | None] | None:
...
+            # report (the publisher's finalized guard makes retries no-ops).
+            logger.exception(
+                "%s: early-stopping computation failed; omitting the map", self.name
+            )
+            return None
+
+    def _compute_es_estimates(
+        self, sorted_values, spec: EarlyStoppingSpec
+    ) -> dict[str, float | None]:
         if spec.percentiles is not None:  # explicit override: tests / offline analysis
```
viraatc, line 258: Suggest a test for the explicit override: with `grid_percentile_key` removed, the only override test left is the `(150.0,)` failure branch — nothing pins that e.g. `(99, 99.9)` yields keys `"99"`/`"99.9"` in descending order, or that `99` vs `99.0` don't become two map entries (the script dedupes on the parsed value; this path doesn't). A short test would keep the new keying from drifting. Non-blocking.

viraatc (commented):
Review notes — 3 small suggestions, all non-blocking. Also independently re-verified the bit-identical claim: old vs new `find_min_passing` across p50–p99.99, t=0..999 — zero diffs.

viraatc (approved):

src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
nvzhihanj: Valid — fixed in 6b39f02, though one level lower than suggested.

I put the guard in `es_percentile_estimate` rather than in the registry branch: it's the single function all three surfaces funnel through (grid path, explicit override, script), so one check makes them agree without the rule existing in two places. Domain is now `[1, 100)`, matching the script's own `< 1` cutoff exactly, with the same did-you-mean hint.

Behaviour on the path you flagged is what you asked for: `EarlyStoppingSpec(percentiles=(0.99,))` now raises into the `_es_estimates` best-effort catch and degrades to "map omitted, exact stats intact" — identical to the `150.0` case. Pinned by `test_override_rejects_fraction_style_percentiles`.

Grid-derived targets can't reach it (`es_targets_from_grid` filters to `>= 50`), and I audited every call site to confirm nothing legitimately passes a sub-1 value.

src/inference_endpoint/metrics/early_stopping.py
nvzhihanj: Correct, fixed in 6b39f02 — the summary line now reads `_odds(h, t, p, d) < 1 - c`, matching both the implementation and the parity sentence below it.

src/inference_endpoint/async_utils/services/metrics_aggregator/registry.py
nvzhihanj: Agreed — added `test_explicit_override_keys_and_dedupes` in 6b39f02, and it turned up the divergence you flagged parenthetically, so I closed that too rather than pinning it.

The override path now dedupes on the parsed value like the script's `--percentiles` fallback: `(99, 99.9, 99.0)` yields `["99.9", "99"]`, not three keys computing p99 twice. The first spelling still wins the key, so an int override stays `"99"` and the map keeps overlaying the grid 1:1 where that matters.

The test pins all three properties — `str()`-of-entry keying, descending order, one entry per value — and fails on the pre-fix code (three keys instead of two).

PR #432, opened 2026-07-29
fix(evaluation): do not score LCB samples 0.0 when the judge cannot run

liayan:
evaluate_livecodebench_worker swallows every exception and returns False, so import/setup failures get scored as wrong answers with subset status "ok". That currently always happens in this env: lcb_runner.prompts imports anthropic (not installed) and datasets>=3 cannot load the script-based code_generation_lite dataset. Observed: 335 samples "graded" 0/335 in 83s; the same outputs score 297/349 through lcb-service. A fake zero on 349/4388 samples is enough to drop a passing DeepSeek-R1 run below the MLPerf accuracy threshold.

- re-raise ImportError/FileNotFoundError/RuntimeError from the worker and the gather loop; the runner already marks the subset failed and sets complete=false
- call load_lcb_benchmark() once before spawning workers so setup errors fail fast with the real traceback

Sample-level grading failures (bad generated code, timeouts) still score as incorrect. This does not fix the local path itself (deps are still missing); it just stops it from reporting a wrong score.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass


arekay-nv (commented):
Thanks for this. 
Can you please add some tests? The one thing to look for is whether a single bad example (one that has a bad import) will cause the entire run to fail. I believe that is the reason for swallowing the exception and just counting it as a failure.

liayan:
> Thanks for this. Can you please add some tests? The one thing to look for is whether a single bad example (one that has a bad import) will cause the entire run to fail. I believe that is the reason for swallowing the exception and just counting it as a failure.

Thanks — tests added in 7f391fb.

On the single-bad-example case: generated code never runs in the worker process, so it can't raise ImportError there. evaluate_livecodebench passes it to the LCB grader, which executes it in its own sandboxed subprocess. A bad import in a sample comes back as a failed test result and gets scored 0, same as before. The except Exception: return False in the worker is kept for exactly that case.

The three re-raised types can only come from the judge env itself: ImportError from the lcb_runner imports (lcb_runner.prompts imports anthropic), FileNotFoundError from the missing submodule check, RuntimeError from datasets>=3 refusing to load code_generation_lite. These hit every sample or none — that's how 335/335 got scored 0.0 in 83s. The new pre-flight load_lcb_benchmark() call catches most of this before workers even start.

Tests:
1. worker returns (qid, False) on a generic exception — run continues
2. worker re-raises each of the three infra types
3. one failing sample scores 0, the other samples still grade
4. all workers hitting an infra error raises instead of returning a fake 0/N
5. broken load_lcb_benchmark raises before the pool is created

1 and 3 also pass on main, so the swallow behavior is unchanged. The pool is stubbed with an in-process fake so the tests don't depend on the multiprocessing start method.

pyproject.toml
```diff
@@ -279,5 +279,6 @@ exclude = [
 module = [
     "inference_endpoint.openai.openai_types_gen",
     "inference_endpoint.metrics.reporter",
+    "inference_endpoint.evaluation.legacy_mlperf_deepseek_r1.mlperf_eval.*",
```
roborluo: I would not recommend this workaround, it is functional but will lost track of format of this file. 

Also I locally ran this file with `mypy`, seems no error, not sure why you add this line? Did you see any error without this?

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
```diff
@@ -655,6 +661,10 @@ def process_livecodebench_parallel(
     if not work_items:
         return 0, 0
 
+    # Pre-flight: a broken judge (missing dependency/dataset) must raise here
+    # with a real traceback, not be scored as 0.0 by every worker.
+    load_lcb_benchmark()
```
roborluo, line 666: I am thinking, Is this really neessary ？ because in worker it also calls `load_lcb_benchmark`. Bascially you called this twice. 

And I think it fine to expose the problem a little bit later in each worker process to make it clean and non duplicated.

roborluo: Also another question is like `load_lcb_benchmark`  its name is confusing as well.  because here you are not just validate/initialize everything but also build and return the actual lookup map which is not used anywhere.

roborluo: So Better remove it

pyproject.toml
liayan: > I would not recommend this workaround, it is functional but will lost track of format of this file.
> 
> Also I locally ran this file with `mypy`, seems no error, not sure why you add this line? Did you see any error without this?

Agree; I put it there because of a CI error. So fixed the error directly and dropped the override here. Thanks for the review.

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
```diff
@@ -567,12 +567,18 @@ def evaluate_livecodebench(code: Optional[str], question_id: str) -> bool:
         os.environ.pop('TQDM_DISABLE', None)
 
 
+# Infrastructure failures (broken judge) must not be scored as wrong answers.
+_LCB_INFRA_ERRORS = (ImportError, FileNotFoundError, RuntimeError)
```
roborluo: nit: my personal perference is to put this type of global definition to the top of file. But fine to put it here.

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
```diff
@@ -777,6 +789,8 @@ def print_evaluation_results(df_evaluated: pd.DataFrame,
     print("\nResults\n")
     print(results)
 
+    return results
```
roborluo: Wonder why return?

tests/unit/evaluation/test_legacy_mlperf_eval_accuracy.py
```diff
@@ -0,0 +1,146 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 CoreWeave, Inc. All rights reserved.
```
roborluo, line 1: nit: maybe consider rename it to `test_legacy_mlperf_eval_accuracy.py` if possible.

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
liayan:
```
def print_evaluation_results(df_evaluated: pd.DataFrame,
                             logger: Optional[logging.Logger] = None) -> Dict[str, Any]:
```

No strong reason — the function claimed a dict return so I completed it, but nothing consumes the value. Happy to make it -> None

roborluo: Oh make sense!  I think if no one using its return value, let us make it none

src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/eval_accuracy.py
liayan: The call does more than validation. load_lcb_benchmark is lru_cached per-process: the parent loads it once, the fork workers inherit it, so it's one load instead of N. It also makes a broken judge fail in the parent with the original traceback, instead of later through the executor's RemoteTraceback.

Agreed the discarded return value reads oddly — I can extend the comment to make that clear. The worker-level re-raise is the actual fix though; this line is just a perf/diagnostics nicety, so I'm fine dropping it if you still prefer.

roborluo (approved):
LGTM, Thanks for the quick revision!

arekay-nv (approved):
Thanks for the fix.

liayan:
Thanks for the approval. I don’t have write permission, so I can’t merge it myself. Feel free to merge it when convenient, or once the merge window opens.

PR #433, opened 2026-07-29
fix(lcb-service): distinguish infra crashes from submission failures in grading children

liayan:
## What

Distinguishes judge-side (infra) failures from submission-side failures when
a LiveCodeBench grading child produces no result, instead of scoring
everything as a submission failure:

- **-6 GradingChildDied vs -8 SubmissionKilledChild**: a shared `started_flag`,
  flipped in `run_test` right after judge setup (`reliability_guard`, suite
  parse) and before the submission's own code runs, tells the parent whether
  the child died during the judge's setup (infra) or while running the
  submission (submission's fault).
- **-7 SubmissionExit**: `sys.exit()` in submitted code is a `BaseException`,
  so it previously fell through the child's `except Exception` and got
  misclassified as `GradingChildDied`. Caught separately now.
- **All-infra guard**: -5 `TestRunnerError` and -6 `GradingChildDied` are both
  judge-side; a batch where every future is infra-attributed now raises
  instead of silently reporting a 0 score.
- Malformed ground-truth JSON in the dataset is now a judge-side (-5)
  failure, parsed alongside the suite parse in `run_test` instead of being
  folded into the submission's -4 inside `grade_call_based`.
- `mp.Manager()` is used in a `with`-block and the killed child is
  `join()`ed, so a killed grading child no longer lingers as a zombie or
  leaks a manager process. `mp.Value` uses `lock=False` (single
  writer/reader, no lock needed).

This PR does **not** touch the multiprocessing start method (fork/spawn) —
that's split out into #457 (pinning the base image back to Python 3.11) to
avoid conflating a perf-sensitive change with this one.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- `pytest -m unit` — full suite passes, no regressions
- `tests/unit/evaluation/test_lcb_serve.py` (new): covers every attribution
  path — -1/-5/-6/-7/-8, the all-infra guard (including a mixed batch that
  must *not* trip it), both call-based and stdio, malformed ground truth
- Verified on real hardware: B300 (x86_64) and GB300 (aarch64) — full
  `tests/unit/evaluation/` suite (314 tests) passes on both

## Checklist

- [x] Tests added and passing
- [x] Existing grading semantics preserved (only affects the previously
      mis-attributed 0-result path)


liayan:
Verified on the Python 3.14 lcb-service image (forkserver default): grading works with the fork pin; dead grading children classify as -6, and the all-infra-errors check raises, so the original failure is still caught; an all-timeout batch scores 0 without raising — that case would have raised on commit one fix.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -270,8 +293,11 @@ def __call__(
         for qid, test_codes in zip(question_ids, codes, strict=False):
             results[qid] = [False] * len(test_codes)
         futures = {}
+        infra_errors = 0
 
-        with ProcessPoolExecutor(max_workers=self.n_lcb_workers) as executor:
+        with ProcessPoolExecutor(
+            max_workers=self.n_lcb_workers, mp_context=_MP_CTX
```
hvagadia: **High — concurrency:** `LCBServe.evaluate()` is invoked through `event_loop.run_in_executor()` in `_server.py`, so this outer `ProcessPoolExecutor` is created from a worker thread in the already multithreaded uvicorn process. Forcing it to use `fork` can inherit locks held by threads that disappear in the child, leaving grading workers deadlocked and the evaluation hung. The reported compatibility problem establishes that the inner grading child needs fork semantics, but not that the outer pool does. Please use separate contexts—for example, `forkserver` for this outer pool and `fork` only for the `Manager`/`Process` inside `run_code_subprocess`, where the pool worker is single-threaded.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -133,17 +144,29 @@ def run_code_subprocess(
...
+                "error_code": -1,
+                "error_message": f"Subprocess did not complete in time ({global_timeout}s)",
+            }
+        else:
+            # Exited before the deadline without reporting a result: the
+            # grading child died (e.g. bad start method, OOM kill), not the
+            # submitted code.
+            res = [-1] * len(suite["inputs"])
+            metadata = {
+                "error": "Grading subprocess died before reporting a result",
```
hvagadia: **Medium — correctness:** `execute_code_single_suppressed_errors()` catches `Exception`, but `SystemExit` derives from `BaseException`. A submitted solution that calls `sys.exit()` therefore terminates the grading child without filling `resp_buffer`, and this branch classifies it as `GradingChildDied (-6)`. For a one-sample batch—or when every submission exits—the all-infrastructure-error guard raises `RuntimeError` instead of recording ordinary failed submissions. This was reproduced through the actual LiveCodeBench path. Please catch `SystemExit` as a submission runtime failure and account for `os._exit`, which can bypass exception handling entirely.

liayan: Good catch — I did see SystemExit getting mixed in with GradingChildDied. Turns out it's specifically the call-based (fn_name) grading path, since the stdio path already guards against it internally. Added a SystemExit except guard as suggested (bc502f1), scoring it as its own -7 SubmissionExit instead of -6, so it doesn't get miscounted as an infra failure anymore.
Let me know if you still see any issues.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -133,17 +154,29 @@ def run_code_subprocess(
...
+                "error_code": -6,
+                "error_message": f"GradingChildDied (exitcode={p.exitcode})",
```
nv-alicheng: [Claude] **high** (data-integrity): The `-6 GradingChildDied` branch conflates a broken judge with a submission that kills its own grading process, and `-6` is in `_LCB_INFRA_ERROR_CODES` (line 64), so an all-crash batch trips the `infra_errors == len(futures)` RuntimeError (line 361) and **refuses to report a legitimately-0 score — the exact opposite of this PR's stated goal.** Untrusted submitted code runs *inside* the child (`_MP_CTX.Process(target=execute_code_single_suppressed_errors)` → `run_test(..., test=code)`), and the child only records a result at the very end (`resp_buffer.append(...)`, line 122). Any path terminating the interpreter before that append — and before the `SystemExit`/`except` handlers run — leaves `resp_buffer` empty with `p.is_alive()==False`, landing here as `-6`. Submission-controlled ways to hit it (all bypass Python exception handling, so the new SystemExit→-7 fix at :106 does NOT cover them): `os._exit(0)`; a native segfault (malicious/broken numpy, ctypes, C-extension); an OOM SIGKILL from a huge allocation. So `-6` is reachable by ordinary adversarial/bad model output, not just infra. Consequence: a batch where every submission crashes its interpreter aborts with RuntimeError instead of reporting the correct `pass@1=0` — discarding a *true* 0 score.

**Root cause / fix:** the infra-vs-submission distinction is asserted in a comment + a set literal, not derived from how the child died. Reserve `-6` for judge-startup failures (the fork/forkserver problem this PR targets) and classify submission-self-terminated deaths as a submission fault alongside `-7` — e.g. treat a clean `exitcode==0` empty-buffer as submission fault, and exclude negative/signal exit codes a submission can self-induce from `_LCB_INFRA_ERROR_CODES`. (Distinct from the already-fixed SystemExit miscount at :176.)

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -52,6 +52,17 @@
...
+# Grading assumes fork semantics; under Python 3.14's forkserver default the
+# grading children die at startup and the service reports 0/N forever.
+# This module only runs inside the Linux lcb-service container, so "fork"
+# is always available.
+_MP_CTX = mp.get_context("fork")
+
+# Error codes that mean the judge is broken, as opposed to the submitted code
+# failing its tests: -5 TestRunnerError, -6 GradingChildDied. Timeouts (-1)
+# are the submitted code's fault and do not count.
+_LCB_INFRA_ERROR_CODES = {-5, -6}
```
nv-alicheng, line 74: [Both] **medium** (code-quality): The judge error-code space is magic integers: `-1/-2/-5/-6/-7` appear as bare literals across the module (`res=[-2]`, `error_code:-7/-6/-5/-1`) with meaning living only in prose comments plus this set literal:
```
_LCB_INFRA_ERROR_CODES = {-5, -6}
```
This is the data-model root of the high-severity miscount above: the infra-vs-submission invariant the PR relies on is encoded in a comment + set membership rather than the type. Define one `IntEnum` (`TIMEOUT=-1, WRONG_ANSWER=-2, TEST_RUNNER_ERROR=-5, GRADING_CHILD_DIED=-6, SUBMISSION_EXIT=-7`) and derive `_LCB_INFRA_ERROR_CODES` from it, so the distinction is enforced structurally instead of by matching literals against a comment. (Codes also recur in run_lcb_tests.py.)

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -117,9 +138,9 @@ def run_code_subprocess(
         suite["inputs"]
     ) + flat_timeout_extension
 
-    manager = mp.Manager()
+    manager = _MP_CTX.Manager()
```
nv-alicheng: [Quality] **medium** (code-quality): `manager = _MP_CTX.Manager()` creates a full multiprocessing Manager (a separate server process) for every single code sample and never shuts it down — no `manager.shutdown()`, no `with` — so it's only reclaimed by GC finalizer and manager processes leak under load. This compounds with the fork change: `run_code_subprocess` already runs inside a forked `ProcessPoolExecutor` worker, so each grade is pool-worker-fork → Manager-server-fork → grading-Process-fork, nested fork-from-fork each carrying inherited parent state. Wrap in `with _MP_CTX.Manager() as manager:` (or reuse one manager per `_LCBWorker` batch) to bound process count deterministically.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -52,6 +52,17 @@
...
+# is always available.
+_MP_CTX = mp.get_context("fork")
+
+# Error codes that mean the judge is broken, as opposed to the submitted code
+# failing its tests: -5 TestRunnerError, -6 GradingChildDied. Timeouts (-1)
+# are the submitted code's fault and do not count.
+_LCB_INFRA_ERROR_CODES = {-5, -6}
+
 
 def execute_code_single(test_suite_json: str, code: str, timeout_sec: int = 60):
```
nv-alicheng: [Quality] **low** (code-quality): Missing return-type annotations on the three grading-path functions that cross the process boundary: `execute_code_single` (67), `execute_code_single_suppressed_errors` (91), `run_code_subprocess` (126) — all return `tuple[list, dict]` (results + metadata) but none declare it, so the `res, metadata = ...` unpack the whole error-code protocol depends on is unchecked. Relatedly, the fork target at line 91 is `def execute_code_single_suppressed_errors(*args, resp_buffer=None, **kwargs)` — fully untyped variadics exactly where `_MP_CTX.Process(target=..., args=..., kwargs=...)` wires args, so a wrong positional/keyword fails only at runtime inside the child (surfacing as a spurious `-6`). Give both concrete signatures / a named result-tuple alias.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -133,17 +154,29 @@ def run_code_subprocess(
     p.start()
     p.join(timeout=global_timeout)
 
-    if p.is_alive():
+    timed_out = p.is_alive()
+    if timed_out:
         p.kill()
 
     if len(resp_buffer) == 0:
```
nv-alicheng: [Both] **low** (code-quality): The two no-result branches duplicate `res = [-1] * len(suite["inputs"])` plus a same-shape `{error, error_code, error_message}` dict (timeout arm 164-169, child-died arm 174-179), and use an `else`-after-`return`. A guard clause flattens it: `if resp_buffer: return resp_buffer[0]` up front, then handle the empty case with no nested `else`; hoist the shared `res` assignment and set only the differing metadata per branch.

nv-alicheng (commented):

Reviewed by: **Claude** + **Code-Quality** (×2: diff + import-neighborhood, per request) | Depth: **quick** + forced code-quality

`codex` was unavailable in this environment. See the summary comment for neighborhood findings on `_server.py`/`run_lcb_tests.py` and untouched-line items that can't be posted inline.

nv-alicheng:

Reviewed by: **Claude** + **Code-Quality** (run twice — diff scope + import-neighborhood, per request) | Depth: **quick** with code-quality **forced on** (normally skipped at quick depth)
(`codex` CLI unavailable in this environment.)

Tight, well-reasoned fix. One **high-severity correctness gap** survives it, plus code-quality/neighborhood items you asked for. Existing bot/human threads corroborated but not duplicated (see bottom).

### 🔴 Must Fix (high)

| File | Line | Category | Reviewer | Summary |
|------|------|----------|----------|---------|
| `lcb_serve.py` | 178 | data-integrity | Claude | `os._exit()` / segfault / OOM-kill in submitted code dies before the result-append → classified `-6 GradingChildDied` → counted as infra → an **all-crash batch trips the RuntimeError and discards a true `pass@1=0`**, the opposite of the PR's goal. Bypasses the new SystemExit→`-7` fix (those never reach an `except`). Reserve `-6` for judge-startup failures; classify submission-induced exits as a submission fault like `-7`. |

### 🟡 Should Fix (medium)

| File | Line | Category | Reviewer | Summary |
|------|------|----------|----------|---------|
| `lcb_serve.py` | 64 | code-quality | Both | Magic error codes (`-1/-2/-5/-6/-7`) + prose-encoded infra invariant → `IntEnum`, derive `_LCB_INFRA_ERROR_CODES` from it. This is the **data-model root of the high finding above.** |
| `lcb_serve.py` | 141 | code-quality | Quality | `_MP_CTX.Manager()` per sample, never `shutdown()` → leaks manager server processes; nested fork-from-fork under the pool. Use `with`. |
| `lcb_serve.py` | 71 | code-quality | Quality (nbhd) | Lazy `import numpy` / `from .run_lcb_tests import run_test` inside the fork target — violates the repo no-lazy-imports rule; hoisting also warms the modules in the parent before fork. _(untouched line — not inline)_ |
| `lcb_serve.py` | 513 | code-quality | Quality (nbhd) | `evaluate_dataframe` mutates the caller's DataFrame in place (`df["extracted_code"] = ...fillna("")`) — hidden side effect on a caller-owned object. Assign to a local. _(untouched line)_ |
| `_server.py` | 405 | code-quality | Quality (nbhd) | Websocket handler passes the module-global `lcb_serve` (typed `LCBServe \| None`) into `EvaluationSession` with no None-guard → opaque `AttributeError` later inside `run_in_executor`; `/info` already 503-guards. Also `asyncio.get_event_loop()` in a coroutine is deprecated → `get_running_loop()`. _(other file — not in PR diff)_ |
| `_server.py` | 32 | code-quality | Quality (nbhd) | `from lib.lcb_serve import LCBServe` — `lib.` prefix doesn't match the module's real location (same dir as `_server.py`); an implicit, undocumented container-packaging contract that `ModuleNotFoundError`s if run in-place. Prefer `from .lcb_serve import LCBServe` or document the `lib` packaging. _(other file)_ |

### 🔵 Consider (low)

| File | Line | Category | Reviewer | Summary |
|------|------|----------|----------|---------|
| `lcb_serve.py` | 67 | code-quality | Quality | Missing `-> tuple[list, dict]` on the 3 grading funcs (67/91/126); fork target at 91 uses untyped `*args/**kwargs` where `Process(...)` wires args → wrong wiring fails only at runtime as `-6`. |
| `lcb_serve.py` | 161 | code-quality | Both | DRY: the two no-result branches duplicate `res`/metadata shape; a `if resp_buffer: return ...` guard clause flattens the `else`-after-`return`. |
| `lcb_serve.py` | 459 | code-quality | Quality (nbhd) | `assert self.df is not None` as a public-method precondition — stripped under `python -O`; raise explicitly. _(untouched line)_ |
| `run_lcb_tests.py` | 491 | code-quality | Quality (nbhd) | `except ValueError as e: raise e` loses the traceback (use bare `raise`) and the following `in_outs = None` is unreachable dead code — on the exact `-5 TestRunnerError` path the PR relies on. _(other file)_ |

### Existing threads (corroborated, not duplicated)
- **`lcb_serve.py:309` (hvagadia, high/concurrency)** — forcing `fork` on the *outer* `ProcessPoolExecutor`, created from a uvicorn worker thread via `run_in_executor`, can inherit thread-held locks and deadlock. **Still live and unaddressed** — the neighborhood Manager-leak/nested-fork finding (141) sits in the same fork-safety area and reinforces it. Not re-filed.
- **`lcb_serve.py:176` (hvagadia/liayan, SystemExit)** — **already fixed by this PR** (`bc502f1`, `-7 SubmissionExit`). The high finding above is the *residual* case (`os._exit`/signals) that fix cannot reach.

> ⚠️ **Commit hygiene:** 9 commits including 4 apparent fixups. Consider squashing before merge.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
nv-alicheng: Oop - this is an artifact of a direct copy/paste port from the official LCB runner repo. I can take the task to clean it up to fit code standards.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: > **High — concurrency:** `LCBServe.evaluate()` is invoked through `event_loop.run_in_executor()` in `_server.py`, so this outer `ProcessPoolExecutor` is created from a worker thread in the already multithreaded uvicorn process. Forcing it to use `fork` can inherit locks held by threads that disappear in the child, leaving grading workers deadlocked and the evaluation hung. The reported compatibility problem establishes that the inner grading child needs fork semantics, but not that the outer pool does. Please use separate contexts—for example, `forkserver` for this outer pool and `fork` only for the `Manager`/`Process` inside `run_code_subprocess`, where the pool worker is single-threaded.

Good catch — tried forkserver as suggested, it passed locally but hung at pool shutdown in the lcb-service container (Python 3.14.5, unreaped zombie workers), so I switched the outer pool to spawn instead: worker startup is a bit slower but it's thread-safe for the same reason and all local + container tests pass; the grading child keeps fork. Let me know if you see any other issues.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: I will keep using the number for now.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: Good catch — os._exit/segfault/OOM do bypass the -7 guard entirely. Went with a shared started-flag instead of exit codes (a submission can fake any exit code via os._exit): the child flips it right before grading starts, so pre-flag deaths stay -6 infra and post-flag deaths become -8 SubmissionKilledChild, scored as a normal fail. Verified in the container that an all-os._exit batch now reports a true 0 instead of tripping the guard (83db6ed).

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: Done — early return for the reported-result case and hoisted the shared res (35feb8e); the duplication was honestly my own doing from stacking the attribution branches one review round at a time.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: Fair point — this actually predates this PR, but worth tightening while we're in here: wrapped it in a with-block and added a join() after kill so a killed grading child gets reaped instead of lingering as a zombie (1918fd7).

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: Done — gave the fork target named parameters and declared tuple[list, dict] on all three grading helpers (7f4d040); the variadics were from previous prs but agreed they hid wiring mistakes until runtime.

nv-alicheng:
One thing to note - I think this is a bug on my part: I'd used py3.14 as the base container for the lcb_runner, but maybe this also should be studied:

From the lcb_runner repo, they use py3.11 (https://github.com/LiveCodeBench/LiveCodeBench)
```
uv venv --python 3.11
source .venv/bin/activate

uv pip install -e .
```

I have personally not checked but there might be some variance between python versions (i.e. version specific language features like walrus operator, same-line multi-context managers, etc.), and the code generated by the worker should be either version-agnostic or catered to a specific version.

This can get solved a little more easily if we just pin to a python 3.11 container for lcb-service.

liayan:
> One thing to note - I think this is a bug on my part: I'd used py3.14 as the base container for the lcb_runner, but maybe this also should be studied:
> 
> From the lcb_runner repo, they use py3.11 (https://github.com/LiveCodeBench/LiveCodeBench)
> 
> ```
> uv venv --python 3.11
> source .venv/bin/activate
> 
> uv pip install -e .
> ```
> 
> I have personally not checked but there might be some variance between python versions (i.e. version specific language features like walrus operator, same-line multi-context managers, etc.), and the code generated by the worker should be either version-agnostic or catered to a specific version.
> 
> This can get solved a little more easily if we just pin to a python 3.11 container for lcb-service.

3.14's forkserver default is what exposed this, but rolling back to 3.11 would just hide it — fork-from-thread, the -6/-8 attribution, and the SystemExit handling are all version-independent bugs that'd still exist, so I don't see much upside to reverting now.

liayan:
Kindly ping.

arekay-nv:
> 3.14's forkserver default is what exposed this, but rolling back to 3.11 would just hide it — fork-from-thread, the -6/-8 attribution, and the SystemExit handling are all version-independent bugs that'd still exist, so I don't see much upside to reverting now.

Lets also revert to 3.11 to keep it consistent with LCB specification. We can do that separately.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -78,16 +101,39 @@ def execute_code_single(test_suite_json: str, code: str, timeout_sec: int = 60):
...
+    resp_buffer: list | None = None,
+    started_flag: Synchronized | None = None,
+) -> tuple[list, dict]:
     """Wrapper around execute code so that all errors are resurfaced as failed tests"""
+    if started_flag is not None:
+        # From here on the wrapper is running and grading (i.e. the
+        # submission) is about to execute; if the interpreter dies now it is
+        # the submission's doing. Deaths before this point are judge startup
+        # failures.
+        started_flag.value = True
```
arekay-nv: **high (correctness):** `started_flag` is flipped `True` at wrapper entry — *before* the judge's own setup runs (`import numpy` / `from .run_lcb_tests import run_test` in `execute_code_single`, `reliability_guard()`, and the trusted-input `json.loads`/parse in `grade_*`). A hard interpreter death in that window (segfault on a bad numpy/arch, OOM `SIGKILL`, `os._exit`) is therefore attributed to the submission as `-8 SubmissionKilledChild`. Because `-8 ∉ _LCB_INFRA_ERROR_CODES`, `infra_errors` stays below `len(futures)`, the all-infra `RuntimeError` guard never fires, and a systemically broken judge reports a silent `pass@1 = 0` — the exact failure this PR exists to prevent.

This is the mirror/residual of the `os._exit`→`-6` fix (83db6ed): that fix moved the attribution boundary one frame too high. The flag should bracket **only** the untrusted submission's execution.

Fix: set `started_flag.value = True` immediately before the submission actually executes — push it into `run_test`/`grade_*`, after imports + `reliability_guard()` + suite parse. Hoisting the two lazy imports in `execute_code_single` to module scope (so they run in the parent before the fork) shrinks the window further and also satisfies the no-lazy-imports rule. The line-113 comment ("submission … is about to execute") is itself inaccurate, since trusted judge setup still runs first.

Independently flagged high by the concurrency reviewer and critical by Grok 4.5.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -117,37 +163,64 @@ def run_code_subprocess(
...
         }
-        return res, metadata
+    elif started:
+        # The interpreter died while grading was running (os._exit(),
+        # segfault, OOM, ...). Grading executes the untrusted submission, so
+        # this is the submission's fault, not the judge's.
+        metadata = {
+            "error": "Grading child killed while executing the submission",
+            "error_code": -8,
+            "error_message": f"SubmissionKilledChild (exitcode={exitcode})",
```
arekay-nv, line 211: **high (testing):** The PR description and checklist claim *"Added a regression test proving submitted code that exits its grading child remains a failed sample."* No such test exists — `grep -rn` over `tests/` for `lcb_serve` / `LCBServe` / `run_code_subprocess` / `started_flag` / `SubmissionKilledChild` / `_LCBWorker` returns zero matches (the only `livecodebench` hits are unrelated dataset/report strings), and codecov's "all covered" is a degenerate no-BASE artifact. The spawn/fork split and the `-7`/`-8` attribution — including this exact `-8` path — ship untested, against AGENTS.md's >90% bar. Please add: an all-`os._exit` batch → true `0` (no raise), and an all-`-6` batch → raises.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -290,9 +366,9 @@ def __call__(
                 qid, code_idx = futures[future]
                 res, metadata = future.result()
                 if "error" in metadata:
-                    logger.warning(
-                        f"Test execution error for question {qid}: {metadata}"
-                    )
+                    if metadata.get("error_code") in _LCB_INFRA_ERROR_CODES:
+                        infra_errors += 1
+                    logger.error(f"Test execution error for question {qid}: {metadata}")
```
arekay-nv: **medium (error-handling):** `warning` → `error` here fires for *every* sample whose metadata has an `"error"` key — including routine outcomes: `-1` timeouts, `-7 SubmissionExit`, `-8 SubmissionKilledChild`, and ordinary submission runtime errors. On a normal accuracy batch (many slow/buggy submissions) this floods ERROR and makes the genuine `-5`/`-6` judge-broken signals indistinguishable in ops. The infra-vs-submission split already exists (`_LCB_INFRA_ERROR_CODES`) — gate the level on it: ERROR for infra codes, WARNING/INFO for submission-attributed ones.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -317,6 +393,17 @@ def __call__(
                             exc_info=True,
                         )
 
+        # Every subprocess hitting an infra error means the judge itself is
+        # broken, not that every code sample failed its tests. Timeouts are
+        # excluded: a batch where every submission loops forever is a valid
+        # 0 score, not a broken judge.
+        if futures and infra_errors == len(futures):
```
arekay-nv, line 406: **medium (error-handling / accuracy):** The all-infra guard is narrower than its intent:

- It only fires when **every** future is `-5`/`-6`. A single legitimate non-infra result among N (a real `-2`/`-7`/timeout) drops `infra_errors` below `len(futures)`, so a judge broken for a *subset* of samples is silently folded into the score.
- `-8` (a post-`started_flag` crash — see the line-117 thread) and `-1` (timeout) are excluded, so a wholesale judge crash during setup escapes entirely.
- The classification gate keys on `"error" in metadata` (line 368), but LCB's own outer handler returns `[-4]` with only `error_code`/`error_message` and **no `"error"` key** (`run_lcb_tests.py:521, 541`). A systemic exception inside `grade_*` → every sample `-4` → not logged, not counted infra → silent `pass@1 = 0`.

Consider deriving the infra set structurally (an `IntEnum`) and treating "no result / judge-side death / grade-time exception" as infra regardless of the `"error"` key.

arekay-nv (commented):

Reviewers: **Codex** (gpt-5.6-sol) + **Claude** (concurrency + accuracy lenses) + **Grok 4.5** (Cursor). Event: comment only (no approve/reject). Findings verified across models; a couple of single-model claims were dropped after checking (e.g. an "open `Manager()` makes the fork worker multi-threaded" claim — empirically the worker stays single-threaded, so the `spawn`-pool / `fork`-child split is sound).

This PR has iterated well: the earlier `SystemExit`→`-7`, outer-pool fork→`spawn`, Manager-leak→`with`+`join`, return-types, and DRY threads are all **resolved** — not re-raised here.

### Posted inline (still open)
| Line | Sev | Finding |
|------|-----|---------|
| 117 | high | `started_flag` flips before judge-side setup → a judge-setup crash is mislabeled `-8` and **escapes the all-infra guard → silent pass@1=0** (mirror of the 83db6ed fix; boundary one frame too high). |
| 213 | high | The regression test the PR body claims does not exist anywhere in `tests/`. |
| 400 | med | All-infra guard is narrow: `-8`/`-1` excluded, a single non-infra result disables it, and LCB's `-4 "Error during testing"` (`run_lcb_tests.py:521/541`) carries no `"error"` key so it's never classified/logged. |
| 371 | med | `warning`→`error` floods ERROR with routine timeouts/submission failures, drowning genuine `-5`/`-6`. |

### Lower priority (not posted inline)
- **Lazy imports** in `execute_code_single` (`numpy`, `run_lcb_tests`) — hoisting to module scope also shrinks the line-117 window (two birds).
- **PR description drift:** the body says fork is used for the "outer process pool," but the code uses `spawn` there (fork only for the grading child) — please update.
- **Document the invariant** that the `fork` child is safe *only* while the pool worker is single-threaded, so a future top-level import that spawns a thread doesn't silently reintroduce a fork+lock deadlock.
- `started_flag`'s `Value` can be `lock=False` (parent reads only post-`join`); the per-sample `Manager` could be a `Pipe`/`SimpleQueue` — hardening only.

### Filed separately
- A pre-existing accuracy bug found during this review — an **empty test suite scores every submission as PASS** (`all([]) == True`) — is tracked in #443 (out of scope for this PR's diff).

### Orthogonal (existing thread)
The Python-3.11 discussion: pinning would only mask the (real, version-independent) MP bugs, so keep these fixes — but grading-interpreter parity with upstream LCB is a separate *result-validity* concern worth its own tracking, not a substitute.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: Fix as suggested, moved the flag into run_test itself, set right after reliability_guard()/suite parse and before dispatching into grade_call_based/grade_stdio, so it only brackets the submission's own execution; hoisted the two lazy imports to module scope like you suggested (f866eb7).

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: Forgot to commit last time, but added tests/unit/evaluation/test_lcb_serve.py with exactly those two plus the rest of the -1/-6/-7/-8 attribution paths, 11 tests total (3241a29).

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: Done — split it exactly that way: ERROR for infra codes, WARNING for submission-attributed ones (225598a).

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: Added the missing error key so -4 at least logs (225598a), but kept it out of _LCB_INFRA_ERROR_CODES on purpose — that outer catch is reached almost entirely by submissions that fail to compile, which is squarely the submission's fault, not the judge's; flagging it infra would abort ordinary "model wrote broken code" batches. The all-or-nothing threshold and -8/-1 exclusion are real gaps but a false-positive/negative tradeoff on the guard, so I'd rather get your take before picking a number than just change it.

liayan: IntEnum's the same ask as before with @nv-alicheng

> Oop - this is an artifact of a direct copy/paste port from the official LCB runner repo. I can take the task to clean it up to fit code standards.

tests/unit/evaluation/test_lcb_serve.py
```diff
@@ -0,0 +1,236 @@
...
+These tests call run_code_subprocess directly (not through _LCBWorker's
+spawn pool) so that monkeypatches applied in the test process are inherited
+by the forked grading child; the batch-level guard tests below need the
+real pool and monkeypatch _MP_POOL_CTX to fork for the same reason.
+"""
+
+import multiprocessing as mp
+import os
+
+import pytest
```
viraatc, line 47: minor-nit: Missing `pytestmark = pytest.mark.unit` — every other file in this dir has it.

liayan: Done.

arekay-nv:
@liayan can you add more details on how to reproduce the failures. I have tried to run LCB with py3.11 and py3.14 base images on x86/linux and unable to see any failures without the PR. Can you share the steps to reproduce the `caused grading children to die at startup and left evaluation at 0/N.`

liayan:
> @liayan can you add more details on how to reproduce the failures. I have tried to run LCB with py3.11 and py3.14 base images on x86/linux and unable to see any failures without the PR. Can you share the steps to reproduce the `caused grading children to die at startup and left evaluation at 0/N.`

Good catch — I originally hit this on vera-rubin with a new lcb-service image on 3.14, but also confirmed it reproduces on GB200/GB300. Just haven't checked on x86 yet. Just reproduced it again

Steps:

1. Get any aarch64 Linux host with Python 3.14 (default forkserver start method) — I used an enroot/pyxis container on a GB300 NVL72 node, but the actual lcb-service image isn't required; a stock python:3.14-slim reproduces it too.
2. Run any workload shaped like ProcessPoolExecutor → mp.Manager() → mp.Process(), all default context — no LCB code, dataset, or server needed. I used a trivial driver with 8 pool workers and 32 dummy "grading" samples.
3. Watch it stall with zero output. ps -eo pid,ppid,stat,cmd on the node while it's stuck:
```
274448  274444 S  /app/venv/bin/python -c from multiprocessing.forkserver import main; main(10, 11, ...)
274455  274445 S  /app/venv/bin/python -c from multiprocessing.forkserver import main; main(8, 9, ...)
274463  274455 Sl /app/venv/bin/python -c from multiprocessing.forkserver import main; main(8, 9, ...)
274481  274458 Z  [python] <defunct>
274484  274459 Z  [python] <defunct>
274487  274461 Z  [python] <defunct>
274490  274462 Z  [python] <defunct>
274495  274457 Z  [python] <defunct>
274506  274455 Z  [python] <defunct>
274515  274456 Z  [python] <defunct>
274518  274460 Z  [python] <defunct>
```

Grading children dead as <defunct> zombies under a nested tree of forkserver helper processes — same shape as the original 0/349-for-3.5h, one defunct child per pool worker.

liayan:
Interesting — same exact steps on x86_64 (same Python 3.14,  same code): finishes in ~1s, every time, at every scale I tried (up to 349 samples / 176 workers). So this looks architecture-specific rather than Python-only-specific.

liayan:
Discussed offline with @arekay-nv — we're moving the explicit spawn/fork context pinning to a separate follow-up PR with more performance data, to avoid a perf regression on the spawn side. Everything else through c14c3c2 (SystemExit handling, started_flag attribution, Manager with-block, type annotations, logging levels) stays as-is here.

Verified in a python:3.11-slim container (Python < 3.14, defaults to fork) — all 11 regression tests pass.

Also opened #457 to pin the base image back to Python 3.11, mirroring LiveCodeBench's own supported version. Could you take a look and merge that one too — it's still blocking lcb-service on aarch64 without the new spawn soluton/fork solution.

liayan:
> Discussed offline with @arekay-nv — we're moving the explicit spawn/fork context pinning to a separate follow-up PR with more performance data, to avoid a perf regression on the spawn side. Everything else through [c14c3c2](https://github.com/mlcommons/endpoints/commit/c14c3c2c5ac3476539e438ad6af92dc2da0ba980) (SystemExit handling, started_flag attribution, Manager with-block, type annotations, logging levels) stays as-is here.
> 
> Verified in a python:3.11-slim container (Python < 3.14, defaults to fork) — all 11 regression tests pass.
> 
> Also opened #457 to pin the base image back to Python 3.11, mirroring LiveCodeBench's own supported version. Could you take a look and merge that one too — it's still blocking lcb-service on aarch64 without the new spawn soluton/fork solution.

Hey @arekay-nv, could you give this another review and approve it if everything looks good? I’d also appreciate your help merging it, since I don’t have write permissions. I opened #457 as well for consistency.

src/inference_endpoint/evaluation/livecodebench/run_lcb_tests.py
```diff
@@ -518,7 +528,11 @@ def run_test(sample, test=None, timeout=6):
                 )
                 return results, metadata
             except Exception as e:
```
arekay-nv, line 533: **medium (data-integrity):** This `-4` path attributes *judge/dataset-side* failures to the submission. `grade_call_based` parses the **ground-truth** outputs with `all_outputs = [json.loads(o) for o in all_outputs]` at `run_lcb_tests.py:298`, **outside** any `try` — so if the dataset's expected-output JSON is malformed it raises here and is labelled `-4` ("submission's fault"). But a malformed *suite* (`sample["input_output"]`) fails earlier at line 491 and propagates to the wrapper's `except Exception` → `-5` (infra, which trips the all-infra guard). The same "bad dataset" fault is thus split across a guard-tripping code (`-5`) and a non-tripping one (`-4`), so a systematically malformed dataset silently reports a 0 without the guard firing — the inverse of this PR's judge-vs-submission goal. Consider parsing ground-truth data before the submission runs and attributing its failures to infra.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -290,9 +359,19 @@ def __call__(
                 qid, code_idx = futures[future]
                 res, metadata = future.result()
                 if "error" in metadata:
-                    logger.warning(
-                        f"Test execution error for question {qid}: {metadata}"
-                    )
+                    if metadata.get("error_code") in _LCB_INFRA_ERROR_CODES:
```
arekay-nv, line 365: **medium (testing):** Two coverage gaps around the infra taxonomy this line keys on. (1) `-5 TestRunnerError` (the in-process judge-crash path from `execute_code_single_suppressed_errors`'s `except Exception`, line 137) is never exercised — every infra test forces `-6` via `os._exit` *before* that `try` runs, so `-5`, its `traceback` payload, and its participation here and in the guard are untested. (2) The guard's `infra_errors == len(futures)` boundary is only tested at its all-or-nothing extremes; a **mixed** batch (some `-5`/`-6`, some pass/fail) — which must *not* fire the guard yet still ERROR-log the infra sample — is uncovered, so a regression to `>=` or an `infra_errors` miscount would pass the entire suite.

src/inference_endpoint/evaluation/livecodebench/run_lcb_tests.py
```diff
@@ -538,7 +552,10 @@ def run_test(sample, test=None, timeout=6):
...
+                # Same as the call_based branch above: a compile/definition
+                # failure in the submission's own code, not a judge bug.
                 return [-4], {
+                    "error": repr(e),
```
arekay-nv, line 561: **low (testing):** This stdio `-4` block (and the whole `grade_stdio` branch) is never executed by the new tests — every suite constant is `CALL_BASED` (`fn_name` set), so `which_type` is always `call_based`. The `"error": repr(e)` key added here is therefore untested. A single stdio suite (drop `fn_name`, e.g. an echo program) would cover the untested half cheaply.

arekay-nv (commented):
## Review council — follow-up pass (round 2)

Ran an independent multi-model council (Codex `gpt-5.6-sol` @ xhigh + two Claude review lenses) over the current HEAD (`7038582`). **Codex found no actionable regressions.** The attribution core verified correct: the `started_flag` boundary is placed right (after `reliability_guard()` / suite-parse, before `grade_*`), the three no-result branches are mutually exclusive and each reachable, `SystemExit`→`-7` handling is correct, and the `with mp.Manager()` switch fixes a pre-existing per-sample Manager-process leak.

New findings posted inline. My earlier round-1 comments (incl. the all-infra guard narrowness at line 403) still stand and are not re-raised here.

| Sev | Where | Issue |
|---|---|---|
| medium | `run_lcb_tests.py:530` | Malformed ground-truth outputs → `-4` (submission) while malformed suite → `-5` (infra): split attribution, dataset faults can silently report 0 |
| medium | `lcb_serve.py:362` | `-5` path and the guard's mixed-batch `==` boundary untested |
| low | `run_lcb_tests.py:558` | stdio / `grade_stdio` branch untested |
| low | `lcb_serve.py:164` | `mp.Value` default `lock=True` — latent deadlock + per-sample semaphore; use `lock=False` |
| low | `test_lcb_serve.py:124` | Docstrings narrate dev-history / reference an unmerged "follow-up PR" (AGENTS.md) |

Additional low-severity, not posted inline:
- **`lcb_serve.py:383`** — `all([])` is `True`, so an empty/degenerate suite (`LCBTestLoader(strict=False)` returns `inputs: []`) scores a free pass; guard with `bool(res) and all(...)`. Pre-existing.
- **`lcb_serve.py:360`** — `future.result()` is unguarded; a `BrokenProcessPool`, or a raise inside `run_code_subprocess` (malformed-suite `json.loads`, Manager spawn failure), bypasses the guard with an opaque crash. Pre-existing.
- **Fork-only fault injection** — `-6`/`-8` attribution is exercised only under forced `fork`; the `forkserver`/`spawn` method that caused the production incident isn't directly tested. Worth confirming the fork-pinned tests pass on Python 3.14 / macOS (fork-after-threads).
- **Docs** — the new `-5..-8` taxonomy and the refuse-to-report-0 `RuntimeError` aren't documented in `livecodebench/README.md` or `docs/evaluation/DESIGN.md`.

<sub>🤖 Generated by an AI review council (Codex + Claude); posted for the author's consideration.</sub>

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
```diff
@@ -117,37 +158,64 @@ def run_code_subprocess(
...
-            "resp_buffer": resp_buffer,
-            "timeout_sec": timeout_sec,
-        },
-    )
-    p.start()
-    p.join(timeout=global_timeout)
+    with mp.Manager() as manager:
+        resp_buffer = manager.list()
+        # typeshed types ctx.Value() as SynchronizedBase, which lacks .value
+        started_flag = cast(Synchronized, mp.Value(ctypes.c_bool, False))
```
arekay-nv: **low (concurrency):** `mp.Value(ctypes.c_bool, False)` defaults to `lock=True`, allocating a semaphore per sample and making every `.value` access acquire it. The parent reads `started_flag.value` (line 188) *after* `p.kill()`; POSIX semaphores aren't robust, so a child SIGKILLed while holding this lock would block the parent's read forever. Unreachable today (the child holds the lock only for the microsecond write at `run_lcb_tests.py:515`, long before the timeout kill), but the flag is single-writer / single-reader-after-join, so `lock=False` removes the deadlock surface and a per-sample semaphore allocation for free.

tests/unit/evaluation/test_lcb_serve.py
```diff
@@ -0,0 +1,270 @@
...
+        lcb_serve.execute_code_single_suppressed_errors = orig
+
+    assert metadata["error_code"] == -6
+    assert metadata["error_code"] in lcb_serve._LCB_INFRA_ERROR_CODES
+
+
+def test_death_during_judge_setup_is_infra_error_not_submission(monkeypatch):
+    """A death inside run_test's own setup (reliability_guard, suite parse)
+    -- after run_test has started but before started_flag is set -- must
+    still be -6. This is the exact boundary the started_flag fix moved: it
```
arekay-nv: **low (convention):** This docstring narrates development history, which AGENTS.md explicitly bars ("describe current state, not development history" — the rule even calls out regression-test docstrings). "the exact boundary the started_flag fix moved: it used to flip True on wrapper entry" describes a prior implementation no longer in the tree; the invariant (a death during judge setup, before `started_flag` is set, is `-6`) can be stated without it. Same file, the module docstring at line 40 references "a follow-up PR" — a dangling unmerged-artifact pointer that AGENTS.md's "no local-only artifacts" rule bars; cite an issue/PR number or drop the forward reference.

arekay-nv (approved):
Looks good. Thanks for putting this together and addressing the review comments. 
There are just a final few ones - please address them before merging.

src/inference_endpoint/evaluation/livecodebench/run_lcb_tests.py
liayan: Good catch. Moved the outputs/inputs parse into run_test next to the suite parse, so a malformed dataset is judge-side (-5) instead of landing in grade_call_based's except as -4.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: Added both: a plain-exception judge-setup test scoring -5, and a mixed -6/real-failure batch test so the all-infra guard's boundary is actually exercised.

src/inference_endpoint/evaluation/livecodebench/run_lcb_tests.py
liayan: Added a stdio suite (no fn_name) hitting the same compile-error path, so grade_stdio's -4 branch is covered too.

tests/unit/evaluation/test_lcb_serve.py
liayan: Trimmed both — the test docstring now states the invariant only (no more "the fix moved..."), and dropped the "follow-up PR" reference from the module docstring at the top.

src/inference_endpoint/evaluation/livecodebench/lcb_serve.py
liayan: Changed to lock=False — single writer (child), single reader (parent after join), no lock needed.

liayan:
All follow-up items from the new round review are addressed in a442550 — see replies on each thread. Ran the full tests/unit/evaluation/ suite on both B300 and GB300 hardware; no regression.

@arekay-nv the new commit landed but seems still need another reviewer approval per requrement, could you help merge it directly or should I ping other reviewers here.

arekay-nv:
@liayan sure will do - can you update the PR description as well - it still seems to reference the fork vs spawn changes.

liayan:
> @liayan sure will do - can you update the PR description as well - it still seems to reference the fork vs spawn changes.

Done. 
Please also give https://github.com/mlcommons/endpoints/pull/457 a review when convenient.

viraatc (approved):

PR #437, opened 2026-08-04
refactor(benchmark): remove redundant accuracy_datasets from BenchmarkContext

arekay-nv:
## Summary

- Removes `accuracy_datasets: list[Dataset]` from `BenchmarkContext` — it was a parallel copy of data already carried by `eval_configs: list[AccuracyConfiguration]` via the `.dataset` field
- `_load_datasets` return type drops the middle `list[Dataset]` element; all call sites updated
- `_summarize_and_log_metrics` `acc_total` now uses `ec.num_repeats` (the canonical `AccuracyConfiguration` field, consistent with `accuracy.py`) and filters by `dataset_type == ACCURACY` explicitly rather than relying on an implicit invariant
- Restores three dataset-level sample assertions in override tests (`eval_configs[0].dataset.load_sample(0)`) that were dropped during the migration

## Test plan

- [ ] `uv run pytest tests/unit/commands/test_benchmark.py` — 178 tests pass
- [ ] All override tests (`TestLoadDatasetsGenerationConfigOverrideChat`, `TestLoadDatasetsGenerationConfigOverrideCompletions`) pass with restored sample-level assertions


roborluo (approved):
LGTM

leopck (approved):

PR #438, opened 2026-08-04
Add Pyxis runtime for SWE-bench accuracy

hvagadia:
## Summary

- add an optional Pyxis runtime to the SWE-bench accuracy service
- run mini-swe-agent tool commands and SWE-bench evaluation through `srun` and Enroot on a retained one-node Slurm allocation
- resolve task images from a registry prefix using the documented `sweb.eval.arm64.<instance_id>:v4.1.0-arm64` convention
- preserve Docker as the default runtime and leave the benchmark client configuration unchanged

## Why

ARM64 Slurm nodes provide Pyxis and Enroot but may not provide Docker. This allows the existing external SWE-bench accuracy service to run agent and evaluator containers natively on those nodes without changing its HTTP API or the endpoint client YAML.

## Validation

- `uv run pytest -q tests/unit/evaluation/swebench_service` — 72 passed
- targeted `pre-commit` checks for all five changed files — passed

### DIFF
--- a/src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
+++ b/src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
@@ -28,6 +28,17 @@ _SAFE_SRUN_ENV = (
     "TMPDIR",
     "XDG_RUNTIME_DIR",
 )
+_STEP_STATUS = "/tmp/.mlperf_srun_status"
+_STEP_SCRIPT = r"""set +e
+status_path=$1
+timeout_s=$2
+shift 2
+printf 'started\n' > "$status_path"
+timeout "$timeout_s" "$@"
+returncode=$?
+printf 'finished:%s\n' "$returncode" > "$status_path"
+exit "$returncode"
+"""
 
 
 def safe_srun_env() -> dict[str, str]:
@@ -45,9 +56,17 @@ def build_srun_command(
     job_id = os.environ.get("SLURM_JOB_ID", "").strip()
     if not job_id:
         raise RunnerError("Pyxis runtime requires SLURM_JOB_ID")
-    command = ["srun", "--overlap", f"--jobid={job_id}", "-N1", "-n1"]
-    if node := os.environ.get("SLURMD_NODENAME", "").strip():
-        command.append(f"--nodelist={node}")
+    node = os.environ.get("SLURMD_NODENAME", "").strip()
+    if not node:
+        raise RunnerError("Pyxis runtime requires SLURMD_NODENAME")
+    command = [
+        "srun",
+        "--overlap",
+        f"--jobid={job_id}",
+        "-N1",
+        "-n1",
+        f"--nodelist={node}",
+    ]
     if image is not None:
         image_ref = str(image.resolve()) if isinstance(image, Path) else image
         command.append(f"--container-image={image_ref}")
@@ -75,6 +94,59 @@ def build_srun_command(
     return command
 
 
+def run_srun_step(
+    *,
+    argv: list[str],
+    status_path: Path,
+    timeout_s: int,
+    failure_path: Path | None = None,
+    image: str | Path | None = None,
+    name: str | None = None,
+    mounts: list[tuple[Path, str]] | None = None,
+    workdir: str | None = None,
+    stderr: int = subprocess.STDOUT,
+) -> subprocess.CompletedProcess[str]:
+    status_path.write_text("pending\n")
+    status_path.chmod(0o666)
+    command = build_srun_command(
+        image=image,
+        name=name,
+        mounts=mounts,
+        workdir=workdir,
+        argv=[
+            "bash",
+            "-c",
+            _STEP_SCRIPT,
+            "pyxis-step",
+            _STEP_STATUS,
+            str(timeout_s),
+            *argv,
+        ],
+    )
+    try:
+        result = subprocess.run(
+            command,
+            text=True,
+            encoding="utf-8",
+            errors="replace",
+            stdout=subprocess.PIPE,
+            stderr=stderr,
+            timeout=timeout_s + 30,
+            env=safe_srun_env(),
+        )
+    except (OSError, subprocess.SubprocessError) as exc:
+        if failure_path is not None:
+            failure_path.touch()
+        raise RunnerError(
+            "Pyxis infrastructure failure before the command completed"
+        ) from exc
+    if status_path.read_text().strip() != f"finished:{result.returncode}":
+        if failure_path is not None:
+            failure_path.touch()
+        raise RunnerError("Pyxis infrastructure failure before the command completed")
+    return result
+
+
 def resolve_image(image_registry: str, instance_id: str) -> str:
     if Path(instance_id).name != instance_id or instance_id in {".", ".."}:
         raise RunnerError(f"invalid SWE-bench instance ID: {instance_id}")
@@ -84,7 +156,7 @@ def resolve_image(image_registry: str, instance_id: str) -> str:
         if not separator:
             raise RunnerError("Pyxis image registry must include a repository")
         image_registry = f"{host}#{repository}"
-    return f"{image_registry}/sweb.eval.arm64.{instance_id}:v4.1.0-arm64"
+    return f"{image_registry}/sweb.eval.arm64.{instance_id.lower()}:v4.1.0-arm64"
 
 
 class PyxisEnvironmentConfig(BaseModel):
@@ -94,6 +166,7 @@ class PyxisEnvironmentConfig(BaseModel):
     env: dict[str, str] = Field(default_factory=dict)
     timeout: int = 30
     interpreter: list[str] = Field(default_factory=lambda: ["bash", "-c"])
+    infrastructure_failure_path: Path | None = None
 
 
 class PyxisEnvironment:
@@ -106,23 +179,18 @@ class PyxisEnvironment:
         self._tmp_dir.chmod(0o1777)
         self._lock = threading.Lock()
         self._cleaned = False
-        command = build_srun_command(
-            image=self.config.image,
-            name=self.name,
-            mounts=[(self._tmp_dir, "/tmp")],
-            workdir=self.config.cwd,
-            argv=["true"],
-        )
         try:
-            subprocess.run(
-                command,
-                check=True,
-                capture_output=True,
-                text=True,
-                timeout=self.config.timeout,
-                env=safe_srun_env(),
+            run_srun_step(
+                image=self.config.image,
+                name=self.name,
+                mounts=[(self._tmp_dir, "/tmp")],
+                workdir=self.config.cwd,
+                argv=["true"],
+                status_path=self._tmp_dir / Path(_STEP_STATUS).name,
+                timeout_s=self.config.timeout,
+                failure_path=self.config.infrastructure_failure_path,
             )
-        except (OSError, subprocess.SubprocessError) as exc:
+        except RunnerError as exc:
             self.cleanup()
             raise RunnerError(
                 f"failed to start Pyxis container for {self.config.image}"
@@ -135,42 +203,31 @@ class PyxisEnvironment:
         argv = ["env"]
         argv.extend(f"{key}={value}" for key, value in self.config.env.items())
         argv.extend([*self.config.interpreter, command])
-        srun_command = build_srun_command(
-            image=None,
+        result = run_srun_step(
+            argv=argv,
+            status_path=self._tmp_dir / Path(_STEP_STATUS).name,
+            timeout_s=timeout or self.config.timeout,
+            failure_path=self.config.infrastructure_failure_path,
             name=self.name,
             mounts=[(self._tmp_dir, "/tmp")],
             workdir=cwd or self.config.cwd,
-            argv=argv,
         )
         output: dict[str, Any]
-        try:
-            result = subprocess.run(
-                srun_command,
-                text=True,
-                timeout=timeout or self.config.timeout,
-                encoding="utf-8",
-                errors="replace",
-                stdout=subprocess.PIPE,
-                stderr=subprocess.STDOUT,
-                env=safe_srun_env(),
-            )
+        if result.returncode == 124:
+            output = {
+                "output": result.stdout,
+                "returncode": -1,
+                "exception_info": "The command timed out",
+                "extra": {
+                    "exception_type": "TimeoutExpired",
+                    "exception": f"command timed out after {timeout or self.config.timeout}s",
+                },
+            }
+        else:
             output = {
                 "output": result.stdout,
                 "returncode": result.returncode,
                 "exception_info": "",
-            }
-        except Exception as exc:
-            raw_output = getattr(exc, "output", None)
-            if isinstance(raw_output, bytes):
-                raw_output = raw_output.decode("utf-8", errors="replace")
-            output = {
-                "output": raw_output or "",
-                "returncode": -1,
-                "exception_info": f"An error occurred while executing the command: {exc}",
-                "extra": {
-                    "exception_type": type(exc).__name__,
-                    "exception": str(exc),
-                },
             }
         lines = output.get("output", "").lstrip().splitlines(keepends=True)
         if (
@@ -178,6 +235,7 @@ class PyxisEnvironment:
             and lines[0].strip() == "COMPLETE_TASK_AND_SUBMIT_FINAL_OUTPUT"
             and output["returncode"] == 0
         ):
+            # mini-swe-agent is installed only in the SWE-bench service subproject.
             from minisweagent.exceptions import Submitted
 
             submission = "".join(lines[1:])
@@ -240,4 +298,7 @@ class PyxisEnvironment:
         try:
             self.cleanup()
         except Exception:
-            pass
+            logging.getLogger(__name__).warning(
+                "Could not clean up Pyxis environment",
+                exc_info=True,
+            )
--- a/src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
+++ b/src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
@@ -14,9 +14,11 @@ import uuid
 from pathlib import Path
 from typing import Any
 
-from pydantic import BaseModel, Field
+from pydantic import AliasChoices, BaseModel, Field
 
 from .runner import RunnerError
+
+logger = logging.getLogger(__name__)
 
 _SAFE_SRUN_ENV = (
     "PATH",
@@ -34,7 +36,7 @@ status_path=$1
 timeout_s=$2
 shift 2
 printf 'started\n' > "$status_path"
-timeout "$timeout_s" "$@"
+unshare --pid --fork --mount-proc timeout "$timeout_s" "$@"
 returncode=$?
 printf 'finished:%s\n' "$returncode" > "$status_path"
 exit "$returncode"
@@ -164,7 +166,11 @@ class PyxisEnvironmentConfig(BaseModel):
     run_id: str
     cwd: str = "/testbed"
     env: dict[str, str] = Field(default_factory=dict)
-    timeout: int = 30
+    timeout_s: int = Field(
+        default=30,
+        validation_alias=AliasChoices("timeout_s", "timeout"),
+        serialization_alias="timeout",
+    )
     interpreter: list[str] = Field(default_factory=lambda: ["bash", "-c"])
     infrastructure_failure_path: Path | None = None
 
@@ -180,6 +186,7 @@ class PyxisEnvironment:
         self._lock = threading.Lock()
         self._cleaned = False
         try:
+            # A no-op initializes and validates the named persistent container.
             run_srun_step(
                 image=self.config.image,
                 name=self.name,
@@ -187,7 +194,7 @@ class PyxisEnvironment:
                 workdir=self.config.cwd,
                 argv=["true"],
                 status_path=self._tmp_dir / Path(_STEP_STATUS).name,
-                timeout_s=self.config.timeout,
+                timeout_s=self.config.timeout_s,
                 failure_path=self.config.infrastructure_failure_path,
             )
         except RunnerError as exc:
@@ -200,13 +207,14 @@ class PyxisEnvironment:
         self, action: dict[str, Any], cwd: str = "", *, timeout: int | None = None
     ) -> dict[str, Any]:
         command = action.get("command", "")
+        logger.debug("Executing Pyxis command: %s", command)
         argv = ["env"]
         argv.extend(f"{key}={value}" for key, value in self.config.env.items())
         argv.extend([*self.config.interpreter, command])
         result = run_srun_step(
             argv=argv,
             status_path=self._tmp_dir / Path(_STEP_STATUS).name,
-            timeout_s=timeout or self.config.timeout,
+            timeout_s=timeout or self.config.timeout_s,
             failure_path=self.config.infrastructure_failure_path,
             name=self.name,
             mounts=[(self._tmp_dir, "/tmp")],
@@ -220,7 +228,9 @@ class PyxisEnvironment:
                 "exception_info": "The command timed out",
                 "extra": {
                     "exception_type": "TimeoutExpired",
-                    "exception": f"command timed out after {timeout or self.config.timeout}s",
+                    "exception": (
+                        f"command timed out after {timeout or self.config.timeout_s}s"
+                    ),
                 },
             }
         else:
@@ -250,7 +260,7 @@ class PyxisEnvironment:
 
     def get_template_vars(self, **kwargs: Any) -> dict[str, Any]:
         return {
-            **self.config.model_dump(),
+            **self.config.model_dump(by_alias=True),
             **platform.uname()._asdict(),
             **kwargs,
         }
@@ -259,7 +269,7 @@ class PyxisEnvironment:
         return {
             "info": {
                 "config": {
-                    "environment": self.config.model_dump(mode="json"),
+                    "environment": self.config.model_dump(mode="json", by_alias=True),
                     "environment_type": (
                         f"{self.__class__.__module__}.{self.__class__.__name__}"
                     ),
@@ -286,7 +296,7 @@ class PyxisEnvironment:
                         env=safe_srun_env(),
                     )
                 except (OSError, RunnerError, subprocess.SubprocessError):
-                    logging.getLogger(__name__).warning(
+                    logger.warning(
                         "Could not remove Pyxis container %s",
                         self.name,
                         exc_info=True,
@@ -298,7 +308,7 @@ class PyxisEnvironment:
         try:
             self.cleanup()
         except Exception:
-            logging.getLogger(__name__).warning(
+            logger.warning(
                 "Could not clean up Pyxis environment",
                 exc_info=True,
             )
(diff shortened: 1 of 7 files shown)


src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_worker.py
```diff
@@ -0,0 +1,245 @@
...
+            }
+        )
+
+    with concurrent.futures.ThreadPoolExecutor(
+        max_workers=args.max_workers
+    ) as executor:
+        futures = [
+            executor.submit(_evaluate_instance, **payload) for payload in payloads
+        ]
+        for future in concurrent.futures.as_completed(futures):
```
nv-alicheng, line 216: [Both] **high** (data-integrity): `_run_eval` swallows every per-instance failure (`except Exception: ... print(...)`) and `main()` then unconditionally calls `make_run_report`, which writes a result file even if zero instances were evaluated. The parent `PyxisSweBenchRunner._run_eval` (runner.py:747) only checks the file exists and returns success. So an **infrastructure fault** — srun scheduling failure, missing `SLURM_JOB_ID`, node down, registry auth failure (all raise/exit non-zero per instance) — is indistinguishable from a genuinely unresolved instance: the service reports a *successful* run with a silently degraded 0%/partial score. For a compliance-grade accuracy tool this turns a broken cluster run into a passing-shaped artifact with an undercounted score. There is no worker→runner propagation path for "all instances failed for infra reasons." **This is the root of the two medium findings below (pyxis_worker.py:138, pyxis_environment.py:162)** — fix the propagation here (distinguish infra failure from unresolved; fail the run when instances error for non-eval reasons) and both symptom-patches downstream become unnecessary.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_worker.py
```diff
@@ -0,0 +1,245 @@
...
+    result = subprocess.run(
+        command,
+        capture_output=True,
+        text=True,
+        timeout=timeout_s + 60,
+        env=safe_srun_env(),
+    )
+    with _PRINT_LOCK:
+        print(f"[{instance_id}]\n{result.stdout}{result.stderr}", flush=True)
+    if result.returncode != 0:
```
nv-alicheng: [Both] **medium** (error-handling): `_evaluate_instance` drops the instance on ANY non-zero container exit with no report and no signal (`if result.returncode != 0: return`). `report_path.unlink(missing_ok=True)` ran at the top, so a missing report makes `make_run_report` count the instance unresolved. But the eval script returns non-zero for a failed patch apply (1), test timeout (124), AND whenever the srun/enroot layer itself fails to run the step. A transient container/enroot/node error therefore permanently marks a possibly-correct patch unresolved, with the only trace being printed stdout. No retry, no infra-vs-patch-fail distinction. Raise (the caller has a try/except) or write an explicit error report so the two cases stay distinguishable.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
```diff
@@ -0,0 +1,243 @@
...
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                env=safe_srun_env(),
+            )
+            output = {
+                "output": result.stdout,
+                "returncode": result.returncode,
+                "exception_info": "",
+            }
+        except Exception as exc:
```
nv-alicheng: [Both] **medium** (error-handling): `execute` converts any subprocess exception — including `subprocess.TimeoutExpired` from a stuck/queued srun — into a synthetic command result with `returncode: -1`. Reasonable for surfacing a shell failure to the agent, but a host-side srun scheduling/timeout failure is then presented to the agent as an ordinary failed command rather than aborting the instance; combined with the swallow above, cluster-contention failures are laundered into "the model's command failed," silently biasing the trajectory. At minimum make a `TimeoutExpired` on the container-start/step path distinguishable from a real command exit. (The broad `except Exception` also masks genuine command-assembly bugs as `-1` — narrow it or log unexpected types.)

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_worker.py
```diff
@@ -0,0 +1,245 @@
...
+    base_agent = swebench.ProgressTrackingAgent
+
+    class LiveTrajectoryAgent(base_agent):  # type: ignore[misc, valid-type]
+        def __init__(self, *agent_args: Any, instance_id: str = "", **kwargs: Any):
+            kwargs["output_path"] = (
+                args.output / instance_id / f"{instance_id}.live.json"
+            )
+            super().__init__(*agent_args, instance_id=instance_id, **kwargs)
+
+    swebench.get_sb_environment = get_pyxis_environment
```
nv-alicheng: [Both] **medium** (design): `_run_agent` injects the Pyxis env by reassigning upstream module globals: `swebench.get_sb_environment = get_pyxis_environment` / `swebench.ProgressTrackingAgent = LiveTrajectoryAgent`. It works because the worker is a one-shot subprocess, but it's a fragile abstraction leak / hidden global mutable state: any upstream rename or pre-capture of these names breaks silently, with no assertion the attributes existed before overwrite. Add a guard (`assert hasattr(swebench, "get_sb_environment")`) so the coupling fails loudly on upstream drift, or isolate it behind a documented shim.

src/inference_endpoint/evaluation/swebench_service/swebench_service/runner.py
```diff
@@ -616,3 +627,187 @@ def _run_eval(
...
+        self._run_logged_subprocess(
+            command,
+            run_dir / "swe_bench_eval.log",
+            cwd=output_dir,
+            timeout_s=self.subprocess_timeout_s,
+            env=env,
+            secret_values=secret_values,
+            cancel_token=cancel_token,
+        )
+        result_path = (
```
nv-alicheng: [Claude] **medium** (bug): The Pyxis eval result-path resolution is stricter than Docker's and can spuriously fail a valid run. Docker `_run_eval` (619-629) falls back to `rglob(f"*{run_id}*.json")` when the exact `{safe_model}.{run_id}.json` isn't at top level; the Pyxis override only checks the exact path and raises `RunnerError("...result file not found...")` otherwise. If upstream ever changes the report filename or writes into a subdir, Pyxis raises while Docker recovers. Share the resolution logic or document why Pyxis can assume the exact name.

nv-alicheng (commented):

Reviewed by: **Claude** + **Code-Quality** | Depth: **thorough**

`codex` was unavailable in this environment — Claude + Code-Quality review. See the summary comment for the tiered breakdown.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
```diff
@@ -0,0 +1,243 @@
...
+                        self.name,
+                        exc_info=True,
+                    )
+        finally:
+            self._tmp.cleanup()
+
+    def __del__(self) -> None:
+        try:
+            self.cleanup()
+        except Exception:
```
nv-alicheng, line 310: [Quality] **medium** (code-quality): `__del__` does `except Exception: pass` with neither a comment nor a logging statement — violates the repo rule that every except must explain why the exception is ignored or log it. Add a one-line justification for discarding finalizer errors.

src/inference_endpoint/evaluation/swebench_service/swebench_service/runner.py
```diff
@@ -616,3 +627,187 @@ def _run_eval(
...
+        return result_path
+
+    def _cleanup_containers(
+        self,
+        run_id: str,
+        *,
+        eval_run_id: str | None = None,
+        instance_ids: list[str] | None = None,
+    ) -> None:
+        from .pyxis_environment import build_srun_command, safe_srun_env
```
nv-alicheng, line 751: [Quality] **medium** (code-quality): Function-level import (`from .pyxis_environment import build_srun_command, safe_srun_env`). The repo bans lazy imports except documented circular-import avoidance. `pyxis_environment` imports `RunnerError` from `runner` at top level, so this is presumably circular-avoidance — but the required documenting comment is absent. Add it (or restructure). The same unsanctioned lazy-import pattern recurs at pyxis_environment.py:181 (`from minisweagent.exceptions import Submitted`) and in pyxis_worker.py — guard those as container-only optional deps with a comment.

tests/unit/evaluation/swebench_service/test_runner.py
```diff
@@ -559,3 +602,562 @@ def fake_run(cmd, **kwargs):
...
+    command, kwargs = calls[0]
+    assert command[1:4] == ["-m", "swebench_service.pyxis_worker", "eval"]
+    assert command[command.index("--max-workers") + 1] == "2"
+    assert command[command.index("--image-registry") + 1] == _PYXIS_IMAGE_REGISTRY
+    assert command[command.index("--instance-ids") + 1 :] == ["repo__repo-1"]
+    assert "OPENAI_API_KEY" not in kwargs["env"]
+    assert result.exists()
+
+
+def test_pyxis_eval_does_not_mount_report_directory(monkeypatch, tmp_path):
```
nv-alicheng, line 1270: [Claude] **medium** (testing): `_evaluate_instance` is only exercised on the non-zero early-return path (this test). The success branch — returncode 0 → `get_eval_report(...)` → `atomic_write_bytes(report_path, ...)` (pyxis_worker.py:141-149) — is never tested, so the report-writing path, the report_path/mount wiring, and JSON serialization are unverified. This is the branch that actually produces the graded output. Mock `get_eval_report` and assert `report.json` is written atomically with expected content.

tests/unit/evaluation/swebench_service/test_runner.py
```diff
@@ -559,3 +602,562 @@ def fake_run(cmd, **kwargs):
...
+    monkeypatch.setenv("HF_TOKEN", "dataset-secret")
+
+    environment = safe_srun_env()
+
+    assert "OPENAI_API_KEY" not in environment
+    assert "SWEBENCH_SERVICE_AUTH_TOKEN" not in environment
+    assert "HF_TOKEN" not in environment
+
+
+def test_pyxis_environment_reuses_named_writable_container(monkeypatch, tmp_path):
```
nv-alicheng: [Claude] **medium** (testing): `PyxisEnvironment.execute` has two untested branches core to correctness: (1) the exception path (pyxis_environment.py:162-174) — no test drives `subprocess.run` to raise, so the `returncode:-1`/`exception_info` synthesis and bytes-decoding of `exc.output` are unverified; (2) `Submitted` detection (175-190) — no test feeds `COMPLETE_TASK_AND_SUBMIT_FINAL_OUTPUT` as first stdout line, so the submission-extraction that decides when a run is complete is entirely unexercised. Also untested: the `build_srun_command` comma-in-mount guard (pyxis_environment.py:69) and `resolve_image` missing-repository guard (85) — add negative cases so these srun-arg guards can't silently regress.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+class PyxisEnvironment:
+    def __init__(self, **kwargs: Any):
+        self.config = PyxisEnvironmentConfig(**kwargs)
+        safe_run_id = re.sub(r"[^A-Za-z0-9_.-]", "-", self.config.run_id)[:24]
+        self.name = f"mswe_{safe_run_id}_{uuid.uuid4().hex[:8]}"
+        self._tmp = tempfile.TemporaryDirectory(prefix=f"pyxis_{self.name}_")
+        self._tmp_dir = Path(self._tmp.name)
+        self._tmp_dir.chmod(0o1777)
```
nv-alicheng, line 185: [Claude] **low** (security): The per-run staging tmp dir is made world-writable: `self._tmp_dir.chmod(0o1777)`, then bind-mounted into the container as `/tmp` with `--container-remap-root`. On a shared SLURM node this dir (under the default tempfile root) is writable by every user on the node for the run's lifetime; the sticky bit stops deletion of others' files but not read/write of predictable filenames the eval steps use (e.g. `swebench_patch.diff`). Prefer a per-user path (`$XDG_RUNTIME_DIR`/`$TMPDIR`) with 0o700, or justify why 0o1777 is required for the remapped-root container to write.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
```diff
@@ -0,0 +1,243 @@
...
+                    raise RunnerError("Pyxis mount paths cannot contain commas")
+                specs.append(f"{source_text}:{destination}")
+            command.append("--container-mounts=" + ",".join(specs))
+        if workdir is not None:
+            command.append(f"--container-workdir={workdir}")
+    command.extend(argv)
+    return command
+
+
+def resolve_image(image_registry: str, instance_id: str) -> str:
```
nv-alicheng, line 152: [Claude] **low** (bug): `resolve_image` does not lowercase `instance_id`, while the Docker cleanup path does (`runner.py:511`, `instance_id.lower()`). Registry image path components are case-sensitive and SWE-bench image tags are conventionally lowercased, so any uppercase instance ID (or a future subset) yields `sweb.eval.arm64.{instance_id}:v4.1.0-arm64` that won't match the pushed lowercase image → image-pull failure. Latent today (Verified/Lite IDs are lowercase) but inconsistent with Docker — normalize.

src/inference_endpoint/evaluation/swebench_service/swebench_service/runner.py
```diff
@@ -616,3 +627,187 @@ def _run_eval(
...
+                        timeout=30,
+                        env=safe_srun_env(),
+                    )
+        except (OSError, subprocess.SubprocessError) as exc:
+            raise RunnerError(
+                f"failed to clean up Pyxis containers for SWE-bench run {run_id}"
+            ) from exc
+
+
+def create_runner(
```
nv-alicheng, line 783: [Quality] **low** (code-quality): `create_runner(runtime: str, ...)` dispatches on a stringly-typed flag with a runtime `raise` for the illegal value, though `__main__.py` already constrains it via `choices=("docker","pyxis")`. A `Literal["docker","pyxis"]` (or small enum) on the parameter makes the illegal state unrepresentable at the type boundary and lets mypy catch a bad caller.

src/inference_endpoint/evaluation/swebench_service/swebench_service/runner.py
```diff
@@ -616,3 +627,187 @@ def _run_eval(
...
+    def _run_eval(
+        self,
+        request: RunRequest,
+        preds_path: Path,
+        output_dir: Path,
+        run_dir: Path,
+        secret_values: set[str],
+        cancel_token: CancellationToken | None = None,
+    ) -> Path:
+        run_id = f"endpoints_{uuid.uuid4().hex[:8]}"
```
nv-alicheng: [Quality] **low** (code-quality): `PyxisSweBenchRunner._run_eval` copy-pastes non-trivial logic from base `SweBenchRunner._run_eval` (582-589): the `run_id = f"endpoints_{uuid.uuid4().hex[:8]}"` + persist step and the identical `{"verified": "princeton-nlp/SWE-bench_Verified", "lite": "princeton-nlp/SWE-bench_Lite"}.get(request.subset)` map. DRY: factor run-id generation and the subset→dataset map into a shared helper/constant instead of duplicating across both overrides.

src/inference_endpoint/evaluation/swebench_service/swebench_service/runner.py
```diff
@@ -616,3 +627,187 @@ def _run_eval(
...
+            cancel_token=cancel_token,
+        )
+        result_path = (
+            output_dir / f"{request.model_name.replace('/', '__')}.{run_id}.json"
+        )
+        if not result_path.exists():
+            raise RunnerError(f"SWE-bench result file not found for run_id={run_id}")
+        return result_path
+
+    def _cleanup_containers(
```
nv-alicheng, line 743: [Claude] **low** (concurrency): `PyxisSweBenchRunner._cleanup_containers` sweeps leaked agent containers by listing enroot on a single node (`build_srun_command` pins `--nodelist=$SLURMD_NODENAME`). Correct for the README's advertised one-node allocation, but there's no assertion or comment tying the single-node cleanup to that invariant. If the allocation is ever >1 node (agent workers `srun --overlap`'d onto siblings), named `pyxis_mswe_*` containers on other nodes leak silently. Add a comment (or guard) documenting the single-node cleanup invariant. Relatedly, the README (line 20) should state the service must run inside a live SLURM allocation (`SLURM_JOB_ID`/`SLURMD_NODENAME` set) and that running outside one yields degraded — not errored — results, given failures are swallowed.

nv-alicheng:

Reviewed by: **Claude** + **Code-Quality** | Depth: **thorough**
(`codex` CLI unavailable in this environment — Claude + Code-Quality only.)

New Pyxis/SLURM SWE-bench subsystem. Subprocess calls use list-argv form with a sanitized env (`safe_srun_env`) — **no `shell=True`, low command-injection risk**. Both reviewers converged strongly on one theme: **infrastructure failures are silently swallowed and laundered into low accuracy scores** rather than propagated. 14 issues posted (26 raw findings from the two reviewers, deduped/merged).

### 🔴 Must Fix (high)

| 1 | `pyxis_worker.py` | 195 | data-integrity | Both | Per-instance failures swallowed + `make_run_report` always writes success → an infra fault (srun fail, missing `SLURM_JOB_ID`, node down, registry auth) becomes a **passing-shaped artifact with an undercounted score**. Root cause of #2/#3. |

### 🟡 Should Fix (medium)

| 2 | `pyxis_worker.py` | 138 | error-handling | Both | Non-zero container exit dropped with no report/signal; no infra-vs-patch-fail distinction, no retry |
| 3 | `pyxis_environment.py` | 162 | error-handling | Both | `execute` launders srun `TimeoutExpired`/exceptions into `returncode:-1` → cluster contention looks like "model's command failed" |
| 4 | `pyxis_worker.py` | 70 | design | Both | Monkeypatches upstream `swebench` globals with no `hasattr` guard — breaks silently on upstream rename |
| 5 | `runner.py` | 747 | bug | Claude | Pyxis result-path lacks Docker's `rglob` fallback → spurious `RunnerError` if upstream report filename changes |
| 6 | `pyxis_environment.py` | 242 | code-quality | Quality | `__del__` bare `except Exception: pass` — violates repo except-comment rule |
| 7 | `runner.py` | 761 | code-quality | Quality | Function-level import w/o circular-avoidance comment (repo bans lazy imports); recurs at pyxis_environment.py:181 & pyxis_worker |
| 8 | `test_runner.py` | 1042 | testing | Claude | `_evaluate_instance` success/report-write branch untested (only the nonzero path is) |
| 9 | `test_runner.py` | 764 | testing | Claude | `execute` exception path + `Submitted` detection + comma/missing-repo guards untested |

### 🔵 Consider (low)

| 10 | `pyxis_environment.py` | 106 | security | Claude | Staging tmp dir `chmod 0o1777` world-writable, bind-mounted as container `/tmp` on shared node |
| 11 | `pyxis_environment.py` | 78 | bug | Claude | `resolve_image` doesn't lowercase `instance_id` (Docker path does) → latent image-pull mismatch |
| 12 | `runner.py` | 793 | code-quality | Quality | Stringly-typed `runtime` → `Literal["docker","pyxis"]` |
| 13 | `runner.py` | 706 | code-quality | Quality | DRY: subset→dataset map + run-id gen duplicated across `_run_eval` overrides |
| 14 | `runner.py` | 754 | concurrency | Claude | Single-node enroot cleanup invariant undocumented; + README should state SLURM-allocation precondition (degraded, not errored, outside one) |

**Also raised, dropped as minor at this cap:** magic image arch/version constants (`pyxis_environment.py:87`), hardcoded FQ class-name string repeated in two modules (`runner.py:651`, `pyxis_worker.py:84`), `_cleanup_containers` override discarding unused params (`runner.py:763`), `**kwargs: Any` erasing the typed `PyxisEnvironmentConfig` (`pyxis_environment.py:100`).

> ⚠️ **Commit hygiene:** 11 commits including 2 apparent fixups. Consider squashing before merge.

src/inference_endpoint/evaluation/swebench_service/swebench_service/templates/swebench_template.yaml
```diff
@@ -126,7 +126,7 @@ environment:
     TQDM_DISABLE: "1"
   environment_class: docker
   pull_timeout: 3600
-  container_timeout: 10h
+  container_timeout: 6h
```
leopck: why is the container_timeout lowered? I guess we don't really need 10h but just wondering if there's an issue that requires a lowered container timeout

leopck:
@hvagadia looking through this code, for the pyxis environment and worker support, as we are adding support into mini-swe agent itself, I would think it would make sense to upstream support for pyxis into mini-swe agent and open a PR in their repo instead for the pyxis support. https://github.com/SWE-agent/mini-swe-agent/tree/main/src/minisweagent/environments seems like the correct place. This would be great for the wider audience at Nvidia as well since others at Nvidia would be able to reuse your components for the pyxis support unblocking others.

Right now my biggest concern is that there is some monkey patched codes:

```py
swebench.get_sb_environment = get_pyxis_environment
swebench.ProgressTrackingAgent = LiveTrajectoryAgent
```

it will cause this module to break if the API changes. If we want to adopt this into this repo, I think we must at the minimum do an assert to pin this version of mini-swe of this library support to ensure that users don't try to use other versions of mini-swe with this library e.g. (`assert minisweagent.__version__`) But this will incur cost of maintenance into endpoints repo and tied down to a very locked version which should be our last resort for this in the event upstream doesn't accept it.

Upstreaming this module would be the path of less maintenance. In the meantime we could sustain a fork while we wait for the upstream to accept this PR and only if the upstream maintainers do not accept, then I believe we should sustain this.

I'll let @nv-alicheng to top up any thoughts on this as well. Let me know what you think about this @hvagadia.

Btw, I'm only referring to the agent part, but the eval part of this code look justified and it does look like something that we should own. (`_evaluate_instance + _EVAL_SCRIPT`)

nv-alicheng:
Agreed with @leopck - the monkeypatching seems a little hacky to me (I believe this is the Medium - 4 in my agent review post), which I think should be maybe bumped up.

Is it possible to make some child class to override and use that instead, or add it directly to swebench / make a fork?

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
hvagadia: We are keeping `01777` intentionally. This directory is bind-mounted as the container `/tmp` under `--container-remap-root`, so it needs standard `/tmp` semantics for the remapped container processes. The Pyxis service is required to run on a compute node inside a user-exclusive Slurm allocation, so other users cannot schedule workloads on that node; same-user processes are already within the service owner trust boundary. Tightening this to `0700` could prevent the remapped container UID from writing without providing practical cross-user isolation for the supported deployment.

hvagadia: TLDR node access is usually exclusive so this is not gonna be a problem. Changing perms might break the workflow within container.

src/inference_endpoint/evaluation/swebench_service/swebench_service/templates/swebench_template.yaml
hvagadia: Reverted

hvagadia:
Summary from today's discussion: I will go ahead and try to create a PR to mini-swe agent and test out the waters. Independent of that, I think we should merge this PR to unblock our experiments. If in future mini swe accepts our proposal we will update accordingly. @leopck @nv-alicheng

tianmu-li (approved):

leopck:
> I think we should merge this PR to unblock our experiments. If in future mini swe accepts our proposal we will update accordingly

Agreed, let's unblock and get this merged. If the PR does get merged upstream, we can always remove it from the codebase.

leopck (approved):

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
```diff
@@ -0,0 +1,304 @@
...
+    if Path(instance_id).name != instance_id or instance_id in {".", ".."}:
+        raise RunnerError(f"invalid SWE-bench instance ID: {instance_id}")
+    image_registry = image_registry.rstrip("/")
+    if "#" not in image_registry:
+        host, separator, repository = image_registry.partition("/")
+        if not separator:
+            raise RunnerError("Pyxis image registry must include a repository")
+        image_registry = f"{host}#{repository}"
+    return f"{image_registry}/sweb.eval.arm64.{instance_id.lower()}:v4.1.0-arm64"
```
arekay-nv, line 161: This seems to be hardcoded to `arm64` - is getting `x86` too much extra work - would prefer if we can add that.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+class PyxisEnvironmentConfig(BaseModel):
+    image: str | Path
+    run_id: str
+    cwd: str = "/testbed"
+    env: dict[str, str] = Field(default_factory=dict)
+    timeout: int = 30
```
arekay-nv: `timeout_s` would be preferable - makes reasoning about timeouts easier

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
```diff
@@ -0,0 +1,304 @@
...
+            run_srun_step(
+                image=self.config.image,
+                name=self.name,
+                mounts=[(self._tmp_dir, "/tmp")],
+                workdir=self.config.cwd,
+                argv=["true"],
+                status_path=self._tmp_dir / Path(_STEP_STATUS).name,
+                timeout_s=self.config.timeout,
+                failure_path=self.config.infrastructure_failure_path,
+            )
```
arekay-nv, line 199: Please add a comment that this is the init step - `true` (if i understand correctly) is a no-op command to pass in.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
```diff
@@ -0,0 +1,304 @@
...
+                f"failed to start Pyxis container for {self.config.image}"
+            ) from exc
+
+    def execute(
+        self, action: dict[str, Any], cwd: str = "", *, timeout: int | None = None
+    ) -> dict[str, Any]:
+        command = action.get("command", "")
+        argv = ["env"]
+        argv.extend(f"{key}={value}" for key, value in self.config.env.items())
+        argv.extend([*self.config.interpreter, command])
```
arekay-nv, line 213: useful to log the command being executed here to debug.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_worker.py
```diff
@@ -0,0 +1,268 @@
...
+timeout "$timeout_s" /bin/bash "$eval_path" >"$output_path" 2>&1
+status=$?
+set -e
+cat "$output_path"
+if [[ $status -eq 124 ]]; then
+    echo "Timeout error: $timeout_s seconds exceeded." >>"$output_path"
+    exit 124
+fi
+exit 0
+"""
```
arekay-nv, line 53: Can you please add a comment explaining that this is replicating the eval script for the docker path, and document differences if any.

arekay-nv (commented):
Thanks for this - please add some documentation especially to the pyxis path for users who might not be too familiar with pyxis. Also some choices appear arbitrary but are derived from compatibility with the docker path - please document those as well.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
hvagadia: Good idea, let me handle that in a separate PR. I need to test it so there is some non trivial work.

PR #441, opened 2026-08-07
Count structured chat tokens accurately

hvagadia:
## Summary

Adds model-aware token counting for structured chat inputs and outputs while keeping the existing pre-tokenized and plain-text paths explicit.

The metrics pipeline now carries one of four tokenization inputs:

- `TokenIdsInput`: already-tokenized data; count with `len(token_ids)`.
- `TextInput`: plain text; use the fast Rust backend when available, otherwise the tokenizer wrapper.
- `MessageInput`: structured assistant output (`content`, `reasoning`, and `tool_calls`); render with `apply_chat_template()` for OSL/TPOT.
- `PromptInput`: complete chat input (`messages`, `tools`, and chat-template options); render with `apply_chat_template()` for ISL.

Tokenizer input routing

## Design

- `PromptData` preserves structured messages, tools, optional `chat_template_kwargs`, and an optional dataset-provided chat template.
- `IslTrigger`, `OslTrigger`, and `TpotTrigger` convert events into the appropriate typed tokenization input.
- `BatchTokenizer.count_batch_async()` performs a single explicit type dispatch across the four paths.
- Plain-text batches use process-sharded fast-tokenizer backends when available. Tokenizers without a fast backend fall back to the public tokenizer wrapper.
- Structured messages and prompts always use the full tokenizer wrapper because chat-template rendering is required.
- Tool-call arguments are normalized for templates that expect dictionaries rather than OpenAI-wire JSON strings.
- List-form multimodal prompts continue to be issued normally; ISL is unavailable for that representation and no per-sample warning is emitted.

## Validation

Server and client token counts were compared on the benchmark workloads. Differences are expected where the server reports generated engine tokens while the client reconstructs tokens from the structured response.

| Model | Requests | Server ISL | Client ISL | ISL difference | Server OSL | Client OSL | OSL difference |
|---|---:|---:|---:|---:|---:|---:|---:|
| Kimi K3 | 457 | 20,286,232 | 20,144,472 | -141,760 (-0.699%) | 183,171 | 177,257 | -5,914 (-3.229%) |
| Qwen 3.6 35B-A3B | 547 | 23,324,820 | 23,095,360 | -229,460 (-0.984%) | 200,694 | 199,051 | -1,643 (-0.819%) |
| DeepSeek-R1 | 500 | 399,347 | 399,347 | 0 (0.000%) | 2,066,482 | 2,065,511 | -971 (-0.047%) |
| GPT-OSS 120B | 500 | 2,542,452 | 2,542,452 | 0 (0.000%) | 502,063 | 492,575 | -9,488 (-1.890%) |

Qwen structured-tokenization drain stress test:

| Trajectory limit | Target concurrency | Runtime | Samples issued | Samples completed | Failed | Pending at drain | Drain time | Pending after drain |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 613 (full dataset, no repetition) | 128 | 1 hour | 5,531 | 5,403 | 0 | 408 | 15.299 s | 0 |

The run reached its one-hour limit. Tokenization remained independent of request issuing; the accumulated structured-tokenization backlog added 15.299 seconds to final drain. One structured output used the existing whitespace fallback because its tool-call arguments were not a mapping.

Additional checks:

- Focused list-form prompt test passes.
- Scoped pre-commit hooks pass, including Ruff formatting and mypy.
- Full PR test, build, audit, BFCL smoke, and schema checks passed on the previous revision; the latest checks are running after the formatting and review-feedback update.


src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -319,6 +319,13 @@ def fire(self, ev_rec, row, pre_change):
         if isinstance(ev_rec.data, PromptData) and ev_rec.data.token_ids is not None:
             self.registry.record(self.metric_name, len(ev_rec.data.token_ids))
             return
+        if isinstance(ev_rec.data, PromptData) and ev_rec.data.messages is not None:
+            if self._queue is not None:
+                self._queue.enqueue_prompt(
+                    (ev_rec.data.messages, ev_rec.data.tools),
+                    self._make_recorder(ev_rec, pre_change),
+                )
+            return
```
arekay-nv: Can this be combined so the common `if isinstance(ev_rec.data, PromptData)` is evaluated one making it more legible. 
The `messages` and `self._queue is not None` and then be combined.

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -349,7 +356,9 @@ def _extract_text(self, ev_rec, row, pre_change):
         return None
 
     def _extract_message(self, ev_rec, row, pre_change):
-        if isinstance(ev_rec.data, TextModelOutput) and ev_rec.data.tool_calls:
+        if isinstance(ev_rec.data, TextModelOutput) and (
+            ev_rec.data.reasoning or ev_rec.data.tool_calls
+        ):
```
arekay-nv: Did this change the TPS we are observing as we didn't account for the reasoning output before?

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -111,6 +113,37 @@ def _normalize_tool_calls_for_template(
...
+
+    def encode(self, text: str, *, add_special_tokens: bool = False) -> list[int]:
+        return self.tokenizer.encode(text, add_special_tokens=add_special_tokens)
+
+    def encode_lengths(self, texts: list[str]) -> list[int]:
+        return [len(self.encode(text, add_special_tokens=False)) for text in texts]
+
+
+def _backend_from_tokenizer(tokenizer: Any) -> Any | None:
+    """Return the tokenizer's supported length-counting path."""
```
arekay-nv: Is this docstring accurate?

src/inference_endpoint/load_generator/session.py
```diff
@@ -263,19 +263,23 @@ def issue(
...
+                tools = data.get("tools")
+                prompt_data = PromptData(
+                    messages=tuple(data["messages"]),
+                    tools=tuple(tools) if isinstance(tools, list | tuple) else None,
+                )
+            else:
+                prompt_text = data.get("prompt")
+                prompt_data = PromptData(
+                    text=prompt_text if isinstance(prompt_text, str) else None
+                )
```
arekay-nv: Recommend hardening this to `elif isinstance(data.get("prompt"), str)` and adding a warning in the else clause to catch multimodal inputs or any other cases - otherwise these may get silently ignored.

src/inference_endpoint/load_generator/session.py
```diff
@@ -263,19 +263,23 @@ def issue(
...
-                text=prompt_text if isinstance(prompt_text, str) else None,
-                token_ids=tuple(token_ids) if token_ids is not None else None,
-            )
+            # Prefer the exact representation sent to the endpoint: existing
+            # token IDs, then structured chat messages, then a plain prompt.
+            # Non-string standalone prompts (for example multimodal content
+            # parts) remain unavailable for ISL reporting.
+            if token_ids is not None:
+                prompt_data = PromptData(token_ids=tuple(token_ids))
+            elif isinstance(data.get("messages"), list | tuple):
```
nvzhihanj: I see there's a priority order of message prep here - Should message, prompt, token_id be mutually exclusive? E.g. throwing warning/error if more than one is present in the parquet?

There were frequent confusions about which field is used in the parquet in the loadgen - and would be good to call this priority / mutual exclusiveness out somewhere

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -319,6 +319,13 @@ def fire(self, ev_rec, row, pre_change):
         if isinstance(ev_rec.data, PromptData) and ev_rec.data.token_ids is not None:
             self.registry.record(self.metric_name, len(ev_rec.data.token_ids))
             return
+        if isinstance(ev_rec.data, PromptData) and ev_rec.data.messages is not None:
```
nvzhihanj: Seems missing comment (which inherits from the prepopulated token ids comment above)

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -111,6 +113,37 @@ def _normalize_tool_calls_for_template(
...
+    """Use a custom tokenizer's public API instead of bypassing its Python logic."""
+
+    def __init__(self, tokenizer: Any) -> None:
+        self.tokenizer = tokenizer
+
+    def encode(self, text: str, *, add_special_tokens: bool = False) -> list[int]:
+        return self.tokenizer.encode(text, add_special_tokens=add_special_tokens)
+
+    def encode_lengths(self, texts: list[str]) -> list[int]:
+        return [len(self.encode(text, add_special_tokens=False)) for text in texts]
```
nvzhihanj: Should we use batch tokenizer for this path?

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -260,25 +296,26 @@ def __init__(
...
-                str,
+                list[int],
                 tok.apply_chat_template(
-                    [_PREFIX_USER_MSG], tokenize=False, add_generation_prompt=False
+                    [_PREFIX_USER_MSG], tokenize=True, add_generation_prompt=False
                 ),
             )
-            self._prefix_len = len(tok.tokenize(prefix))
+            self._prefix_len = len(prefix)
             with_assistant = cast(
```
arekay-nv: rename to with_assistant_tokens to help disambiguate text data from tokens.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -260,25 +296,26 @@ def __init__(
...
         # prefix alone, so the assistant frame is subtracted from message counts.
         try:
             prefix = cast(
-                str,
+                list[int],
                 tok.apply_chat_template(
-                    [_PREFIX_USER_MSG], tokenize=False, add_generation_prompt=False
+                    [_PREFIX_USER_MSG], tokenize=True, add_generation_prompt=False
                 ),
             )
```
arekay-nv, line 298: `apply_chat_template` [returns a dictionary](https://github.com/huggingface/transformers/blob/f93eb6dbec8c2f5563f6d69bd998771929e8263f/src/transformers/tokenization_utils_base.py#L3002) by default if tokenize is set to True. This is only trigerred when tokenize is True. Please update call sites. Also, can you add tests that would catch this behavior - fail with current change and pass with `return_dict` set to False for `tokenize` set to `True`.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -462,6 +496,36 @@ async def token_count_message_async(
...
+        kwargs: dict[str, Any] = {
+            "tokenize": True,
+            "add_generation_prompt": True,
+        }
+        if tools is not None:
+            kwargs["tools"] = list(tools)
+        encoded = self._tokenizer.apply_chat_template(  # type: ignore[union-attr]
+            list(messages), **kwargs
+        )
+        return len(encoded)
```
arekay-nv: **[critical] `len(apply_chat_template(..., tokenize=True))` counts `BatchEncoding` keys, not tokens.**

On the pinned `transformers==5.5.0`, `apply_chat_template(..., tokenize=True)` defaults to `return_dict=True` and returns a `BatchEncoding`, so `len(encoded)` is the key count (~2), not the token count. Live repro on `Qwen/Qwen2.5-0.5B-Instruct`: a 44‑token prompt → ISL `2`. The same pattern corrupts the baseline (`_prefix_len`/`_baseline` become `2`/`0`) and `_token_count_message` (line 465), so structured OSL/TPOT collapse to `0` for every reasoning/tool-call sample.

Fix: pass `return_dict=False` (or index `["input_ids"]`) on every `tokenize=True` site — here plus lines 309, 318, 465.

Suggested test (fails now, passes after the fix):
```python
def test_structured_prompt_isl_is_token_count_not_batchencoding_keys():
    bt = BatchTokenizer("Qwen/Qwen2.5-0.5B-Instruct",
                        live_workers=1, cores_per_worker=0, n_workers=0)
    try:
        msgs = ({"role": "user", "content": "Write a detailed essay about the fall of Rome."},)
        assert bt._token_count_prompt(msgs, None) > 10   # returns 2 today
    finally:
        bt.close()
```

tests/unit/async_utils/services/metrics_aggregator/test_token_metrics.py
```diff
@@ -139,7 +139,7 @@ class _FakeTokenizerWithTemplate(_FakeTokenizer):
     """Tokenizer that supports apply_chat_template for tool-call testing."""
 
     def apply_chat_template(
-        self, messages, tokenize=False, add_generation_prompt=False
+        self, messages, tools=None, tokenize=False, add_generation_prompt=False
```
arekay-nv: **[high] This fake hides the `BatchEncoding` defect, so the suite stays green while production is wrong.**

`apply_chat_template(..., tokenize=True)` here returns a bare `list`, so `len()` equals the token count — but real `transformers` returns a `BatchEncoding` whose `len()` is the key count (~2). That is why the new chat-template tests pass while the real path returns `2`.

Make the fake return a mapping for `tokenize=True` unless `return_dict=False`, so the suite fails until the call sites are fixed:
```python
class _BE(dict):  # BatchEncoding stand-in: len() == number of keys
    pass

def apply_chat_template(self, messages, tools=None, tokenize=False,
                        add_generation_prompt=False, return_dict=True):
    ids = list(range(len(...)))            # existing token logic
    if tokenize and return_dict:
        be = _BE(); be["input_ids"] = ids; be["attention_mask"] = [1]*len(ids)
        return be
    return ids if tokenize else rendered
```

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -349,7 +356,9 @@ def _extract_text(self, ev_rec, row, pre_change):
         return None
 
     def _extract_message(self, ev_rec, row, pre_change):
-        if isinstance(ev_rec.data, TextModelOutput) and ev_rec.data.tool_calls:
+        if isinstance(ev_rec.data, TextModelOutput) and (
+            ev_rec.data.reasoning or ev_rec.data.tool_calls
```
arekay-nv: **[high] Reasoning is silently dropped from OSL/TPOT (independent of the `return_dict` fix).**

Routing `reasoning`-only outputs through the input chat template renders `{"role": "assistant", "reasoning_content": ...}`, which reference templates (Qwen, DeepSeek-R1, …) ignore — and it does not raise, so the whitespace fallback never fires. Verified on Qwen2.5: `_token_count_message("hi", "<long reasoning>", None) == _token_count_message("hi", None, None)`. Previously the plain-text path counted reasoning via `str(TextModelOutput)`, so this under-reports OSL/TPOT (and inflates TPS) for exactly the reasoning models this change targets. Same applies to `TpotTrigger` (line 404).

Consider counting reasoning as generated text rather than through the input-side template.

Suggested test:
```python
def test_reasoning_tokens_are_counted():
    bt = BatchTokenizer("Qwen/Qwen2.5-0.5B-Instruct",
                        live_workers=1, cores_per_worker=0, n_workers=0)
    try:
        assert bt._token_count_message("hi", "a long private chain of reasoning", None) \
             > bt._token_count_message("hi", None, None)
    finally:
        bt.close()
```

src/inference_endpoint/load_generator/session.py
arekay-nv: **[medium] Empty and multimodal `messages` are silently dropped from ISL.**

`isinstance([], list | tuple)` is `True`, so a dataset `messages: []` builds `PromptData(messages=())`; `_token_count_prompt` then calls `apply_chat_template([], add_generation_prompt=True)`, which raises `IndexError` (verified) — and with no fallback that drops the sample's ISL and taints the drain. This branch is also a regression for multimodal chat: the previous code extracted text via `_extract_prompt_text(...)` (now dead code), whereas list-shaped `content` sent here makes text-only templates error.

Guard with truthiness so empty falls through to the text branch, and sanitize/fall back for non-text `content`:
```python
elif data.get("messages"):
    ...
```

Suggested test:
```python
def test_empty_messages_fall_through_not_indexerror():
    from transformers import AutoTokenizer
    tok = AutoTokenizer.from_pretrained("Qwen/Qwen2.5-0.5B-Instruct")
    with pytest.raises(IndexError):   # documents current behavior; guard removes it
        tok.apply_chat_template([], tokenize=True, add_generation_prompt=True, return_dict=False)
```

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -462,6 +496,36 @@ async def token_count_message_async(
             self._thread, self._token_count_message, content, reasoning, tool_calls
         )
 
+    def _token_count_prompt(
```
arekay-nv, line 495: **[medium] `_token_count_prompt` lacks the safeguards `_token_count_message` already has.**

Three gaps versus the message path:
1. **No `_normalize_tool_calls_for_template`** on assistant `tool_calls` embedded in prompt history. Agentic multi-turn history carries wire-JSON `arguments` **strings**; Hermes/Qwen templates that iterate `arguments` as a mapping then diverge or raise (verified 50 vs 49 tokens on Qwen2.5), while OSL handles them.
2. **No try/except fallback** — a template failure propagates and taints the drain instead of degrading to an approximate count.
3. **`chat_template_kwargs` ignored** — the adapter forwards these to the server (e.g. Qwen3 `enable_thinking`: 19 vs 15 tokens), but they are not threaded through `PromptData`, so ISL diverges from the served prompt.

Mirror the message path: normalize embedded tool calls, wrap in the same try/fallback, and carry `chat_template_kwargs` through `PromptData`.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -111,6 +113,37 @@ def _normalize_tool_calls_for_template(
...
+def _backend_from_tokenizer(tokenizer: Any) -> Any | None:
+    """Return the tokenizer's supported length-counting path."""
+    backend = getattr(tokenizer, "backend_tokenizer", None)
+    if backend is not None:
+        return backend
+
+    model = getattr(tokenizer, "model", None)
+    model_type = type(model)
+    if (
+        model_type.__module__ == "tiktoken.core"
```
arekay-nv: **[medium] The custom-tiktoken backend is unreachable — `tiktoken` is not a declared dependency.**

For a Kimi-style tokenizer, `AutoTokenizer.from_pretrained(...)` imports `tiktoken` during load and raises `ImportError` before `_backend_from_tokenizer` is ever reached (`tiktoken` is not in `pyproject.toml`; not importable in the current env). As-is the advertised backend can never activate in a clean install. Declare `tiktoken` in the appropriate dependency group (pinned `==`) or document the requirement, and add a test that exercises a real tiktoken-backed tokenizer rather than a fake.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -676,13 +762,29 @@ async def _flush(self, live: bool) -> None:
...
                     raise
                 except Exception as exc:  # noqa: BLE001 — isolate items.
                     failure = failure or exc
                     if live:
                         self._msg.append(((content, reasoning, tool_calls), on_count))
                     continue
                 self._record(on_count, count)
+            for i, ((messages, tools), on_count) in enumerate(prompt_items):
+                try:
+                    count = await self._tokenizer.token_count_prompt_async(
```
arekay-nv: **[medium] The prompt lane never shards and drains serially under the flush lock.**

`token_count_prompt_async` runs on the small in-process thread pool (like the message path), and `_flush` awaits each prompt item one-by-one while holding `self._lock`. On `main` this in-process lane was justified as "rare (tool calls)"; this change makes it the path for **every** structured-chat ISL sample. At high QPS the end-of-run drain can exceed `--drain-timeout`, leaving ISL permanently uncounted (`n_pending_tasks > 0`) and blocking concurrent live flushes. Consider batching prompt render+encode through the sharded path like `count_texts_async`.

Suggested test (peak-concurrency probe with a stub tokenizer):
```python
async def test_prompt_drain_is_not_serialized_per_item():
    peak = active = 0
    class T:
        async def count_texts_async(self, *a, **k): return []
        async def token_count_message_async(self, *a, **k): return 0
        async def token_count_prompt_async(self, m, t, loop):
            nonlocal active, peak
            active += 1; peak = max(peak, active)
            await asyncio.sleep(0.01); active -= 1; return 3
    q = TokenBatchQueue(T(), asyncio.get_running_loop())
    for _ in range(8):
        q.enqueue_prompt((({"role":"user","content":"x"},), None), lambda c: None)
    await q.drain_all()
    assert peak > 1
```

arekay-nv (commented):
## Consolidated review

This change targets accurate structured-chat token counting, but as written the core counts are incorrect on the pinned `transformers==5.5.0`, and two further issues persist even after the primary fix. All findings below are verified against the pinned tokenizer (live repro on `Qwen/Qwen2.5-0.5B-Instruct` / `Qwen3-8B`).

**Must fix**
- **`tokenize=True` returns a `BatchEncoding`; `len()` counts keys, not tokens** (`token_metrics.py` 309/318/465/513). ISL → ~2, structured OSL/TPOT → 0. Fix: `return_dict=False` on every `tokenize=True` site.
- **A unit-test fake masks the above** — it returns a bare list for `tokenize=True`, so the suite stays green (`test_token_metrics.py`).

**Persists after the fix above**
- **Reasoning silently dropped** from OSL/TPOT — input templates ignore `reasoning_content` (`metrics_table.py` 360/404).
- **Empty/multimodal `messages` dropped** from ISL — `messages: []` → `IndexError`; multimodal `content` errors; `_extract_prompt_text` is now dead (`session.py` 272).

**Should fix**
- `_token_count_prompt` lacks tool-call normalization, a try/fallback, and `chat_template_kwargs` parity that `_token_count_message`/the adapter already have (`token_metrics.py` 499).
- Custom-tiktoken backend unreachable — `tiktoken` undeclared (`token_metrics.py` 138).
- Prompt lane drains serially under the flush lock, never sharded — high-QPS runs can exceed `--drain-timeout` (`token_metrics.py` 775).

**Checked and fine:** msgspec wire round-trip + `gc=False` safety, `_inflight` accounting across the three flush phases, and `add_generation_prompt=True` for ISL.

Suggested tests are inline on the relevant lines; each is written to fail on the current code and pass after the described fix. Real-tokenizer tests use the cached `Qwen/Qwen2.5-0.5B-Instruct` and can be `pytest.skip`-guarded when uncached.

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
nv-alicheng: [Claude] **medium** (testing): The new ISL structured-prompt lane has no queue-integration test. Grep over `tests/` finds zero references to `IslTrigger` or `enqueue_prompt`. Untested: (a) `IslTrigger.fire` messages routing here (322-328), including the silent drop when `self._queue is None`; (b) `TokenBatchQueue.enqueue_prompt` and the whole prompt lane in `_flush` (token_metrics.py:773-787) — the `_record`/`_inflight` decrement, drain-vs-live requeue, and the CancelledError handler that requeues *all* `prompt_items` (token_metrics.py:765). The only prompt coverage is `test_token_count_prompt_preserves_messages_tools_and_generation_prompt`, which calls `token_count_prompt_async` directly and never touches the queue or trigger; the metrics_table message-routing test covers OSL/TPOT only, not ISL. This is a hot-path lane with nontrivial cancellation/requeue logic — add direct queue + trigger tests.

src/inference_endpoint/core/types.py
```diff
@@ -284,17 +284,22 @@ class PromptData(
...
     Attributes:
         text: Raw prompt string. Set when the adapter sends text prompts.
         token_ids: Pre-computed token IDs. Set when the adapter pre-tokenizes (e.g. SGLang).
+        messages: Structured messages sent to a chat-completions endpoint.
+        tools: Tool declarations accompanying ``messages``.
     """
 
     text: str | None = None
     token_ids: tuple[int, ...] | None = None
+    messages: tuple[dict[str, Any], ...] | None = None
```
nv-alicheng, line 308: [Both] **low** (code-quality): `PromptData` is `gc=False` (types.py:283) but carries two mutable-container fields, `messages: tuple[dict[str, Any], ...]` and `tools: tuple[dict[str, Any], ...]`, holding arbitrary nested chat/tool dicts. Every sibling struct in this file documents the gc-audit convention: `TextModelOutput` carries an explicit `AT-RISK (gc=False): Has mutable container field 'tool_calls'...` note, `StreamChunk`/`ErrorData` justify gc=False with "metadata contains only scalar key-value pairs." PromptData's docstring has no such note even though its dicts are the least scalar of any. The choice is practically safe (JSON-native dataset dicts don't cycle back; encoded synchronously and short-lived) — so this is a convention/documentation gap, not a runtime defect — but it breaks the file's own documented gc-safety process. Add the AT-RISK annotation. Separately, `dict[str, Any]` for `messages`/`tools` is loosely typed where a `TypedDict` (`role`/`content`/optional `reasoning_content`/`tool_calls`) would pin the keys the template code reaches for.

nv-alicheng (commented):

Reviewed by: **Claude** + **Code-Quality** | Depth: **standard**

`codex` unavailable in this environment. This PR already has a thorough human review; this pass **independently corroborates** the existing critical/high findings and adds only net-new items (see summary).

nv-alicheng:

Reviewed by: **Claude** + **Code-Quality** | Depth: **standard**
(`codex` CLI unavailable in this environment.)

This PR already carries a thorough human review (arekay-nv, nvzhihanj). This council pass **independently corroborates** those findings and adds net-new ones — it does not re-file what's already commented.

### ✅ Corroborated (already commented — not re-filed)
- **token_metrics.py:513 (critical):** confirmed. `apply_chat_template(..., tokenize=True)` is called at **305, 312, 462, 510 with no `return_dict=False` anywhere** in the file, so on `transformers==5.5.0` `len(encoded)` counts `BatchEncoding` keys (~2), not tokens → structured ISL/OSL/TPOT collapse. Real.
- **test_token_metrics.py:142 (high)** and **metrics_table.py:360 (high, reasoning dropped)** — consistent with our independent read.

### 🟡 Net-new — Should Fix (medium)

| 1 | `session.py` | 53 | dead-code | Both | `_extract_prompt_text` has **zero production callers** after this PR rerouted the `messages` branch to `PromptData(messages=...)` — grep confirms only its own def remains; it stays "green" solely because `test_async_session.py` still exercises it directly, masking the deadness. Delete it (and its orphaned tests), or wire it in as the text fallback for chat-template render failures. _(def is at an untouched line — reported here, not inline)_ |
| 2 | `metrics_table.py` | 322 | testing | Claude | New ISL structured-prompt lane has **no queue-integration test** (zero refs to `IslTrigger`/`enqueue_prompt` in `tests/`): the `_queue is None` silent-drop, `enqueue_prompt`, and the prompt-lane `_flush` requeue/CancelledError path (token_metrics.py:765,773-787) are all uncovered. Hot-path cancellation/requeue logic — add direct tests. |

### 🔵 Net-new — Consider (low)

| 3 | `core/types.py` | 301 | code-quality | Both | `PromptData` is `gc=False` with mutable `messages`/`tools` dict-tuples but — unlike every sibling struct in the file — carries **no `AT-RISK` gc-audit note**, breaking the file's documented gc-safety convention (safe in practice; add the note). Also `dict[str, Any]` → a `TypedDict` would pin message keys. |
| 4 | `token_metrics.py` | 299 | code-quality | Quality | `self._backend` is set inside `_load_tokenizer` but never declared/annotated in `__init__`, unlike its sibling `self._tokenizer: PreTrainedTokenizerBase \| None = None`. Declare it for a pinned type. |
| 5 | `token_metrics.py` | 207 | code-quality | Quality | `[len(getattr(item, "ids", item)) for item in encoded]` conflates two return shapes (`Encoding.ids` vs raw list) via a self-referential `getattr` default — branch on the known backend type instead. |

**Dropped as duplicates of existing threads:** IslTrigger double-`isinstance`/combine (already arekay-nv:328), tiktoken magic-string coupling (same line as the existing tiktoken-not-declared thread at :138).

> Commit hygiene fine: 2 commits, no fixups.

src/inference_endpoint/load_generator/session.py
```diff
@@ -262,21 +236,58 @@ def issue(
...
+                )
+
+            if input_tokens is not None or token_ids is not None:
+                selected_token_ids = (
+                    input_tokens if input_tokens is not None else token_ids
+                )
+                prompt_data = PromptData(token_ids=tuple(selected_token_ids))
+            elif isinstance(messages, list | tuple) and messages:
+                tools = data.get("tools")
+                chat_template_kwargs = data.get("chat_template_kwargs")
```
viraatc, line 261: **high**: `chat_template` is forwarded to the endpoint but not to token metrics. It is an allowed per-sample dataset column (`openai_msgspec_adapter.py:105`) and is sent on the request (`openai_msgspec_adapter.py:189`), but `PromptData` only carries `chat_template_kwargs`. When a dataset overrides the template, the server renders with the custom template while ISL (and the message-path baseline) is rendered with the tokenizer's default template — reported ISL/OSL/TPOT silently diverge from the served tokens, which is the exact failure mode this PR is meant to eliminate. Suggest carrying `chat_template` through `PromptData` alongside `chat_template_kwargs` and passing it into `apply_chat_template` (it accepts a `chat_template=` kwarg).

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -295,20 +342,20 @@ def _setup_shards(self, cores_per_worker: int, n_workers: int) -> None:
...
         """
         if cores_per_worker <= 0 or n_workers == 0:
             logger.info("BatchTokenizer: in-process tokenization (explicit)")
             return
-        if getattr(self._tokenizer, "backend_tokenizer", None) is None:
-            raise RuntimeError(
-                f"tokenizer {self._tokenizer_name!r} has no fast (Rust) "
-                "backend; token metrics require one to keep up with "
-                "completions. Use a fast tokenizer, or disable token metrics."
+        if self._text_backend is None:
```
viraatc, line 333: **high**: Tokenizers without a fast backend now start ("structured chat tokenization remains available"), but nothing prevents plain-text items from being enqueued — and the text lane now hard-fails for them. `OslTrigger`/`TpotTrigger` take the message path only when `reasoning or tool_calls` is set; a content-only response goes through `enqueue_text` → `count_texts_async` → `_encode_lengths_inproc`, which raises `RuntimeError` when `self._text_backend is None` (main fell back to `[len(tok.tokenize(t)) for t in texts]`). The same applies to plain string-`prompt` datasets. Net effect: a run on a wrapper tokenizer whose responses lack reasoning produces OSL/TPOT items that stay pending forever and surface as an incomplete drain (`n_pending_tasks > 0`) only at end of run — the whole run is spent before the failure is visible. Either fail fast at startup when the run can produce text-lane items, or give the no-backend case a working text path (e.g. the wrapper's `encode` on the in-process thread lane).

src/inference_endpoint/load_generator/session.py
```diff
@@ -262,21 +236,58 @@ def issue(
...
+                name
+                for name, present in (
+                    ("input_tokens", input_tokens is not None),
+                    ("token_ids", token_ids is not None),
+                    ("messages", isinstance(messages, list | tuple) and bool(messages)),
+                    ("prompt", isinstance(prompt, str)),
+                )
+                if present
+            ]
+            if len(representations) > 1:
```
viraatc: **high**: The multiple-representation check raises mid-run, per sample, on the issuing hot path. Three concerns: (a) dataset schemas are homogeneous, so this validation belongs in phase setup — as written it aborts the run on the first issued sample after setup time has already been spent; (b) it is a behavior change for datasets carrying both `prompt` and `messages` through the default OpenAI adapter: the adapter's own `ColumnFilter` requires `prompt` while allowing `messages`, and `to_endpoint_request` prefers `messages` when present, so that combination worked on main with a clear precedence and now hard-fails; (c) the `representations` list is built on every `issue()` call — including every agentic turn and every re-issue under with-replacement ordering — per-sample allocation for a validation whose result never changes. Suggest validating once per phase at setup and keeping documented precedence (or a warning) for the `prompt`+`messages` combination the default adapter tolerates.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -462,6 +513,61 @@ async def token_count_message_async(
...
+        )
+        if tools is not None:
+            kwargs["tools"] = list(tools)
+        prompt_messages = _normalize_prompt_messages_for_template(messages)
+        try:
+            encoded = self._tokenizer.apply_chat_template(  # type: ignore[union-attr]
+                prompt_messages, **kwargs
+            )
+        except Exception:
+            text_only_messages, changed = _text_only_multimodal_messages(messages)
```
viraatc: **medium**: The multimodal fallback re-renders from the raw `messages`, discarding the tool-call normalization applied to the first attempt. The first attempt uses `prompt_messages = _normalize_prompt_messages_for_template(messages)` (converts wire-JSON `function.arguments` strings to dicts, since mapping-iterating templates raise on strings); the `except` branch calls `_text_only_multimodal_messages(messages)` on the un-normalized input. A multimodal prompt whose history also contains tool calls fails the retry for the tool-call reason, and the sample's ISL is dropped. Fix is one word: `_text_only_multimodal_messages(prompt_messages)`.

src/inference_endpoint/load_generator/session.py
```diff
@@ -262,21 +236,58 @@ def issue(
...
+
+            if input_tokens is not None or token_ids is not None:
+                selected_token_ids = (
+                    input_tokens if input_tokens is not None else token_ids
+                )
+                prompt_data = PromptData(token_ids=tuple(selected_token_ids))
+            elif isinstance(messages, list | tuple) and messages:
+                tools = data.get("tools")
+                chat_template_kwargs = data.get("chat_template_kwargs")
+                prompt_data = PromptData(
```
viraatc, line 262: **medium**: ISSUED events now carry the full structured message history plus all tool definitions on every issued sample. On main the payload was a flattened content string; here a large tool set (62 definitions / ~35k input tokens per turn in this PR's own validation) rides every ISSUED event through msgspec encode → ZMQ pub, is decoded by the aggregator, and is written verbatim to `events.jsonl` — and in agentic runs the growing history makes per-conversation event bytes O(turns²). The data is needed for accurate ISL — but the impact of this event-size growth should be measured at high QPS before this path is relied on there.

src/inference_endpoint/load_generator/session.py
```diff
@@ -262,21 +236,58 @@ def issue(
...
+                    chat_template_kwargs=(
+                        dict(chat_template_kwargs)
+                        if isinstance(chat_template_kwargs, dict)
+                        else None
+                    ),
+                )
+            elif isinstance(prompt, str):
+                prompt_data = PromptData(text=prompt)
+            else:
+                logger.warning(
```
viraatc: **medium**: This warning fires once per issued sample for a supported input shape. List-of-content-parts prompts are valid multimodal inputs (the in-repo Shopify preset emits them, and the adapter docstring documents list-form prompts), so a multimodal run now formats and writes a WARNING inside the busy-wait issuing path for every sample, at target QPS — plus the matching log-file growth. Same for the non-dict branch below. Suggest logging once per phase per reason (a seen-flag), and treating list-form prompts as expected rather than warned.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -154,32 +200,30 @@ def _init_worker(tokenizer_name: str, core_set: list[int]) -> None:
...
+        _WORKER_TEXT_BACKEND.encode("warmup", add_special_tokens=False)
 
 
 def encode_lengths(backend: Any, texts: list[str]) -> list[int]:
-    """Per-text token counts via the raw tokenizers backend, one rayon call."""
+    """Per-text token counts via one bounded backend batch call."""
     encode_batch = getattr(backend, "encode_batch_fast", None) or backend.encode_batch
-    return [len(e.ids) for e in encode_batch(texts, add_special_tokens=False)]
+    encoded = encode_batch(texts, add_special_tokens=False)
+    return [len(getattr(item, "ids", item)) for item in encoded]
```
viraatc: **nit**: `getattr(item, "ids", item)` exists to handle an `encode_batch` that returns plain token-id lists. No backend that reaches this function does that: every caller passes a Rust `tokenizers` backend, and its `encode_batch`/`encode_batch_fast` always return `Encoding` objects that have `.ids`. So the fallback never fires — it only adds a `getattr` call per item in the single busiest loop of the drain path. `item.ids` directly (as on main) is sufficient.

viraatc (commented):
## Review Council — multi-AI code review

Reviewed at commit `1ee7b47`; focus: integration, hot-path performance, overheads, code quality, consistency with codebase conventions. Existing review comments were checked first — nothing below duplicates them. (Noted in passing: the earlier `return_dict`/BatchEncoding issue is addressed at HEAD, and `tiktoken` is now declared and locked.)

| # | File | Line | Severity | Summary |
| 1 | `load_generator/session.py` | 266 | high | `chat_template` reaches the server but not token metrics → silent ISL/OSL divergence |
| 2 | `.../token_metrics.py` | 352 | high | No-fast-backend tokenizers start, but content-only OSL/TPOT still routes to the text lane and fails at drain |
| 3 | `load_generator/session.py` | 253 | high | Multi-representation check: per-sample raise mid-run, on the issuing hot path; breaks prompt+messages datasets the default adapter tolerates |
| 4 | `.../token_metrics.py` | 536 | medium | Multimodal fallback drops tool-call normalization → retry re-raises, ISL lost |
| 5 | `load_generator/session.py` | 267 | medium | ISSUED events carry full messages+tools per sample: encode/ZMQ/events.jsonl amplification, O(turns²) in agentic runs — measure at high QPS |
| 6 | `load_generator/session.py` | 279 | medium | Per-sample WARNING on the issuing path for supported multimodal prompts |
| 7 | `.../token_metrics.py` | 213 | nit | Unreachable `getattr` fallback in the hottest tokenize loop |

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
hvagadia: Removed this part. In general we are using batch tokenizer only for plain-text prompts. All structured prompts will go through apply chat template.

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
hvagadia: I dont think reasoning was going in a separate channel earlier, so unlikely that was the gap.

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
hvagadia: Qwen2.5 and DSR1 dont support tokenizing reasoning via chat template path. This limitation does not exist in newer models, and as we discussed we will not support this for DSR1.

src/inference_endpoint/core/types.py
hvagadia: Added the `AT-RISK (gc=False)` documentation for `messages`, `tools`, and `chat_template_kwargs`. I am intentionally keeping these fields as `dict[str, Any]` rather than adding a `TypedDict`: the payload is passed through to model-specific chat templates, whose accepted keys vary by model and serving framework (for example `reasoning`, `reasoning_content`, `thinking`, and framework-specific tool-call fields). A shared strong schema here would either reject valid template-native fields or become an inaccurate union of every supported framework. The OpenAI adapter retains its stronger wire-level `ChatMessage` type; `PromptData` preserves the original template input for metrics tokenization.

src/inference_endpoint/load_generator/session.py
hvagadia: Good catch. I have added `chat_template` propagation for the structured ISL path: dataset sample -> `PromptData` -> `PromptInput` -> `apply_chat_template(chat_template=...)`, with focused propagation and transport tests. This directly matches how the server renders the input request. Custom-template OSL/TPOT reconstruction is not supported yet. The server applies the chat template to the input, not to generated output tokens, and our structured OSL/TPOT path reconstructs parsed output fields without retaining the per-request template context. Supporting that correctly would require carrying request context through sample completion and handling a per-template assistant baseline, so I am keeping it out of this fix. This also appears to be a rarely used override path; deployments generally use the tokenizer/model standard chat template rather than supplying a per-sample custom template.

nvzhihanj (approved):

src/inference_endpoint/async_utils/services/metrics_aggregator/tokenization.py
```diff
@@ -0,0 +1,52 @@
...
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, TypeAlias
+
+
+@dataclass(frozen=True, slots=True)
+class TokenIdsInput:
```
nv-alicheng, line 18: Is there a good reasoning these are single-field frozen dataclasses rather than type aliases? Or is it just for consistency?

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -241,21 +233,14 @@ def record(count: int) -> None:
...
-        if not text:
-            # Empty output (no text and no tool calls) is not an anomaly:
-            # there is nothing to tokenize, so we record no token-count
-            # sample rather than a spurious 0 that would skew the series.
-            return
-        self._queue.enqueue_text(text, self._make_recorder(ev_rec, pre_change))
+        if isinstance(item, TokenIdsInput):
+            self.registry.record(self.metric_name, len(item.token_ids))
+        elif isinstance(item, TextInput | MessageInput | PromptInput):
+            if self._queue is not None:
```
nv-alicheng, line 245: Is there ever an instance where `self._queue` is None? What does that state symbolize and when can it happen?

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
```diff
@@ -314,18 +299,20 @@ def __init__(
...
-        # unavailable (e.g. OpenAI-compatible endpoints). Enqueued by the base.
-        super().fire(ev_rec, row, pre_change)
-
-    def _extract_text(self, ev_rec, row, pre_change):
-        if isinstance(ev_rec.data, PromptData) and ev_rec.data.text is not None:
-            return ev_rec.data.text
+    def _extract_tokenization_input(self, ev_rec, row, pre_change):
+        if not isinstance(ev_rec.data, PromptData):
+            return None
+        if ev_rec.data.token_ids is not None:
```
nv-alicheng, line 315: Can we have some docs on the priority / hierarchy? Are these states mutually exclusive?

I know for instance, sglang sometimes contains responses that contain both text and token_ids.
Is there ever a case where token_ids is (mistakenly) empty list (not None) but text is a non-empty string?

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -447,20 +478,111 @@ def _token_count_message(
...
-        content: str,
-        reasoning: str | None,
-        tool_calls: tuple[dict[str, Any], ...] | None,
-        loop: asyncio.AbstractEventLoop,
-        /,
+        messages: tuple[dict[str, Any], ...],
+        tools: tuple[dict[str, Any], ...] | None,
+        chat_template_kwargs: dict[str, Any] | None = None,
+        chat_template: str | None = None,
     ) -> int:
```
nv-alicheng, line 502: Needs docstring.  Can you add docstrings for both this and _token_count_message? What is the difference? What are they counting?

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -447,20 +478,111 @@ def _token_count_message(
...
+        live: bool = False,
+    ) -> list[int | Exception]:
+        """Count a mixed batch while preserving input order."""
+        outcomes: list[int | Exception | None] = [None] * len(inputs)
+        text_indices: list[int] = []
+        texts: list[str] = []
+        structured: list[tuple[int, MessageInput | PromptInput]] = []
+
+        for index, item in enumerate(inputs):
+            if isinstance(item, TokenIdsInput):
```
viraatc: minor nit: long stacks of isinstance are well suited to python switch, easier to read (applies to the entire MR)

nv-alicheng: nit:
```
match item:
    case TokenIdsInput():
        outcomes[index] = len(item.token_ids)
    case TextInput():
        ...
```

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -447,20 +478,111 @@ def _token_count_message(
...
+                outcomes[index] = len(item.token_ids)
+            elif isinstance(item, TextInput):
+                text_indices.append(index)
+                texts.append(item.text)
+            elif isinstance(item, MessageInput):
+                structured.append((index, item))
+            elif isinstance(item, PromptInput):
+                structured.append((index, item))
+
+        if texts:
```
nv-alicheng: It is confusing what the states in the various variables are given this condition. Can you lift this into the `isinstance(item, TextInput)` block?

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -447,20 +478,111 @@ def _token_count_message(
...
+            try:
+                counts = await self._count_texts_async(texts, loop, live=live)
+                if len(counts) != len(texts):
+                    raise RuntimeError(
+                        f"tokenizer returned {len(counts)} counts for "
+                        f"{len(texts)} texts"
+                    )
+                for index, count in zip(text_indices, counts, strict=True):
+                    outcomes[index] = count
+            except Exception as exc:  # noqa: BLE001 - isolate this input kind.
```
nv-alicheng: nit: Can you isolate each error handling part to its relevant cause? Stealing a paradigm from Golang:

```
def _list_of_err(err, count) -> list[Exception]:
    return [err for _ in range(count)]

def _texts_to_outcomes_helper(texts, text_indices) -> <type of `outcomes`>:
    try:
        counts = await self._count_texts_async(texts, loop, live=live)
    except Exception as e:  # Not sure what errors might get thrown by self._count_texts_async
        return _list_of_err(e, len(text_indices))
        
    if len(counts) != len(text_indices):  # IMO len(text_indices) is a better check than len(texts) as it's used directly in the following list construction
        return _list_of_err(RuntimeError(
            f"tokenizer returned {len(counts)} counts for "
            f"{len(text_indices)} texts"
        ), len(text_indices))

    # optionally, assert / check if len(texts) == len(text_indices), and max(text_indices) == len(counts) - 1, min(text_indices) == 0
    return [counts[i] for i in text_indices]
```

src/inference_endpoint/core/types.py
```diff
@@ -284,17 +284,30 @@ class PromptData(
...
+        tools: Tool declarations accompanying ``messages``.
+        chat_template_kwargs: Model-specific arguments used to render ``messages``.
+        chat_template: Per-request template used to render ``messages``.
     """
 
     text: str | None = None
     token_ids: tuple[int, ...] | None = None
+    messages: tuple[dict[str, Any], ...] | None = None
+    tools: tuple[dict[str, Any], ...] | None = None
+    chat_template_kwargs: dict[str, Any] | None = None
```
nv-alicheng, line 310: [Codex] **medium** (data-integrity): `PromptData` carries `messages`, `tools`, `chat_template_kwargs`, `chat_template` — but **not `tool_choice`**, which the adapter both accepts as a per-sample dataset column and forwards to the endpoint (`openai_msgspec_adapter.py:99` ColumnFilter, `:192` `to_endpoint_request(tool_choice=query.data.get("tool_choice"))`). For datasets that set `tool_choice` (e.g. BFCL v4, whose preset notes some servers render differently when tools are present but `tool_choice` is omitted), the server renders the prompt one way while the ISL/baseline path renders another → reported ISL/OSL/TPOT silently diverge from the served tokens, the exact failure mode this PR targets. Thread `tool_choice` through `PromptData` and into `apply_chat_template` alongside the other template inputs.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -447,20 +478,111 @@ def _token_count_message(
...
     ) -> int:
-        """Chat-template message token count without blocking the loop."""
-        if self._thread is None:
-            raise RuntimeError("BatchTokenizer is closed")
-        return await loop.run_in_executor(
-            self._thread, self._token_count_message, content, reasoning, tool_calls
+        kwargs = dict(chat_template_kwargs or {})
+        kwargs.update(
+            tokenize=True,
+            add_generation_prompt=True,
```
nv-alicheng, line 513: [Codex] **medium** (data-integrity): `_token_count_prompt` copies caller kwargs then unconditionally overwrites `add_generation_prompt`:
```
kwargs = dict(chat_template_kwargs or {})
kwargs.update(tokenize=True, add_generation_prompt=True, return_dict=False)
```
A request that intentionally passes `chat_template_kwargs={"add_generation_prompt": false}` (or any conflicting value) is sent to the endpoint one way but counted for ISL with an extra assistant-generation stub, since `openai_msgspec_adapter` forwards `chat_template_kwargs` unchanged to the real request. Systematically wrong prompt-token metrics for those configs. Only `tokenize`/`return_dict` are infra-forced; let `add_generation_prompt` default to `True` **without** clobbering a caller override (e.g. `kwargs.setdefault("add_generation_prompt", True)` after the update of the infra-only keys).

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -447,20 +478,111 @@ def _token_count_message(
...
+            kwargs["tools"] = list(tools)
+        if chat_template is not None:
+            kwargs["chat_template"] = chat_template
+        prompt_messages = _normalize_prompt_messages_for_template(messages)
+        try:
+            encoded = self._tokenizer.apply_chat_template(  # type: ignore[union-attr]
+                prompt_messages, **kwargs
+            )
+            return len(encoded)
+        except Exception as exc:
```
nv-alicheng, line 528: [Code-Quality] **medium** (code-quality): DRY: the warn-once-then-fallback block in `_token_count_prompt` is a near-verbatim copy of the one in `_token_count_message` (~463-472):
```
key = f"{self._tokenizer_name}:{type(exc).__name__}"
if key not in self._fallback_warned:
    self._fallback_warned.add(key)
    logger.exception("apply_chat_template failed for %s (%s); falling back to ...")
```
Both then fall back to `_token_count_text(...)`. Factor the guard/warn into a single `_warn_template_fallback(exc)` helper so the two chat-template paths can't drift.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
```diff
@@ -260,25 +282,30 @@ def __init__(
     def _load_tokenizer(self) -> None:
```
nv-alicheng, line 282: [Claude] **low** (performance): `transformers_logging.set_verbosity_error()` is called only in `_init_worker` (line 182, shard subprocesses), never in the main aggregator process where `_load_tokenizer` runs and the LIVE lane executes `apply_chat_template(tokenize=True)` (456/500) and the no-fast-backend `tokenizer.encode(...)` (400). The rework moved these onto the Python-wrapper path, which emits transformers' "Token indices sequence length is longer than the specified maximum..." warning; for structured ISL on chat endpoints `_token_count_prompt` runs per-sample in the near-hot live lane, so long prompts spam that warning on every count. Call `set_verbosity_error()` in the main process (`__init__`/`_load_tokenizer`) mirroring the worker.

nv-alicheng (commented):
## Review Council — Multi-AI Code Review (hybrid)

Reviewed by: **Codex** + **Claude** + **Code-Quality** | Depth: **standard** (re-review of updated HEAD)

Codex is back online (dmg app CLI). This pass verified the prior critical/high fixes landed and adds net-new findings — see the summary comment.

nv-alicheng:
## Review Council — Multi-AI Code Review (hybrid, re-review)

Reviewed by: **Codex** + **Claude** + **Code-Quality** | Depth: **standard** | HEAD `3f7b66c4`
Codex is back online (dmg-app CLI) — this is a true 3-reviewer hybrid on the updated PR.

### ✅ Prior critical/high fixes — verified landed
Independently confirmed against the new commits:
- **BatchEncoding critical** — all four `apply_chat_template(tokenize=True)` sites (291/301/456/500) now pass `return_dict=False`. ✅
- **`chat_template` not threaded to tokenizer** (viraatc high) — now carried end-to-end: `session.py:270` → `PromptData.chat_template` → `_token_count_prompt`. ✅
- **Empty/multimodal `messages` IndexError** — truthiness guard + fallthrough to `PromptData()`. ✅
- **`_token_count_prompt` safeguards** — tool-call normalization + try/except fallback + `chat_template_kwargs` present. ✅
- **No-fast-backend text lane** — now uses `tokenizer.encode` instead of hard-failing. ✅
- **tiktoken unreachable** — now declared `tiktoken==0.13.0`. ✅

### 🟡 Net-new — Should Fix (medium)

| 1 | `core/types.py` | 309 | data-integrity | **Codex** | `PromptData` drops `tool_choice` (adapter forwards it, `openai_msgspec_adapter.py:192`) → ISL renders a different prompt than the server for `tool_choice`-setting datasets (BFCL v4). Same metrics-divergence class this PR targets. |
| 2 | `token_metrics.py` | 491 | data-integrity | **Codex** | `_token_count_prompt` does `kwargs.update(add_generation_prompt=True)`, **clobbering** a caller `chat_template_kwargs={"add_generation_prompt": false}` the adapter forwards unchanged → systematically wrong ISL for those configs. Use `setdefault`. |
| 3 | `token_metrics.py` | 504 | code-quality | Code-Quality | DRY: the warn-once-then-fallback block is duplicated between `_token_count_prompt` and `_token_count_message` — factor to one helper so they can't drift. |

### 🔵 Net-new — Consider (low)

| 4 | `token_metrics.py` | 282 | performance | Claude | `set_verbosity_error()` only set in the worker, not the main process → per-sample "sequence longer than model_max_length" spam in the near-hot live lane. |
| 5 | `token_metrics.py` | 539 | code-quality | Code-Quality | `MessageInput`/`PromptInput` `elif` arms have identical bodies → collapse to `isinstance(item, MessageInput \| PromptInput)`. _(untouched-line context — not inline)_ |
| 6 | `token_metrics.py` | 485 | code-quality | Code-Quality | `_token_count_prompt`'s `chat_template_kwargs=None`/`chat_template=None` defaults are dead — the only caller always passes all four positionally. |
| 7 | `commands/benchmark/accuracy.py` | 107 | documentation | Claude | Comment references `OslTrigger._extract_text`, which this PR renamed to `_extract_tokenization_input` — dangling symbol ref (repo rule). `accuracy.py` isn't in the diff, so summary-only. |

**Headline:** the two Codex P2s (#1, #2) are novel and hit the PR's own core goal — reported tokens must match served tokens. Both verified against source.

> Commit hygiene: 13 commits / 1 fixup — fine.

viraatc (approved):

src/inference_endpoint/async_utils/services/metrics_aggregator/tokenization.py
hvagadia: These are concrete wrapper types because `TokenizationInput` is a runtime-discriminated union, not only a typing aid. `BatchTokenizer` dispatches each variant to a different operation (`len(token_ids)`, batched text encoding, assistant-message chat-template rendering, or complete-prompt chat-template rendering). Type aliases disappear at runtime and would not give `isinstance`/`match` a distinct semantic variant. `frozen=True` also prevents queued work from being mutated after enqueueing, and `slots=True` keeps these short-lived wrappers lightweight.

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
hvagadia: Yes. `self._queue is None` is a production state: it means no tokenizer was configured or discoverable for the run, so the aggregator is constructed with `tokenizer=None`. Pre-tokenized inputs still produce token metrics synchronously via `len(token_ids)`, while text/message/prompt inputs cannot be tokenized and those token metrics are skipped; non-token metrics continue normally. I added this explanation to the `TokenTrigger` class docstring where the optional queue is consumed.

src/inference_endpoint/async_utils/services/metrics_aggregator/metrics_table.py
hvagadia: Added hierarchy documentation to both `IslTrigger` and `OslTrigger`. For ISL, `PromptData` created by `PhaseIssuer` normally has exactly one representation; defensive precedence is `token_ids` → non-empty text → structured messages. An explicitly empty token-ID tuple intentionally wins and counts as zero rather than falling through. For OSL, reasoning or tool calls select `MessageInput` (including ordinary content); otherwise non-empty output selects `TextInput`. `TextModelOutput` does not carry server token IDs, so the SGLang response token-ID/text combination is not part of this OSL dispatch.

hvagadia: > An explicitly empty token-ID tuple intentionally wins and counts as zero rather than falling through.

To clarify, the code referenced is for ISL path only. The only way we could get inaccurate counts is if someone creates a dataset with empty token ids and non-empty text, but that's a bad dataset design and should be fixed there.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
hvagadia: I consolidated the text-specific collection into a single `indexed_texts` value in the `TextInput` branch. The actual encode still needs to happen after the classification loop: calling it directly inside that branch would tokenize each text separately and lose the single batched fast-tokenizer call. I kept that boundary to preserve batching while making the state carried from the branch explicit.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
hvagadia: Endpoints does not currently expose or forward `add_generation_prompt` as a first-class chat-completion request field. Therefore vLLM uses its top-level request default (`true`), while SGLang’s normal chat path also renders with `add_generation_prompt=True`. Honoring `chat_template_kwargs["add_generation_prompt"] = false` only in the local tokenizer would make ISL counting differ from the prompt actually sent to the model. For that reason, the counter intentionally forces `true` for the request shape Endpoints currently supports. Supporting `false` correctly would require adding and propagating a first-class request field to both the endpoint and token-counting paths. This is also an uncommon mode for normal next-assistant generation, and none of the benchmark configurations in this repository currently use it.

src/inference_endpoint/load_generator/session.py
hvagadia: Added high QPS multi turn data in PR description.

src/inference_endpoint/async_utils/services/metrics_aggregator/token_metrics.py
hvagadia: Tracking here - https://github.com/mlcommons/endpoints/issues/444

PR #445, opened 2026-08-13
feat(ci): PR size (non-test churn) pre-commit hook + size-labeling workflow

arekay-nv:
## What does this PR do?

Adds a PR-size classifier used two ways, implementing the "CI job for LOC count / label large PRs" item from the Endpoints PR Review Policy:

- **pre-commit hook** `pr-size-check` — prints the one-line size class on **every commit** (`verbose: true`). Run `python3 scripts/pr_size_check.py --verbose` for the threshold grid and per-class review requirements.
- **CI workflow** `.github/workflows/pr-size-label.yml` — labels each PR `size/normal | size/large | size/very-large` from the same logic.

Size = non-test churn (additions + deletions and changed-file count) excluding `tests/**`, lockfiles, vendored sources, and generated files. Class = the higher of the lines/files dimensions. **Advisory only — nothing blocks.**

Per its own tool, this PR classifies as **Normal**.

## Type of change

- [x] New feature (tooling / CI)

## Testing

- [x] Tests added — `tests/unit/scripts/test_pr_size_check.py` (21 cases): classify boundaries, numstat parsing, verbose gating, temp-git measurement, `$GITHUB_OUTPUT`, the CI two-ref `--head` path, top-level `*.lock` exclusion, and base-ref auto-detection (candidate scan + `GITHUB_BASE_REF`).
- [x] All new tests pass locally (21 passed)
- [x] Manual testing — verified per-commit hook output and the label logic against a temp repo

## Checklist

- [x] Code follows project style (ruff / ruff-format / license header)
- [x] Pre-commit hooks pass — mypy skipped locally (known macOS `sched_setaffinity` env issue on unrelated files; green on Linux CI)
- [x] Documentation updated — base-branch resolution / forked-PR behavior (below)

## Notes for reviewers

- The labeling workflow uses `pull_request_target` so `GITHUB_TOKEN` can write labels on **fork** PRs (a plain `pull_request` token is read-only for forks). It never checks out or runs PR code — only the trusted base-repo copy of `scripts/pr_size_check.py` over the *fetched* PR diff.
- Because `pull_request_target` runs the workflow from the **base branch**, this very PR won't be auto-labeled; labeling takes effect for PRs opened after this merges.
- **Label provisioning is a one-off manual step** (labels are no longer auto-created). The per-PR workflow only needs `pull-requests: write` to *attach* labels, whereas creating label definitions needs `issues: write`; keeping them separate keeps the per-PR token minimally scoped. Create the labels once with:
  ```bash
  gh label create size/normal     -c 0E8A16 -d "PR Review Policy: <=500 non-test lines & <=20 files" --force
  gh label create size/large      -c FBCA04 -d "PR Review Policy: 501-1500 lines or 21-50 files"     --force
  gh label create size/very-large -c B60205 -d "PR Review Policy: >1500 lines or >50 files"          --force
  ```
  Applying is best-effort: a missing label or transient API error emits a `::warning::` and never fails the PR, and stale `size/*` labels are removed only **after** the new label is successfully applied (so a failed add never leaves a PR unlabeled).

## Base-branch resolution & forked PRs

The size class is computed against a **base branch**. When `--base` / `PR_BASE_REF` are unset, `resolve_base_ref()` tries, in order: `GITHUB_BASE_REF` → `origin/main` → `main` → `upstream/main` → any other `<remote>/main`.

- **CI (labeling workflow):** always measures against the exact `pull_request.base.sha`, so the applied label is correct regardless of local remotes.
- **Local pre-commit in a forked clone:** in the common fork layout — `origin` = your fork, `upstream` = `mlcommons/endpoints` — `origin/main` resolves **before** `upstream/main`. If your fork's `origin/main` is stale, the *local advisory* size can count already-merged upstream commits and over-classify. This affects only the local hint, never the CI label. To pin the base locally, either keep your fork's `main` synced with upstream, or set the base explicitly:
  ```bash
  PR_BASE_REF=upstream/main pre-commit run pr-size-check   # or, ad hoc:
  python3 scripts/pr_size_check.py --base upstream/main --verbose
  ```


leopck (approved):

scripts/pr_size_check.py
```diff
@@ -0,0 +1,248 @@
...
+
+_GRID = "Normal <=500 & <=20 files | Large 501-1500 or 21-50 | Very large >1500 or >50"
+
+
+def classify(churn: int, files: int) -> str:
+    """Return the size class from non-test churn and file count.
+
+    Thresholds are evaluated independently and the higher class wins.
+    """
+    if churn > 1500 or files > 50:
```
nv-alicheng: [Code-Quality] **medium (single source of truth)**: the policy thresholds `500/1500/20/50` live in three places — inlined here in `classify` (`churn > 1500 or files > 50` / `churn > 500 or files > 20`), hand-written into the `_GRID` string (L67: `Normal <=500 & <=20 files | Large 501-1500 or 21-50 | Very large >1500 or >50`), and restated in the module docstring (L23-25). A threshold change must be mirrored in all three or the printed grid silently contradicts the real classification. Suggest named constants (`NORMAL_MAX_LINES=500`, `LARGE_MAX_LINES=1500`, `NORMAL_MAX_FILES=20`, `LARGE_MAX_FILES=50`) with `classify` and `_GRID` both derived from them.

scripts/pr_size_check.py
```diff
@@ -0,0 +1,248 @@
...
+    ":(exclude)vendor/**",
+    ":(exclude)src/inference_endpoint/openai/openai_types_gen.py",
+    ":(exclude)src/inference_endpoint/openai/openapi.yaml",
+    ":(exclude)src/inference_endpoint/config/templates/**",
+    ":(exclude)src/inference_endpoint/evaluation/legacy_mlperf_deepseek_r1/mlperf_eval/**",
+)
+
+# Base refs tried in order when neither PR_BASE_REF nor GITHUB_BASE_REF is set.
+# Covers the common local-checkout and fork-remote layouts.
+_BASE_CANDIDATES = ("origin/main", "main", "upstream/main")
```
nv-alicheng, line 58: [Codex] **low (design)**: `_BASE_CANDIDATES` tries `origin/main` before `upstream/main`. In the common fork layout (`origin` = personal fork, `upstream` = canonical), a stale fork `main` makes the **local** pre-commit hint count already-merged commits and over-classify — exactly the contributor workflow this hook exists to help. You already document this tradeoff in the PR description ("Base-branch resolution & forked PRs"), and it never affects the CI label (that measures `base.sha` directly), so this is not a bug. But since the hint is *for* contributors and the fork layout is their common case, worth reconsidering whether `upstream/main` should take precedence by default (or resolving to whichever candidate yields the nearer merge-base).

scripts/pr_size_check.py
```diff
@@ -0,0 +1,248 @@
...
+_REQUIREMENTS = {
+    NORMAL: "2 approvals (>=1 from @mlcommons/endpoints-developers).",
+    LARGE: 'add a "How to Review" section and split the PR when practical (2 approvals).',
+    VERY_LARGE: 'add "Why This Cannot Be Split" and "How to Review" sections (3 approvals).',
+}
+
+_GRID = "Normal <=500 & <=20 files | Large 501-1500 or 21-50 | Very large >1500 or >50"
+
+
+def classify(churn: int, files: int) -> str:
```
nv-alicheng: [Code-Quality] **low**: `classify(...) -> str` (and the `cls: str` params on `build_summary`/`build_details`/`_write_github_output`) is stringly-typed over a closed 3-value set (`NORMAL`/`LARGE`/`VERY_LARGE`). A `SizeClass = Literal["normal", "large", "very-large"]` alias would make the contract precise and let mypy verify `_REQUIREMENTS[cls]` key exhaustiveness. Pairs naturally with the threshold-constants cleanup above.

scripts/pr_size_check.py
```diff
@@ -0,0 +1,248 @@
...
+    candidates: list[str] = []
+    gh_base = os.environ.get("GITHUB_BASE_REF")
+    if gh_base:
+        candidates += [f"origin/{gh_base}", gh_base]
+    candidates += list(_BASE_CANDIDATES)
+    for ref in candidates:
+        if _rev_parse_ok(ref):
+            return ref
+
+    for remote in _git(["remote"]).stdout.split():
```
nv-alicheng, line 146: [Claude] **low (testing)**: the graceful-degradation fallbacks — the whole point of this resilient advisory check — are untested. (1) The remote-scan base fallback here (`for remote in _git(["remote"]).stdout.split(): ref = f"{remote}/main"`, L129-132): every `test_main_*` resolves the base via `PR_BASE_REF`/`GITHUB_BASE_REF`/`--base`/`_BASE_CANDIDATES`, so this loop never runs. (2) `measure()` returning `None` on a failed merge-base/diff and the corresponding "could not diff … skipping" path in `main` (L230-235) — `test_main_skips_when_base_unresolvable` only covers `base_ref is None`, not the unrelated-history/shallow-clone case. The workflow's "No size class computed … nothing to label" handling depends on these skipping cleanly. A small unit test for each (a repo with only an `upstream/main` remote; a diff against unrelated history) closes the gap.

nv-alicheng (commented):

Reviewed by: **Codex** + **Claude** + **Code-Quality** | Depth: **standard**

Found **4** issues (0 critical/high). Solid, well-tested tooling PR. **Security note:** all three reviewers independently cleared the `pull_request_target` workflow — it checks out the trusted `base.sha`, fetches the PR head as data only (never executes PR code), validates `PR_NUMBER` against injection, uses env-var indirection (no untrusted PR fields in `run:`), minimizes `permissions`, and pins the action SHA. No token-exfil / script-injection vector found.

| 1 | `scripts/pr_size_check.py` | 75 | medium | code-quality | Code-Quality | Thresholds `500/1500/20/50` duplicated across `classify`, the `_GRID` string, and the docstring — no single source; grid can silently drift from behavior. |
| 2 | `scripts/pr_size_check.py` | 57 | low | design | Codex | `origin/main` resolved before `upstream/main`; stale fork main over-classifies the local hint (documented tradeoff; CI label unaffected). Reconsider default precedence. |
| 3 | `scripts/pr_size_check.py` | 70 | low | code-quality | Code-Quality | `classify() -> str` stringly-typed over a closed 3-value set; a `Literal` would let mypy check `_REQUIREMENTS` exhaustiveness. |
| 4 | `scripts/pr_size_check.py` | 129 | low | testing | Claude | Resilience fallbacks (remote-scan base resolution; `measure()` -> None skip path) untested — the exact CI-degradation paths the design relies on. |

_Codex also raised a finding on `tests/performance/.../test_token_metrics_perf.py`; dropped as out of scope (not in this PR's diff)._

.github/workflows/pr-size-label.yml
```diff
@@ -0,0 +1,79 @@
...
+          # Strip the other size/* labels only after the add succeeds, so a
+          # failed add never leaves the PR with no size label at all.
+          if gh pr edit "$PR" --add-label "size/$CLS"; then
+            for other in normal large very-large; do
+              if [ "$other" != "$CLS" ]; then
+                gh pr edit "$PR" --remove-label "size/$other" 2>/dev/null || true
+              fi
+            done
+          else
+            echo "::warning::Could not apply label size/$CLS — ensure the size/* labels exist (create them once with 'gh label create')."
```
nvzhihanj, line 78: I didn't seem to see it in the labels. Can we use this PR to test it out?

nvzhihanj (approved):

.github/workflows/pr-size-label.yml
arekay-nv: They should be visible in the labels section on the right. The job won't trigger until it is merged. I will update this PRs label manually.
image

scripts/pr_size_check.py
arekay-nv: Thanks for raising this. We will keep the current precedence intentionally. The local check treats `origin/main` as the default authoritative base; optimizing auto-detection for fork layouts is outside the intended scope of this advisory hook. Fork users can explicitly select the canonical base with `PR_BASE_REF=upstream/main` or `--base upstream/main`. The CI workflow is unaffected because it measures against the PR exact `base.sha`. Given that, we are treating this as an accepted design tradeoff rather than changing the precedence.

scripts/pr_size_check.py
arekay-nv: Addressed in the latest revision. `test_resolve_base_ref_scans_remotes_when_candidates_absent` exercises the arbitrary `<remote>/main` fallback using `myfork/main`, ensuring the final remote-scan loop is actually reached. `test_measure_returns_none_on_unrelated_history` and `test_main_skips_when_diff_fails_on_unrelated_history` cover the failed merge-base and graceful `main()` skip paths, including the no-label behavior. No additional code changes should be needed for this thread.

PR #446, opened 2026-08-13
Make agentic routing headers configurable

hvagadia:
## Summary

- add a YAML-configurable list of routing headers for agentic inference requests
- populate each configured header with the conversation ID
- preserve `X-Session-ID` as the default when `routing_headers` is omitted
- allow `routing_headers: []` to disable conversation routing headers

Example:

```yaml
agentic_inference:
  routing_headers:
    - X-Session-ID
    - X-SMG-Routing-Key
```

## Validation

- `uv run pytest -q tests/unit/load_generator/test_async_session.py tests/unit/config/test_yaml_loader.py tests/unit/commands/test_benchmark.py::TestBuildPhases` (81 passed)
- `uv run pre-commit run --all-files`

### DIFF
--- a/src/inference_endpoint/commands/benchmark/execute.py
+++ b/src/inference_endpoint/commands/benchmark/execute.py
@@ -572,6 +572,16 @@ def _build_phases(
     drain_cfg = ctx.config.settings.drain
 
     if ctx.dataloader is not None and ctx.rt_settings is not None:
+        perf_dataset = next(
+            dataset
+            for dataset in ctx.config.datasets
+            if dataset.type == DatasetType.PERFORMANCE
+        )
+        routing_headers = (
+            perf_dataset.agentic_inference.routing_headers
+            if perf_dataset.agentic_inference is not None
+            else ()
+        )
         warmup_cfg = ctx.config.settings.warmup
         if warmup_cfg.enabled:
             warmup_dataset: Dataset = (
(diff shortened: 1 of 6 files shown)


src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -572,6 +572,16 @@ def _build_phases(
...
 
     if ctx.dataloader is not None and ctx.rt_settings is not None:
+        perf_dataset = next(
+            dataset
+            for dataset in ctx.config.datasets
+            if dataset.type == DatasetType.PERFORMANCE
+        )
+        routing_headers = (
+            perf_dataset.agentic_inference.routing_headers
+            if perf_dataset.agentic_inference is not None
```
nvzhihanj, line 582: Should we unconditionally accept routing headers (not just on agentic inference dataset)?

nvzhihanj (approved):

leopck (approved):
LGTM, much more scalable

src/inference_endpoint/commands/benchmark/execute.py
hvagadia: So the yaml only specifies key in the headers. The value is the session-id that comes from the dataset. It's a field in our multi-turn dataset, but that won't be present in others and hence won't work OOTB. We can think about adding it to other datasets separately.

PR #447, opened 2026-08-13
docs: steady-state metrics reporting design proposal

nv-alicheng:
Post-processing step (and ad-hoc CLI over events.jsonl) that reports steady-state metrics — super-pass windowing, guarded drain-tail cut, CoV ensemble + trend gate (drift up/down), hairball, edge-case handling.

## What does this PR do?

Adds a design doc for a new steady-state window detection and metrics reporting feature

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor/cleanup

## Related issues

cc @arekay-nv to tag any related issues

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass

### DIFF
--- a/docs/steady-state-detection.md
+++ b/docs/steady-state-detection.md
@@ -120,6 +120,12 @@ result. Definitions used throughout:
   issued.
 - **Long-running sample.** A sample whose output length (OSL) or sample latency is
   much larger than the dataset average.
+- **Level shift (staircase).** A step change from one flat metric level to a higher
+  flat level partway through a run — distinct from a gradual drift or the drain tail.
+  Typically caused by long-running samples triggering KV-cache eviction toward the
+  end of a long run, or sudden failures in a subset of workers partway or at the end of a run.
+  It produces two (or more) legitimate plateaus; the window selection (§5.5) reports the first
+  and flags the shift to be reported as an anomaly.
 - **Hairball.** A build-up of long-running samples that grows as more dataset
   passes are issued and completed: because datasets are issued without
   replacement, fresh copies of a long-running sample are issued before earlier
@@ -212,10 +218,13 @@ The core is a sequence of pure stages over the per-super-pass series.
   [ guarded drain-tail cut ]              only if per-token-invariant
      |
      v
-  [ convergence: CoV ensemble + trend ]   window edge + steady/drift verdict
+  [ plateau segmentation: CoV + trend ]   admissible windows -> plateaus
      |
      v
-  { steady_state metrics + status }
+  [ select first plateau + shift flag ]   MSER precision; Pettitt level-shift
+     |
+     v
+  { steady_state metrics + status + anomaly }
 ```
 
 ### The hairball (drain tail)
@@ -264,53 +273,108 @@ workloads the tail is a strong long-output selection, so output-length coverage 
 reported alongside the steady metrics so a reader can see the steady set
 under-samples the output-length tail.
 
-## 5.5 Convergence: CoV ensemble, trend gate, drift up/down
+## 5.5 Convergence and steady-window selection
 
-The window edge and the steady/drift verdict come from two signals combined.
+The reported window and the steady/drift verdict come from two per-metric signals —
+a coefficient-of-variation (CoV) stopping rule and a mandatory trend gate — combined
+under a window-selection rule adapted from the steady-state simulation literature.
 
-**Metric set.** The convergence and trend tests run on TTFT and TPOT at the 50th,
-95th, 99th, and 99.9th percentiles (the 99.9th only where a super-pass holds
-enough samples to estimate it). End-to-end latency is reported as context only,
-never as a convergence signal (§5.1). A window is steady only when _every_ metric
-and percentile plateaus and is within its CoV threshold.
-
-_note: There is some discussion to relax core reported metric percentiles to p95 from p99.
-In this case, then the p99 and p99.9 percentiles not converging will be a warning, not a
-hard-fail signal._
+**Metric set.** The convergence and trend tests run on TTFT and TPOT. The **p50 and
+p95** percentiles are _gating_: a window is steady only when both plateau and are
+within CoV for both metrics. The **p99** (and p99.9 where a super-pass holds enough
+samples to estimate it) are carried as _diagnostic warnings_, not hard gates — a tail
+percentile that fails to converge is surfaced as a warning rather than voiding the
+window, because tail percentiles are estimated from far fewer samples per super-pass
+and are the noisiest signal available. End-to-end latency is reported as context
+only, never as a convergence signal (§5.1).
 
 **CoV stopping rule.** The coefficient of variation `CoV = sigma / mu` of a
-metric's per-super-pass percentile, over a trailing window of super-passes, is a
+metric's per-super-pass percentile, over a window of super-passes, is a
 scale-free measure of how much the metric is still moving relative to its own
-level. A region is a candidate steady state when `CoV < bound` for every tracked
-metric and percentile. The bound loosens toward the tail (a p99 is estimated from
+level. A region is a candidate steady state when `CoV < bound` for every gating
+metric and percentile. The bound loosens toward the tail (a p95 is estimated from
 fewer samples per super-pass, so its sampling-noise floor is higher than a p50's).
 
-**Trend gate (mandatory), drift up vs down.** A low CoV over a _trailing_ window
-certifies local flatness, not global convergence: a slowly drifting series can
-sit locally flat while climbing overall. An ordinary-least-squares trend test over
-the whole post-warmup series is therefore applied on top. A metric is _drifting_
-when the run-length change is both a large fraction of its typical value and large
-relative to the super-pass-to-super-pass residual scatter (so a noisy-but-flat
-series is not mistaken for a trend). The slope sign classifies the metric into one
-of three states — **Drifting Down**, **Plateau**, **Drifting Up**:
+**Trend gate (mandatory), drift up vs down.** A low CoV over a window certifies
+local flatness, not that the metric has stopped moving: a slowly drifting series can
+sit locally flat while climbing overall. A trend test over the window is therefore
+applied on top. Because per-super-pass series are autocorrelated, the primary gate is
+the rank-based **Mann–Kendall test with the Hamed–Rao autocorrelation correction**
+(so serial correlation does not fake a trend); an OLS slope-vs-scatter check and a
+Newey–West (autocorrelation-consistent) slope test corroborate it. The verdict
+classifies each metric into one of three states — **Drifting Down**, **Plateau**,
+**Drifting Up**:
 
-- **Drifting Up** — the metric worsens across the run (for example a p99 TTFT tail
-  that never plateaus). This is pathological: there is _no_ steady state to report
-  for that metric, and the step says so rather than emitting a number.
-- **Plateau** — near-zero slope: the metric is genuinely steady. Only a Plateau
+- **Drifting Up** — the metric worsens across the window (for example a p99 TTFT tail
+  that never plateaus). Pathological: there is _no_ steady state to report for that
+  metric, and the step says so rather than emitting a number.
+- **Plateau** — no significant trend: the metric is genuinely steady. Only a Plateau
   metric is eligible to contribute a steady value.
-- **Drifting Down** — the metric is still settling downward, i.e. the warmup crop
-  was slightly short. This is transparent (not a false alarm); the follow-up is a
-  larger crop for that metric.
+- **Drifting Down** — the metric is still settling downward, i.e. the warmup crop was
+  slightly short. Transparent (not a false alarm); the follow-up is a larger crop.
 
-**Ensemble and window selection.** CoV alone is insufficient — a window is only
-meaningful if the metric is in Plateau. In practice, over a large set of runs, no
-single `(window, bound)` fits all metrics and workloads, so the rule is run as an
-**ensemble** of preset `(window, bound)` settings. Among the settings that report
-a steady window, the **largest** steady window is selected (most statistically
-significant). Detector concordance (agreement on where the run settles) is a
-corroborating guardrail; the trend gate remains mandatory and primary, the CoV
-ensemble secondary.
+**Selection principle (MSER): maximize precision, never the metric.** Once a run may
+contain several windows that are both in Plateau and within CoV, the question is
+which to report. The governing rule is taken from the Marginal Standard Error Rule
+(MSER; White 1997) for steady-state truncation: **select the window by the precision
+of the estimate, never by the value of the reported metric.** Choosing the window by
+the very quantity being reported — the highest-TPS window, say — is selection bias:
+it reports the most favorable noise realization and inflates the number. MSER instead
+minimizes the standard error of the mean, `SE = sigma / sqrt(n)`. Because `n` grows
+with window length, `SE` is minimized by the _longest_ admissible window unless
+extending it drags in non-steady super-passes that inflate `sigma` faster than `n`
+grows — the classic bias-versus-variance knee. The operational consequence is simply:
+prefer more steady data, chosen by a criterion that never looks at the throughput
+level.
+
+**Admissibility.** A candidate window is _admissible_ when, for every gating metric
+and percentile, it is (1) in Plateau (trend gate) and (2) within at least one CoV
+bound of the ensemble. These two gates are the explicit form of MSER's implicit
+"post-transient" restriction, and they make each rejection interpretable (a window is
+excluded for a named reason, not a black-box score).
+
+**Candidate family — general contiguous windows, not fixed-endpoint truncation.**
+MSER's textbook form fixes the window's right edge at the end of the run and moves
+only the start: it drops warm-up and keeps everything to the end. That is insufficient
+here because of a possible failure mode seen in longer runs — a **staircase**: a flat region
+that steps up to a higher flat region, possible if KV-cache eviction is triggered towards
+the end of the run, or if workers suddenly crash, causing overall max throughput to degrade
+by a constant amount. A staircase contains a legitimate steady plateau that
+**ends before the end of the run**, and a fixed-endpoint window cannot
+isolate it — it can only capture the final, degraded plateau, or fail the gate on the
+jump. The candidate family is therefore **all contiguous super-pass windows**
+`[lo, hi)`, searched from longest down to a minimum-length floor (the trend test's
+minimum sample count). The floor is essential: minimizing `SE` over unconstrained
+contiguous windows is degenerate — a two-point flat window has `SE = 0` — so the floor
+together with the Plateau/CoV gates rules out the trivial micro-window solution.
+
+**Plateau segmentation and the reported window (first plateau).** The trend and CoV
+gates _implicitly segment_ the run: a window spanning a staircase jump has high CoV
+and reads as a trend, so it is inadmissible, while a window inside a single plateau is
+admissible. Growing an admissible window from the first post-warmup super-pass until
+admissibility breaks isolates the **first plateau**; resuming past the break isolates
+each subsequent plateau. **The first plateau is the reported steady state.** This
+deliberately overrides the pure longest/min-`SE` choice, because empirically the later
+plateaus of a staircase are generally _degradation_ steps — a skewed long-output workload
+building up, or a server going unhealthy — so the first plateau is the representative
+healthy steady state and the later steps are anomalies, not the number to report.
+
+**Level shifts (staircase) are detected and flagged, never hidden.** Reporting the
+first plateau must not silently discard the fact that the run degraded. A level-shift
+detector runs alongside the segmentation: when two or more disjoint admissible
+plateaus have pooled means differing by more than the CoV band, corroborated by a
+**Pettitt** change-point test (a nonparametric, rank-based single-change-point test
+that pairs naturally with the rank-based trend gate) on the TPOT series, the result
+carries an **anomaly flag** — the change-point super-pass and the magnitude and
+direction of the shift. The steady result is still reported from the first plateau,
+with an explicit, honest note that the run changed regime toward the end; every
+plateau is carried in the machine-readable output.
+
+**Ensemble.** Because no single `(window, bound)` CoV setting fits all metrics and
+workloads, the CoV rule is evaluated as an **ensemble** of preset settings; a window
+passes CoV if at least one ensemble setting certifies it for every gating metric.
+Detector concordance is a corroborating guardrail. The trend gate remains mandatory
+and primary; the CoV ensemble is secondary.
 
 ## 5.6 Edge cases and error handling
 
@@ -384,8 +448,20 @@ machine-readable summary produced from the report). Each tracked metric carries
 its steady value plus its state (`Plateau` / `Drifting Up` / `Drifting Down`), and
 the run carries its coverage `status`. Reported quantities:
 
-- **TPS** (output tokens/s).
-- **TTFT and TPOT** percentiles and histograms.
+- **The steady window** itself: its super-pass range `[start, end)` and sample count,
+  so a reader can see where in the run it was drawn from.
+- **TPS**, reported two ways because they answer different questions: **per-user TPS**
+  = `1 / mean(TPOT)` (output tokens/s/user, the interactivity number) and **system
+  TPS** = total output tokens in the window divided by its wall-clock span
+  (aggregate throughput). Each carries a confidence interval computed by
+  **non-overlapping batch means** with super-passes as batches, which accounts for
+  the per-super-pass autocorrelation rather than assuming independent samples.
+- **TTFT and TPOT** percentiles (p50/p90/p95/p99) and histograms, over the pooled raw
+  samples of the steady window.
+- **Anomaly.** When the level-shift detector fires (§5.5), an `anomaly` block records
+  the change-point super-pass, the shift magnitude and direction, and every detected
+  plateau — so a degradation toward the end is surfaced next to the (first-plateau)
+  steady result rather than hidden by it.
 - **QPS is dropped** for text/token-based LLM workloads — a request is not a unit
   of work when output length varies widely, so QPS is a legacy metric of little
   meaning; it is retained only for token-free, uniform-work loads.
@@ -414,10 +490,14 @@ per-super-pass series.
 - **Guard distance measure.** The exact two-sample statistic and acceptance bound
   for the per-token-invariance guard (§5.4) need to be fixed.
 - **No-steady-state runs.** When no steady state is found (a tracked metric is
-  Drifting Up), should the run be reported _invalid_ — analogous to legacy
-  LoadGen's statistical-significance gate — or reported with the offending metrics
-  flagged as unstable while the rest are reported steady? The current proposal is
-  the latter (show which metrics were stable vs unstable).
+  Drifting Up _throughout_, with no admissible plateau at any window), should the run
+  be reported _invalid_ — analogous to legacy LoadGen's statistical-significance gate
+  — or reported with the offending metrics flagged as unstable while the rest are
+  reported steady? The current proposal is the latter (show which metrics were stable
+  vs unstable). This is distinct from the **staircase** case, which is already decided
+  (§5.5): a run with a first steady plateau followed by a higher plateau _does_ have a
+  steady state — the first plateau is reported and the later shift is flagged as an
+  anomaly, not treated as no-steady-state.
 - **Offline drain-onset.** In offline/max-throughput the drain has no client-side
   boundary; the steady region is read from the TPS trend, and the robust definition
   (server occupancy dropping below saturation) is server-side and currently
--- a/docs/steady-state-detection.md
+++ b/docs/steady-state-detection.md
@@ -1,0 +1,440 @@
+# Steady-State Metrics Reporting
+
+# 1 Objective {#1-objective}
+
+Add a post-processing step that runs after a benchmark completes and reports the
+sustained steady-state metrics, rather than the whole-run average,
+which is deflated by the ramp-up and drain transients. The step is a pure
+function over the durable event log; it also ships as an ad-hoc command-line tool
+that ingests an `events.jsonl` file.
+
+Goals:
+
+- Emit a `steady_state` block alongside the existing whole-run (`total`) metrics
+  in the run report, reported as the **official result** with the whole-run
+  (`total`) metrics retained as supplementary context.
+- Detect _when there is no steady state_ (a progressively degrading run) and say
+  so, instead of reporting an unstable/unsteady result.
+- Provide a single implementation reachable two ways: automatically as the
+  post-processing step at the end of a run, and manually as the same step invoked
+  ad hoc over any recorded `events.jsonl`.
+- Add no cost to the measured run: all work is off the hot path.
+
+Non-goals:
+
+- Changing the whole-run (`total`) numbers, the live metrics snapshot cadence, or
+  the wire schema of the live aggregator.
+- Changing how load is issued during a run. A staggered-issuance option is
+  discussed as complementary (§5.7) but is out of scope for this step.
+- Accuracy scoring, submission checking, or any change to the audit path.
+
+# 2 Background {#2-background}
+
+A run's reported metrics today are aggregated over the entire measurement window.
+Two regions of that window are not steady state:
+
+- **Ramp-up.** At the start of a run the client raises offered load to its target.
+  Under a concurrency load pattern the target in-flight population is filled as a
+  burst; against a finite-rate server the leading requests queue and
+  time-to-first-token (TTFT) inflates, and the inflation grows with concurrency.
+- **Ramp-down (drain).** After issuance stops, in-flight decays below the target
+  while the last requests finish. Throughput deflates because the wall-clock
+  denominator keeps advancing while offered load is below steady state.
+
+Averaging over these transients understates throughput and overstates tail
+latency. The magnitude is workload-dependent and can be large for the tail: in
+experiments over recorded runs (single-turn concurrency, offline/max-throughput,
+Poisson, and multi-turn agentic), the reported p99 TTFT
+was dominated by the ramp spike and fell substantially once the ramp was
+excluded, while per-token latency (TPOT) was essentially unchanged.
+
+There is already a precedent in the codebase for exactly this shape of feature:
+`src/inference_endpoint/metrics/early_stopping.py` computes MLPerf early-stopping
+percentile estimates as a cold-path calculation, and
+`scripts/early_stopping_estimate_from_events.py` re-runs the same math ad hoc
+from a recorded `events.jsonl` (see `docs/early_stopping.md`). This design follows
+the same two-entry-point pattern. There is also a precedent for a post-run
+orchestration step in `src/inference_endpoint/commands/audit.py`, which
+`src/inference_endpoint/commands/benchmark/cli.py` dispatches after the main
+benchmark completes.
+
+# 3 Assumptions and Risks {#3-assumptions-and-risks}
+
+- **The durable event log is complete and authoritative.** The step reads the
+  per-sample event stream, not the live snapshot (which can lag under load). Risk:
+  a run killed by `SIGKILL`/OOM before the log is flushed yields a truncated log;
+  the step must degrade to a best-effort result with a status flag rather than
+  fail (§5.6).
+- **Per-token latency is approximately sample-invariant in a healthy run.** The
+  guarded tail-cut (§5.4) rests on this. It held across the recorded corpora, but
+  it is a property of a healthy run, not a guarantee, so the cut is _guarded_: it
+  is applied only when the condition is measured to hold, and otherwise the tail
+  is retained.
+- **Bucketing is by issue order.** The unit of analysis is a fixed-size group of
+  issued requests (§5.1). Risk: for load modes with no repeated dataset pass, the
+  group size is a free parameter; experiments show the qualitative verdict is
+  robust to that choice, but it is called out as tunable (§6).
+- **Token counts are available or derivable.** TPOT needs an output-token count
+  per request. When the run already records one it is used directly; otherwise the
+  step tokenizes on the cold path (§5.6). Risk: cost on very large logs, bounded
+  by sampling.
+
+# 4 Alternatives considered {#4-alternatives-considered}
+
+- **Fix it live, in the aggregator.** Detect and crop the ramp inside the hot-path
+  aggregator so `total` is already steady. Rejected: it adds latency-critical work
+  to the hot path, the live snapshot can lag under load, and it couples a
+  still-evolving heuristic to the measured numbers. A cold-path step keeps the hot
+  path untouched and the heuristic revisable.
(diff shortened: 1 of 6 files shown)


docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+  burst; against a finite-rate server the leading requests queue and
+  time-to-first-token (TTFT) inflates, and the inflation grows with concurrency.
+- **Ramp-down (drain).** After issuance stops, in-flight decays below the target
+  while the last requests finish. Throughput deflates because the wall-clock
+  denominator keeps advancing while offered load is below steady state.
+
+Averaging over these transients understates throughput and overstates tail
+latency. The magnitude is workload-dependent and can be large for the tail: in
+experiments over recorded runs (single-turn concurrency, offline/max-throughput,
+Poisson, and multi-turn agentic, including runs on Hecate), the reported p99 TTFT
```
nvzhihanj: Remove the name of the cluster (and overwrite the commit please)

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+`src/inference_endpoint/commands/benchmark/cli.py` dispatches after the main
+benchmark completes.
+
+# 3 Assumptions and Risks {#3-assumptions-and-risks}
+
+- **The durable event log is complete and authoritative.** The step reads the
+  per-sample event stream, not the live snapshot (which can lag under load). Risk:
+  a run killed by `SIGKILL`/OOM before the log is flushed yields a truncated log;
+  the step must degrade to a best-effort result with a status flag rather than
+  fail (§5.6).
```
nvzhihanj, line 67: Maybe I am not reading it in the right way - but I thought it should fail if the original run fails?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
+# Steady-State Metrics Reporting
+
+# 1 Objective {#1-objective}
+
+Add a post-processing step that runs after a benchmark completes and reports the
+metrics the system _sustains in steady state_, rather than the whole-run average,
```
arekay-nv: reword `the metrics the system _sustains in steady state_`  - one suggestion: `reports the sustained steady state metrics`

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+function over the durable event log; it also ships as an ad-hoc command-line tool
+that ingests an `events.jsonl` file.
+
+Goals:
+
+- Emit a `steady_state` block alongside the existing whole-run (`total`) metrics
+  in the run report, reported as the **official result** with the whole-run
+  (`total`) metrics retained as supplementary context.
+- Detect _when there is no steady state_ (a progressively degrading run) and say
+  so, instead of reporting a false point estimate.
```
arekay-nv: suggestion - "false point estimate" -> "unstable/unsteady result"?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+- Provide a single implementation reachable two ways: automatically at the end of
+  a run, and manually as a CLI over any recorded `events.jsonl`.
```
arekay-nv: suggestion - `manually as a CLI` -> `manually as a post-processing step`

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+  copies finish, so long-running samples remain in flight long after the rest have
+  completed.
+- **Hairball weight.** At issuance-stop under max-concurrency `C`, the percentage
+  of in-flight samples that are _not_ among the last `C` issued (the lingering
+  hairball). Ideally 0 — at stop the `C` in-flight samples would be exactly the
+  last `C` issued.
+- **Relative active concurrency.** In-flight count as a percentage of the
+  max-concurrency budget (100% at saturation).
+- **Issue-time window.** A contiguous range of super-passes `[start, end)`. Its
+  measured set is every request _issued_ in that range. Throughput uses the
```
arekay-nv, line 141: This might requires some clarification - as far as i understand from below `a sample issued but whose response is not received has no logged TTFT and is naturally excluded`. So either it is `issue only` or `issue with at least one event` to be included in a window.

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+## 5.1 Overview and definitions
+
+The step is a pure function from a recorded event stream to a `steady_state`
+result. Definitions used throughout:
+
+- **Healthy server.** A server that can support the maximum load issued by the
+  client. The window measures the sustained behavior of a healthy server; a
+  genuinely unhealthy server produces bad-but-real numbers, which the drift
+  detector (§5.5) distinguishes from transient pollution.
+- **Super-pass.** One full **dataset pass** — `dataset_size` consecutive requests
```
arekay-nv: minor - would it be easier to just call it a `pass` instead of a `super-pass` since it is a single pass over the entire dataset?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+
+- **No steady state.** If a tracked metric is Drifting Up, report drift for that
+  metric instead of a point estimate (§5.5). This is a first-class outcome (see the
+  open question on invalid runs, §6).
+- **Truncated / interrupted log.** If the run did not complete cleanly, the step
+  computes a best-effort result over whatever was logged and marks it as partial,
+  mirroring how the report already distinguishes interrupted runs.
+- **Missing token counts.** If the log carries no per-request output-token count,
+  TPOT is derived by tokenizing outputs on the cold path. For large logs this is
+  bounded by tokenizing a sample sufficient for the per-super-pass percentiles;
```
arekay-nv: so the full pass (or super-pass) is not tokenized?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+  full re-tokenization of every output is avoided. If neither a count nor a
+  tokenizer is available, the step falls back to TTFT-only and records that TPOT
+  was not assessed (a metric with no data is skipped, not treated as zero, so an
+  absent metric can never fake convergence).
+- **Degenerate modes.** Offline/max-throughput has a degenerate issue time
+  (everything issued at `t=0`), so there is no client-side issue-time window and no
+  client-side drain boundary. The step detects the mode and finds the steady region
```
arekay-nv, line 534: for TPOT calculations, we could potentially use the completion-first_token time, but that would require the  server to not evict requests in decode phase.

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+  from the **TPS trend** over time — the plateau before throughput falls off —
+  rather than an issue-time window; the exact drain-onset (server occupancy dropping
+  below saturation) is server-side and is handwaved for now (§5.3). Near-saturation
+  Poisson backlogs like offline and is detected by the completion rate falling below
+  the offered rate.
+
+## 5.7 Complementary: staggered ("feathered") issuance
+
+The ramp exists because the target concurrency `C` is filled as a burst. A
+staggered fill flattens the ramp-up spike: issue in steps of `ceil(C/k)` and, between
```
arekay-nv: add a small definition for `k`.

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+- **TPS** (output tokens/s).
+- **TTFT and TPOT** percentiles and histograms.
+- **QPS is dropped** for text/token-based LLM workloads — a request is not a unit
+  of work when output length varies widely, so QPS is a legacy metric of little
+  meaning; it is retained only for token-free, uniform-work loads.
+- **ISL / OSL** are reported for analytics, not as validation numbers: once the
+  window is allowed to drop samples (not enforcing full dataset-pass boundaries),
+  the input/output-length distribution is skewed relative to the constructed
+  dataset and is no longer a meaningful validation quantity, though it remains
+  useful for analysis.
```
arekay-nv, line 587: If the window is allowed to drop samples, should the ISL/OSL statistics be computed for the samples included in the measurement?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+
+The `total`-vs-`steady_state` divergence is itself surfaced: a large gap indicates
+excessive ramp relative to run length, or a run too short to window. Existing
+plotting (`src/inference_endpoint/metrics/results_plots.py`,
+`scripts/plot_results.py`) can be extended to overlay the window on the
+per-super-pass series.
+
+# 6 Open questions {#6-open-questions}
+
+- **Default parameters.** The warmup band, trailing-window length, per-percentile
```
arekay-nv, line 597: Can we ignore the first pass/super-pass over the dataset as a warmup pass and the other parameters can be benchmark specific decided by TF?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,412 @@
...
+  over recorded runs; the defaults should be reviewed on a wider set before they
+  are locked as the official reporting parameters.
+- **Super-pass sizing for modes without dataset repetition.** For multi-turn
+  agentic and other single-pass workloads there is no natural dataset-pass unit.
+  Experiments show the steady/drift verdict is robust to the group-size choice, but
+  a principled default (for example keyed to concurrency) is still to be picked.
+- **Guard distance measure.** The exact two-sample statistic and acceptance bound
+  for the per-token-invariance guard (§5.4) need to be fixed.
+- **No-steady-state runs.** When no steady state is found (a tracked metric is
+  Drifting Up), should the run be reported _invalid_ — analogous to legacy
```
arekay-nv: We should declare the gates for each benchmark and also define whether a submission should be invalid if it fails a certain gate. It should be a benchmark TF decision on what gates to include.

docs/steady-state-detection.md
nv-alicheng: What does "failing" mean here — a request timeout or a process crash? My intent is that a **timeout** should still yield a best-effort (partial) result over whatever was logged, while a hard **crash** of the post-process can just fail. Could you clarify which case you had in mind so I word the edge-case handling to match?

docs/steady-state-detection.md
nv-alicheng: Kept `super-pass` deliberately rather than `pass`: it is **not** always one full dataset pass. A low-concurrency run may not issue even a single full pass, so reusing `pass` would create a terminology conflict. It's also a tunable hyperparameter — e.g. 2 dataset passes per super-pass to reduce per-super-pass variance. Reworded the §5.1 definition in bad5e83f to make this explicit.

docs/steady-state-detection.md
```diff
@@ -0,0 +1,440 @@
...
+## 5.1 Overview and definitions
+
+The step is a pure function from a recorded event stream to a `steady_state`
+result. Definitions used throughout:
+
+- **Healthy server.** A server that can support the maximum load issued by the
+  client. The window measures the sustained behavior of a healthy server; a
+  genuinely unhealthy server produces bad-but-real numbers, which the drift
+  detector (§5.5) distinguishes from transient pollution.
+- **Super-pass.** The atomic unit of the analysis: a contiguous block of requests
```
hvagadia, line 112: Is there an assumption here that we need more than one super-pass? How many would be needed to get a good estimate?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,440 @@
...
+The core is a sequence of pure stages over the per-super-pass series.
+
+```text
+  event log
+     |
+     v
+  [ ingest -> per-super-pass series ]     issue-order bucketing
+     |
+     v
+  [ adaptive warmup crop ]                remove the ramp
```
hvagadia: How is ramp detected?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,440 @@
...
+- **Plateau** — near-zero slope: the metric is genuinely steady. Only a Plateau
+  metric is eligible to contribute a steady value.
+- **Drifting Down** — the metric is still settling downward, i.e. the warmup crop
+  was slightly short. This is transparent (not a false alarm); the follow-up is a
+  larger crop for that metric.
+
+**Ensemble and window selection.** CoV alone is insufficient — a window is only
+meaningful if the metric is in Plateau. In practice, over a large set of runs, no
+single `(window, bound)` fits all metrics and workloads, so the rule is run as an
+**ensemble** of preset `(window, bound)` settings. Among the settings that report
```
hvagadia: Can you explain what does "ensemble of (window, bound)" mean exactly? How are the values of window and bound decided?

docs/steady-state-detection.md
nv-alicheng: Two run-failure modes, handled differently. A **timeout / interrupt** (SIGTERM, drain-timeout) still flushes a partial event log, so the step computes a **best-effort** steady result over what was logged and marks `status = partial` — the same way the report already flags interrupted runs. A **hard crash before the log is flushed** (SIGKILL/OOM) yields a truncated log; the step still degrades to best-effort + partial rather than erroring. The post-process itself never hard-fails a run. Separately, a run that *completes but never reaches steady state* is reported with the offending metrics flagged unstable (§5.6), not failed — whether that should count as an invalid submission is the open TF question in §6. Does that cover the case you had in mind?

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: Membership is **issued within the window**; a sample then contributes to a given metric only if that metric was **logged inside the window** (e.g. a first-token event for TTFT). So an issued-but-no-response sample is a member of the issue set (it counts toward the offered / issue-time denominator) but contributes no TTFT/TPOT — hence "naturally excluded" from the latency families, not from issuance. I'll reword §5.1 to separate *membership* from *per-metric contribution* so it's unambiguous.

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: Addressed above and reworded in §5.1: `super-pass` is kept deliberately because it is **not** always one full dataset pass (a low-concurrency run may not issue even one full pass, and it's a tunable hyperparameter). Resolving.

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: Correct — not every output needs tokenizing. TPOT only needs each super-pass's *percentiles*, so a sample of outputs per super-pass is sufficient; full re-tokenization is avoided on large logs (§5.6). Note: the reference script (`scripts/steady_state_diagnostics.py`) currently tokenizes all completions for simplicity — the per-super-pass sampling is a design allowance to add if the cold-path cost becomes a concern.

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: Agreed — that's exactly the §5.6 caveat: the timestamp-derived TPOT (`complete − first_token`) assumes the request stays resident through decode; if the server evicts/pages it mid-decode, the span includes non-decode queue time and inflates TPOT, so a server-reported per-token count is preferred where eviction is possible. This is also why the design now adds **staircase / level-shift detection** (Pettitt + multi-plateau): KV-cache-eviction-driven degradation toward the end of long runs shows up as a step in TPOT, which we flag as an `anomaly` rather than fold into the steady number.

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: `k` is now defined inline in §5.7: the number of fill steps used to reach the target concurrency (issue in steps of `ceil(C/k)`). Resolving.

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: Yes — ISL/OSL are computed over the **included (windowed)** samples today, which skews them relative to the constructed dataset once the window is allowed to drop samples. §5.8 flags them as analytics-only, not validation numbers. Whether the official artifact should report them over the full issued set instead (or carry both) is deferred to the benchmark TF (§6, "ISL/OSL reporting basis").

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: Agreed, and captured as the §6 open question "First pass as warmup (task force)." The current design infers the warmup band adaptively from the data, but treating the first full pass as warmup-by-construction is a reasonable TF-set default; the accompanying band / window / bound defaults would be the TF's to fix.

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: Agreed — captured in §6 ("Per-benchmark gates and invalidation"): the CoV bounds and trend thresholds are tunable per benchmark, and whether failing a gate invalidates a submission versus reports-with-flags is explicitly a benchmark-TF decision, not fixed by this proposal.

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: Yes. The coverage-status table (§5.6) requires ≥ `warmup + 1` full super-passes to window normally; below that it's best-effort / low-confidence. For a *trustworthy* estimate the trend test needs ≥ 4 super-passes (its minimum), so in practice you want ~5–6+ post-warmup super-passes. Fewer still yields a number, but flagged `insufficient_passes`.

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: Band-adaptive warmup crop (§5.4): estimate the steady level from the median of the series' back half for a driver metric (e.g. TTFT p99), then drop leading super-passes whose value exceeds `steady × (1 + band)`. Band-based (not slope-based) so a single settling outlier can't collapse the crop, and only *high* leading values are dropped so a later upward drift is still measurable. (The reference script currently exposes a fixed `--warmup`; the adaptive crop is specced but not yet ported to it.)

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

docs/steady-state-detection.md
nv-alicheng: It means the CoV rule is evaluated at several preset `(trailing-window, bound)` settings rather than one; a window passes CoV if at least one setting certifies every gated metric. No single setting fits all metrics/workloads, so the ensemble adds robustness (concordance across settings is a corroborating guardrail). The values come from sweeps over recorded runs (defaults: bounds 0.03 / 0.05 / 0.08 loosening toward the tail, trailing windows 4–5); §6 flags that these defaults should be reviewed on a wider set before being locked.

---
_This reply was drafted by Claude (Claude Code) and posted on my behalf._

scripts/steady_state_diagnostics.md
```diff
@@ -0,0 +1,135 @@
...
+## Requirements
+
+- `uv` (the script declares its deps inline via a PEP 723 header — only `transformers`).
+- A tokenizer the run's model uses (HF id or local dir), e.g. `openai/gpt-oss-120b`,
+  `deepseek-ai/DeepSeek-R1`.
+
+## Run
+
+```bash
+uv run scripts/steady_state_diagnostics.py <run>/events.jsonl \
```
hvagadia: Is the expectation here that we will "fix" this command for a particular benchmark (across all systems)?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,534 @@
...
+  { steady_state metrics + status + anomaly }
+```
+
+### Adaptive warmup crop
+
+The ramp is removed by a data-driven crop rather than a fixed count. The steady level of
+a driver metric is estimated from the median of the per-super-pass series' back half, and
+leading super-passes whose driver value is more than a fractional `band` away from that
+level — in _either_ direction — are dropped, capped at half the run so it can never crop
+everything. The driver is **TPOT p50**: aggregated per super-pass it is a smooth, monotone
```
arekay-nv, line 236: An example would be very useful here.

docs/steady-state-detection.md
```diff
@@ -0,0 +1,534 @@
...
+drain, which loses meaning. The drain here is instead _observed_ from the TPS trend
+over time (throughput falls once the server can no longer keep the system
+saturated). A more robust definition — the drain begins once the server no longer
+has enough in-flight samples to saturate the pipeline — depends on server-side
+occupancy that the client cannot see, so it is handwaved for now. Under Poisson,
+arrival pacing throttles the pile-up, so the tail is smallest. The windowing
+response is therefore mode-specific: concurrency crops the ramp and applies the
+guarded tail-cut; offline finds its steady region from the TPS/completion-rate
+trend rather than an issue-time window (a client-side tail-cut is not meaningful);
+Poisson reuses the concurrency tooling with a single-pass super-pass.
```
arekay-nv, line 263: This seems more relevant to the tail section.

docs/steady-state-detection.md
```diff
@@ -0,0 +1,534 @@
...
+metric and percentile. The bound loosens toward the tail (a p95 is estimated from
+fewer samples per super-pass, so its sampling-noise floor is higher than a p50's).
+
+**Trend gate (mandatory), drift up vs down.** A low CoV over a window certifies
+local flatness, not that the metric has stopped moving: a slowly drifting series can
+sit locally flat while climbing overall. A trend test over the window is therefore
+applied on top. Because per-super-pass series are autocorrelated, the primary gate is
+the rank-based **Mann–Kendall test with the Hamed–Rao autocorrelation correction**
+(so serial correlation does not fake a trend); an OLS slope-vs-scatter check and a
+Newey–West (autocorrelation-consistent) slope test corroborate it. The verdict
```
arekay-nv, line 325: Please add references here similar to the White'97 for MSER below.

docs/steady-state-detection.md
```diff
@@ -0,0 +1,534 @@
...
+
+The step never hard-fails a run; every input yields a result plus a `status`.
+
+- **Coverage status.** Before windowing, classify the run by how much steady
+  signal it supports and window accordingly:
+
+```text
+  status                condition                             behavior
+  --------------------  ------------------------------------  -----------------------------
+  windowable            >= warmup + 1 full super-passes       normal warmup-cropped window
```
arekay-nv, line 502: With 1 super pass, how do we determine the trends?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,534 @@
...
+
+## 5.8 Output and consumption
+
+The steady-state metrics are reported as the **official result**; the whole-run
+(`total`) metrics remain as **supplementary** context (in `report.txt` and the
+machine-readable summary produced from the report). Each tracked metric carries
+its steady value plus its state (`Plateau` / `Drifting Up` / `Drifting Down`), and
+the run carries its coverage `status`. Reported quantities:
+
+- **The steady window** itself: its super-pass range `[start, end)` and sample count,
```
arekay-nv, line 566: Can we include some way of correlating them with the log entries such as request ids? The super-pass range is useful if we can trace it back to the log with ease.

docs/steady-state-detection.md
```diff
@@ -0,0 +1,534 @@
...
+  so a reader can see where in the run it was drawn from.
+- **TPS**, reported two ways because they answer different questions: **per-user TPS**
+  = `1 / mean(TPOT)` (output tokens/s/user, the interactivity number) and **system
+  TPS** = total output tokens in the window divided by its wall-clock span
+  (aggregate throughput). Each carries a confidence interval computed by
```
arekay-nv, line 571: Is the wall-clock span of the window determined by request-issue as well. So the window ends when the last request in that window is issued?

docs/steady-state-detection.md
```diff
@@ -0,0 +1,534 @@
...
+  the per-super-pass autocorrelation rather than assuming independent samples.
+- **TTFT and TPOT** percentiles (p50/p90/p95/p99) and histograms, over the pooled raw
+  samples of the steady window.
+- **Anomaly.** When the level-shift detector fires (§5.5), an `anomaly` block records
+  the change-point super-pass, the shift magnitude and direction, and every detected
+  plateau — so a degradation toward the end is surfaced next to the (first-plateau)
+  steady result rather than hidden by it.
+- **QPS is dropped** for text/token-based LLM workloads — a request is not a unit
+  of work when output length varies widely, so QPS is a legacy metric of little
+  meaning; it is retained only for token-free, uniform-work loads.
```
arekay-nv, line 582: 👍

docs/steady-state-detection.md
```diff
@@ -0,0 +1,534 @@
...
+  task force.
+- **First pass as warmup (task force).** Whether to treat the first full dataset
+  pass (or first super-pass) as warmup by construction — rather than inferring the
+  warmup band from the data — and the exact band/window/bound defaults that would
+  accompany such a rule are deferred to the benchmark task force.
+- **Per-benchmark gates and invalidation (task force).** Whether the steady-state
+  gates (CoV bounds, trend thresholds) are tuned per benchmark, and whether a run
+  that fails them is declared _invalid_ versus reported-with-flags (see
+  "No-steady-state runs" above), is a benchmark-task-force decision, not fixed by
+  this proposal.
```
arekay-nv, line 635: Based on your analysis can you share some sensible defaults for any of the benchmarks so there is a starting point.

scripts/steady_state_diagnostics.py
```diff
@@ -0,0 +1,1393 @@
...
+
+
+def run(
+    events_path: str,
+    superpass_size: int,
+    count_tokens: Callable[[list[str]], list[int]],
+    window_sizes: Sequence[int] = (4, 5),
+    warmup: int | str = "auto",
+    cov_bounds: Sequence[float] = (0.03, 0.05, 0.08),
+    alpha: float = 0.05,
```
arekay-nv, line 1598: This doesn't seem to be consumed downstream.

docs/steady-state-detection.html
```diff
@@ -0,0 +1,695 @@
...
+<a href="#minduration">9 · Minimum-duration gate</a>
+<a href="#e2e">10 · End-to-end worked example</a>
+<a href="#validation">11 · Validation sweep</a>
+<a href="#output">12 · Output &amp; flags reference</a>
+</div>
+
+<h2 id="vocab">0 · Vocabulary</h2>
+<table>
+<tr><th>Term</th><th>Meaning</th></tr>
+<tr><td><b>super-pass</b></td><td>A contiguous block of <code>dataset_size</code> samples in <em>issue order</em>
```
arekay-nv: Preferably window size to differentiate from the full dataset.

docs/steady-state-detection.html
```diff
@@ -0,0 +1,695 @@
...
+(one full pass over the dataset). The atomic unit of the analysis — every metric becomes a
+per-super-pass series. Think of it as one "batch" for batch-means statistics.</td></tr>
+<tr><td><b>trajectory</b></td><td>The per-super-pass series of one metric percentile,
```
arekay-nv: Isn't a trajectory a unit of execution?

docs/steady-state-detection.html
```diff
@@ -0,0 +1,695 @@
...
+<code>tpot_p50</code>, <code>tpot_p95</code> (decode-rate steadiness). <b>TTFT is not gated</b> — its
+high-concurrency tail variance is structural (prefill/ISL skew + queue), so it's tracked as a diagnostic
+and raises a soft drift warning, but never hard-fails a window. <code>p99</code>, sample latency, warm-turn
+TTFT are diagnostic too.</td></tr>
+<tr><td><b>plateau</b></td><td>A maximal run of super-passes that is trend-flat and low-variance for
+every gated metric. A run can contain several (a staircase).</td></tr>
+</table>
+
+<h2 id="pipe">1 · The pipeline</h2>
+<p>One pass over <code>events.jsonl</code>, then a fixed sequence of transforms. Each stage below links
```
arekay-nv: Clarify that this is post benchmark execution when an events.jsonl file has been generated.

docs/steady-state-detection.html
```diff
@@ -0,0 +1,695 @@
...
+is required. Token counts use plain tokenization; the live aggregator uses the chat-template path, so
+absolute TPOT ms can differ for reasoning models — but CoV and trend tests are scale-invariant, so the
+steady/drift <em>verdicts</em> are unaffected.</p>
+
+<h2 id="warmup">4 · Adaptive warmup crop</h2>
+<p>The start of every run is a ramp: connections opening, caches filling, autoscale settling. A fixed
+crop either wastes good data or leaves ramp contamination. Instead the crop is <em>data-driven</em>:</p>
+<ol>
+<li>Take the driver trajectory (<code>tpot_p50</code> — TPOT ramps <em>up</em> to steady, monotone and clean).</li>
+<li>Estimate the steady level as the median of the <em>back half</em> of the run.</li>
```
arekay-nv, line 182: `back half` -> `second half`

docs/steady-state-detection.html
```diff
@@ -0,0 +1,695 @@
...
+<circle cx="700.0" cy="68.0" r="3.2" fill="#1f6feb"/>
+</svg><figcaption><b>Worked example.</b> Back-half median ≈ 5.0 ms → band [4.75, 5.25].
+Super-passes 0,1,2 (3.6, 4.1, 4.5 ms) are below the band; super-pass 3 (4.85) is the first inside it, so
+<b>warmup = 3</b> and the reported window starts at index 3. All later indices are <em>post-warmup relative</em>.</figcaption></figure>
+
+<h2 id="gate">5 · The admissibility gate (trend + CoV)</h2>
+<div class="note"><b>Gate on TPOT only.</b> Admissibility uses <code>tpot_p50</code> and
+<code>tpot_p95</code> — decode-rate steadiness. <b>TTFT is not a hard gate:</b> at high concurrency its
+tail is dominated by prefill time (tracking dataset input-length skew) + queue wait, so its variance is
+<em>structural</em> (<code>ttft_p95</code> CoV floors at 0.3–0.9 regardless of super-pass size) rather than
```
arekay-nv: Is this empirical? Can you add some context?

docs/steady-state-detection.html
```diff
@@ -0,0 +1,695 @@
...
+<code>z = (S∓1)/√Var(S)</code>, and a two-sided p-value. Verdict is <code>up</code>/<code>down</code>
+only when significant at α = 0.05, else <code>steady</code>.</p>
+
+<div class="two">
+<div class="card"><b>Flat window</b> — <code>[5.0, 5.02, 4.98, 5.01, 4.99, 5.0]</code><br>
+<span class="small">Rising and falling pairs roughly cancel → <code>S ≈ 0</code> → <code>z ≈ 0</code> →
+<span class="tag" style="background:var(--ok)">steady</span></span></div>
+<div class="card"><b>Rising window</b> — <code>[4.0, 4.3, 4.6, 4.9, 5.2]</code><br>
+<span class="small">Every one of the 10 pairs is positive → <code>S = 10</code> (max) → large <code>z</code> →
+<span class="tag" style="background:var(--bad)">up</span></span></div>
```
arekay-nv, line 278: Can you expand the score calculation.

arekay-nv (approved):

scripts/steady_state_diagnostics.py
nv-alicheng: Correct — `alpha` was echoed into the JSON but never threaded into the gate (the trend algorithms carry their own significance default of 0.05), so it did nothing downstream. Removed the dead `run()` parameter, the `--alpha` CLI flag, and the `result["alpha"]` field. Fixed in f1445873.

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

tests/unit/scripts/test_synth_events.py
```diff
@@ -0,0 +1,249 @@
...
+        "sample.recv_first",
+        "sample.complete",
+    } <= types
+
+
+def test_issued_events_are_chronological(tmp_path):
+    out, _, _ = _generate(tmp_path)
+    issued_ts = [
+        json.loads(line)["timestamp_ns"]
+        for line in open(out)
```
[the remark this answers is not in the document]
nv-alicheng: Fixed — wrapped the read in a `with open(out) as fh:` context manager so the handle always closes even if JSON parsing raises. Fixed in f1445873.

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

docs/steady-state-detection.md
nv-alicheng: You cannot - I'm assuming this question is geared towards low-concurrency runs where we only run 1 dataset pass or a partial pass. Those can be handled one of 2 ways:
1. As is - steady state will report 'not enough superpasses' and report will default to using the whole run. This is fine as at very low concurrency, ramp up/down effects are much less impactful 
2. Enforce that with any run, we always subdivide into some M=40~ish  superpasses of equal size. I would recommend avoiding this as due to dataset distribution, this will make the metrics quite choppy since the superpass size will be small.

Recommended (and current) is (1)

docs/steady-state-detection.md
nv-alicheng: I had this thought too - will add in a follow-up, since this might be useful for analytics.

docs/steady-state-detection.md
nv-alicheng: Yes - this is similar to how other benchmarks handle it (they stop measuring once the last sample is issued).

Simple example, superpass size = 2:
- Sample A issued at t = 123 (Window start)
- Sample A response chunk 1 at t = 125
- Sample B issued at t = 126 (Window stop)
- Sample A response chunk 2 at t = 126.01

Here the wall clock is 3 units, and the only metric available would a single TTFT event for sample A. Sample B will have no metrics, and the TPOT event for sample A is dropped, since (though barely) it was received after sample B's issue timestamp.

The TPOT event for sample A would get included in the *following* superpass.

docs/steady-state-detection.md
nv-alicheng: Added a worked example to the Adaptive warmup crop — a small super-pass series of `TPOT p50` values, the ±5% band, and which super-passes are cropped (super-passes 0–2 below the band; super-pass 3 first inside → `warmup = 3`). Also changed "back half" → "second half". Added in bc33d297.

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

docs/steady-state-detection.md
nv-alicheng: Agreed — moved the mode-specific windowing / tail-formation paragraph out of the adaptive-warmup subsection into §5.4 (Guarded drain-tail cut). Added in bc33d297.

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

docs/steady-state-detection.md
nv-alicheng: Added inline citations (Mann–Kendall, Hamed–Rao, Newey–West, Pettitt) in the trend-gate section and a new §7 References with resolving DOIs (bare `<…>` autolinks so the parenthesised Hamed–Rao DOI isn't truncated) plus the MLPerf Inference arXiv link (1911.02549). Added in bc33d297.

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

docs/steady-state-detection.html
nv-alicheng: Done — reworded the `super-pass` vocabulary entry to describe it as **one analysis window** (a contiguous issue-order block of `dataset_size` samples), rather than "one full pass over the dataset", so the window-vs-dataset distinction is explicit.

Added in `b2215a8c`.

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

docs/steady-state-detection.html
nv-alicheng: Agreed — "trajectory" collides with the agentic unit-of-execution meaning. Renamed the concept to **metric series** throughout the explainer (the per-super-pass series of one metric percentile, e.g. `tpot_p50`); "trajectory" is now reserved for the agentic multi-turn sense.

Added in `b2215a8c`.

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

docs/steady-state-detection.html
nv-alicheng: Added — the §1 intro now states this is a **post-run, offline (cold-path)** step that runs *after* the benchmark finishes, reading the recorded `events.jsonl`, and never touches the hot path.

Added in `b2215a8c`.

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

docs/steady-state-detection.html
nv-alicheng: Yes — added that context. The TPOT-only gate is empirical: on `events.jsonl` logs from **GPT-OSS** and **DeepSeek-R1** across a range of concurrencies, the TTFT tail (`ttft_p90`) CoV floors at 0.3–0.9 regardless of super-pass grain while `tpot` CoV is ~0.01–0.03, so gating on the tail fragmented runs whose decode was genuinely steady, while gating on TPOT alone recovers them. The gate note now says so explicitly, and the design-doc §5.3 wording matches.

Added in `b2215a8c`.

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

docs/steady-state-detection.html
nv-alicheng: Expanded — added a worked Mann–Kendall block for the rising-window example (`[4.0, 4.3, 4.6, 4.9, 5.2]`): all 10 pairs rising → `S = 10`; `Var(S) = n(n−1)(2n+5)/18 = 16.67`; continuity-corrected `z = (S−1)/√Var = 9/4.08 = 2.20`; two-sided `p = 0.028 < 0.05 → up`. Plus the flat-window contrast (`S ≈ 0 → z ≈ 0 → steady`) and a one-line note on the Hamed–Rao variance inflation under autocorrelation.

Added in `b2215a8c`.

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

docs/steady-state-detection.md
nv-alicheng: Data-backed starting points from an internal study over GPT-OSS and DeepSeek-R1 logs at a range of concurrencies:

- **super-pass (window) size:** one super-pass = one issue-order window of `dataset_size` samples. Verdicts are robust across a wide range of super-pass sizes, so this is not a sensitive knob.
- **gated metric:** `tpot_p50` + `tpot_p90` must Plateau. TTFT is a **diagnostic**, not a gate — its tail is dataset-skew-driven (hairball), so gating it rejects genuinely decode-steady runs.
- **CoV bounds:** `{p50: 0.03, p90: 0.05}` (looser toward the tail); trend gate mandatory (`mk_hamed_rao`) with an effect-size floor `|rel_drift| ≥ 0.05` so a negligible drift doesn't fragment a steady run.
- **min-duration:** `max(precision = k*·τ_sp, relaxation = 5·p90_latency, floor = 600 s)`, over the offered span. The 600 s floor covers clean high-throughput points; the only thing that moves it per-benchmark is **output length** via the relaxation term (outlier requests with extremely long sample latencies lift it well past the floor).

These are starting points; final per-benchmark gate values and invalidation-vs-report-with-flags are a TF decision (§6).

— _Drafted by Claude (Claude Code), posted by @nv-alicheng._

nv-alicheng: We can go with the legacy of 10m minimum to be considered a valid steady state window (i.e. the window's duration is >= 10m) but ideally 15/20min would be better.

PR #452, opened 2026-08-19
fix(swebench-service): let srun find its config and honour proxy policy

leopck:
The Pyxis runtime builds each container step's environment from an explicit allow-list, and two variables that srun and enroot genuinely need were missing. Neither failure is visible in a unit test, because both live in the environment handed to a subprocess we mock.

SLURM_CONF. Without it the child srun falls back to /etc/slurm/slurm.conf. On a configless site that file does not exist and srun aborts with "Could not establish a configuration source"; on a multi-cluster site it exists but is a different cluster's file whose plugins are not installed locally, and srun aborts with "failed to initialize cli_filter plugin". Either way every step dies before a container is created, and the run reports only the generic "Pyxis infrastructure failure before the command completed". The remaining SLURM_* variables stay out of the allow-list deliberately: inheriting SLURM_JOB_ID / SLURM_STEP_ID is exactly what breaks a nested srun, which is why the allow-list exists.

Proxy policy. enroot performs the registry pull inside the step, so it needs the same proxy configuration as the caller. A site that pins a container-cache proxy system-wide will 403 the CONNECT for any registry outside that cache's allow-list, and every per-instance image import fails with "curl: (56) CONNECT tunnel failed, response 403" -- including the SWE-bench task images this runtime is built to pull.

Verified on a GB200 cluster whose enroot pins a container cache: before the change no sweb.eval.arm64 image could be imported from any node; after it, the image imports and the container starts.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

### DIFF
--- a/src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
+++ b/src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
@@ -30,22 +30,16 @@ _SAFE_SRUN_ENV = (
     "TMPDIR",
     "XDG_RUNTIME_DIR",
     # Proxy policy must reach enroot, which performs the registry pull inside
-    # the step. Clusters that pin a container-cache proxy system-wide 403 any
-    # registry outside its allow-list, and without no_proxy every per-instance
-    # image import fails with "CONNECT tunnel failed, response 403".
+    # the step.
+    "all_proxy",
     "http_proxy",
     "https_proxy",
     "no_proxy",
+    "ALL_PROXY",
     "HTTP_PROXY",
     "HTTPS_PROXY",
     "NO_PROXY",
-    # srun locates its own configuration through SLURM_CONF. Dropping it makes
-    # the child fall back to /etc/slurm/slurm.conf, which on a configless or
-    # multi-cluster site is either absent ("Could not establish a configuration
-    # source") or a different file whose plugins are not installed
-    # ("failed to initialize cli_filter plugin"). Every step then fails before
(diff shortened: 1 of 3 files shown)


tianmu-li (approved):
Comments can be trimmed a bit.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
```diff
@@ -29,6 +29,24 @@
...
+    # Proxy policy must reach enroot, which performs the registry pull inside
+    # the step. Clusters that pin a container-cache proxy system-wide 403 any
+    # registry outside its allow-list, and without no_proxy every per-instance
+    # image import fails with "CONNECT tunnel failed, response 403".
+    "http_proxy",
+    "https_proxy",
+    "no_proxy",
+    "HTTP_PROXY",
+    "HTTPS_PROXY",
+    "NO_PROXY",
```
arekay-nv, line 41: minor : missing `all_proxy`/`ALL_PROXY` which can be set in some envs.

arekay-nv (approved):
One aspect to think about is not passing the variables to the SWEBench environment which can further harden it. 
Also, can you add some tests to ensure that the env-vars are being passed as expected. And if we want to isolate the SWEBench environment, the corresponding tests there too.

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
leopck: oh thanks, forgot about them

PR #454, opened 2026-08-19
fix(benchmark): an accuracy run that produces no number must fail

leopck:
An accuracy run could complete every unit of work, exit 0, and report `N/A`. `finalize_benchmark()` scored, wrote `accuracy_results.json` with `score: null`, printed the summary, and returned -- so `main.py` exited 0 and every wrapper downstream read the run as a pass.

This is not hypothetical. A distributed SWE-bench run drove all 20 of its units to terminal records; 17 were abandoned to infrastructure failures, the merge gate correctly refused, `score()` returned None -- and the sbatch wrapper wrote `disposition=run completed (driver rc=0, results=20/20)` over a run with no accuracy number at all. rc=0, all work "done", no number is the failure shape that costs whole GPU allocations, so make it loud where it originates instead of asking each caller to notice.

`_require_accuracy_numbers()` runs last, after every artifact is on disk, so the failure never costs the evidence needed to diagnose it. A real number flagged `complete=False` (a partial headline) still passes: the number exists and the entry already says it is partial. A PERF-mode run owes nothing for an externally-scored dataset it never dispatched.

Also count what was evaluated, not what was loaded, in the accuracy-only summary line: SWE-bench Verified loads all 500 rows and scores `num_instances` of them, so the run above printed "500 samples evaluated" directly beneath its own "unit=200" headline.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

### DIFF
--- a/src/inference_endpoint/commands/benchmark/execute.py
+++ b/src/inference_endpoint/commands/benchmark/execute.py
@@ -1088,11 +1088,20 @@ def _summarize_and_log_metrics(
 
     logger.info(f"Completed in {perf_elapsed:.1f}s")
     if ctx.accuracy_only:
-        acc_total = sum(
-            ds.dataset.num_samples() * ds.num_repeats
-            for ds in ctx.eval_configs
-            if ds.dataset_type == DatasetType.ACCURACY
-        )
+        # Count what was actually evaluated, not what was loaded. An external
+        # scorer (SKIP_ENDPOINT_PHASE) evaluates its own clamped subset --
+        # SWE-bench Verified loads all 500 rows but scores `num_instances` of
+        # them -- and reporting the loaded size instead contradicts the
+        # per-dataset `unit=` line in the same summary.
+        acc_total = 0
+        for ec in ctx.eval_configs:
+            if ec.dataset_type != DatasetType.ACCURACY:
+                continue
+            external = effective_external_sample_count(ec)
+            if external is not None:
+                acc_total += external
+            else:
+                acc_total += ec.dataset.num_samples() * ec.num_repeats
         logger.info(f"Accuracy-only: {acc_total} samples evaluated")
     else:
         logger.info(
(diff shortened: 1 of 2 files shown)


src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1208,6 +1217,53 @@ def finalize_benchmark(ctx: BenchmarkContext, bench: BenchmarkResult) -> None:
...
+        and not (ctx.test_mode == TestMode.PERF and ec.scorer.SKIP_ENDPOINT_PHASE)
+    }
+    if not expected:
+        return
+
+    scored = {
+        entry["dataset_name"]
+        for entry in accuracy_scores
+        if entry.get("dataset_type") == DatasetType.ACCURACY.value
+        and entry.get("score") is not None
```
arekay-nv, line 1256: This is checking for `None` which is not the same as being a number - suggest to either harden the check or make the distinction clear in the method description/documentation.

arekay-nv (approved):
Thanks!

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -1128,11 +1128,20 @@ def _summarize_and_log_metrics(
...
+        # Count what was actually evaluated, not what was loaded. An external
+        # scorer (SKIP_ENDPOINT_PHASE) evaluates its own clamped subset --
+        # SWE-bench Verified loads all 500 rows but scores `num_instances` of
+        # them -- and reporting the loaded size instead contradicts the
+        # per-dataset `unit=` line in the same summary.
+        acc_total = 0
+        for ec in ctx.eval_configs:
+            if ec.dataset_type != DatasetType.ACCURACY:
+                continue
+            external = effective_external_sample_count(ec)
```
roborluo, line 1140: [P2]:Only to make this consistant with the notes, suggst to add 
```
if ec.scorer.SKIP_ENDPOINT_PHASE:
```
before external = effective_external_sample_count(ec).

roborluo (approved):

tests/unit/commands/test_benchmark.py
```diff
@@ -2563,6 +2602,129 @@ def _make_report(state: str) -> Report:
...
+        finalize_benchmark(ctx, _make_benchmark_result(tmp_path))
+
+        results = json.loads(
+            (tmp_path / "accuracy" / "accuracy_results.json").read_text()
+        )
+        assert results["accuracy_scores"][0]["score"] == 0.25
+        assert results["accuracy_scores"][0]["complete"] is False
+
+    @pytest.mark.unit
+    def test_perf_only_run_owes_no_accuracy_number(self, tmp_path):
```
roborluo, line 2693: Sugguest to delete this test, this will never happen since we already have 
```
if not (test_mode == TestMode.PERF and scorer_cls.SKIP_ENDPOINT_PHASE):
    ...
    eval_configs.append(...)
```
Maybe you want to test such `scorer_cls` will not be added into `eval_configs`?

PR #457, opened 2026-08-19
fix(lcb-service): pin base image to Python 3.11 to match LiveCodeBench

liayan:
LiveCodeBench requires Python >= 3.10, and its README uses `uv venv --python 3.11`. Switch both build/runtime stages from dhi.io/python:3.14-debian13(-sfw-dev) to dhi.io/python:3.11-debian13(-sfw-dev) to match.

This is related to https://github.com/mlcommons/endpoints/pull/433
https://github.com/mlcommons/endpoints/pull/433#issuecomment-5293679069

Python 3.14 changed the Linux multiprocessing default to forkserver. With the shipped Python 3.14 service image, grading child processes died during startup, leaving evaluation stuck at 0/N.

After discussion, the current conclusion is to revert the base image to Python 3.11. An explicit spawn solution is still worth investigating, but it needs additional performance evaluation before we consider adopting it.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass

### DIFF
--- a/src/inference_endpoint/evaluation/livecodebench/lcb_serve.dockerfile
+++ b/src/inference_endpoint/evaluation/livecodebench/lcb_serve.dockerfile
@@ -1,7 +1,8 @@
 # syntax=docker/dockerfile:1
 
 ## -----------------------------------------------------
-FROM dhi.io/python:3.14-debian13-sfw-dev AS build-stage
+# Mirrors LiveCodeBench's own Python version (README: `uv venv --python 3.11`).
+FROM dhi.io/python:3.11-debian13-sfw-dev AS build-stage
 
 
 ENV PYTHONDONTWRITEBYTECODE=1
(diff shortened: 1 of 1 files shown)


liayan:
Discussed offline with @arekay-nv — we're moving the explicit spawn/fork context pinning to a separate follow-up PR with more performance data, to avoid a perf regression on the spawn side. Everything else through c14c3c2 (SystemExit handling, started_flag attribution, Manager with-block, type annotations, logging levels) stays as-is here.

Verified in a python:3.11-slim container (Python < 3.14, defaults to fork) — all 11 regression tests pass.

Also opened #457 to pin the base image back to Python 3.11, mirroring LiveCodeBench's own supported version. Could you take a look and merge that one too — it's still blocking lcb-service on aarch64 without the new spawn soluton/fork solution.

arekay-nv (approved):
Thanks!

viraatc (approved):

PR #458, opened 2026-08-20
Distributed SWE bench by sharding SWE tasks across cpus for faster runtime

leopck:
## What does this PR do?

Depends on https://github.com/mlcommons/endpoints/pull/452, https://github.com/mlcommons/endpoints/pull/453, https://github.com/mlcommons/endpoints/pull/454, https://github.com/mlcommons/endpoints/pull/456 to be merged first

Distributed SWE shards the tasks across multiple CPUs into one server, this is to tackle the limitation on SWE bench itself. Namely:

1. SWE bench spawns multiple containers depending on the number of workers, however, the number of containers are limited to the host's memory/RAM, if the worker count is set too high, the container will eat up all host memory and OOM causing the SWE to crash
2. Because of low worker count/low concurrency, we are observing low utilization on the inference engine side as well.
3. Some SWE bench containers fluctuates in memory depending on what it's compiling or building (60GB ~ 90GB)
4. Docker/Pyxis has issues on disk contention if there's too many different container spawning at the same time, but there's no issues if we are spawning multiple of the same containers
5. Pyxis has limitation to 16 workers due to slurm limitation rather that memory limitation

This PR tackles these problems by distributing the containers across multiple CPU nodes, this reduces the memory pressure applied to the host memory and enabling the containers to spawn more reliably. As this depends on the number of CPU nodes you have available, this can greatly increase your worker count enabling extremely fast runtime (3x ~ 6x time saving). 

This PR has been tested on both x86 (Podman/Docker) and ARM (Pyxis):

```
NSC  x86 + docker/podman   71.00% 
OCI  ARM + pyxis           72.50%  
reference                  70.58% 
```

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor/cleanup


tianmu-li:
I was using a system with 64GiB host memory for accuracy testing running 30 workers and didn't run into host memory issue. Also didn't recall seeing multi-hundred GiB memory usage with 200-worker runs. Could you double-check if this is really needed?

leopck:
> I was using a system with 64GiB host memory for accuracy testing running 30 workers and didn't run into host memory issue. Also didn't recall seeing multi-hundred GiB memory usage with 200-worker runs. Could you double-check if this is really needed?

@tianmu-li I think the maximum that I'm able to run this on is around 20 workers and I have around 1TB of host RAM, and the containers I did manually observe the host RAM increasing to 90GB at some points.

Following this, this PR would help us to increase to 200 workers so that we can run SWE bench faster, which is the whole purpose of this PR, because even in your case, you are only limited to 30 workers am I right? With 200 workers we should be able to complete this SWE bench on a faster runtime. Open to discuss more if this is the right direction.

tianmu-li:
@leopck From my observation, 1TiB ram can run 200 concurrent workers with plenty of headroom. I'm not sure why you face RAM pressure unless you are using ram-backed storage, though even then 1TiB is sufficient for all images + running state +ram requirement.

leopck:
@tianmu-li which runtime engine are you using? I'm assuming Docker? we are using root-less podman and pyxis runtime, we might be seeing different behaviours on our end

tianmu-li:
@leopck I'm using docker. Currently redoing a run with 30 workers and barely seeing memory usage over 10GiB. Even when adding storage used it's less than 100GiB. I'll be surprised if podman/pyxis vs docker makes such a large difference.

leopck:
> @leopck I'm using docker. Currently redoing a run with 30 workers and barely seeing memory usage over 10GiB. Even when adding storage used it's less than 100GiB. I'll be surprised if podman/pyxis vs docker makes such a large difference.

@tianmu-li let me do a thorough check on the rootless podman where I am observing this memory pressure issue.

Putting the memory issue aside, we still need this for pyxis environment as we are limited to 16 workers due to slurm limitation on this. And on memory limited machines, this distributed SWE is still helpful to increase the worker count.

leopck:
moved to gh stack

PR #409, opened 2026-07-13
feat(config): consolidate all time knobs into settings.timeouts; make --timeout a real run watchdog

viraatc:
## Changes

- Consolidates benchmark deadlines and drain budgets under `settings.timeouts`; workload issue-duration bounds remain under `settings.runtime`, and endpoint-worker lifecycle waits remain under `settings.client`.
- Makes `--timeout` an alias for `settings.timeouts.run_timeout_s`. The deadline is captured before synchronous setup, checked when setup returns, and remains active through service startup, benchmark phases, response/metrics drains, and worker shutdown.
- Treats watchdog, metrics-drain, transport, and user interruption as non-success outcomes with honest interrupted/incomplete artifacts and non-zero exits.
- Uses one Ctrl-C contract: the first signal requests bounded graceful shutdown; a later signal hard-stops the benchmark process group without report/salvage work.
- Prevents event logger, metrics tokenizer, and endpoint-worker subprocess bootstrap tracebacks by making the parent own SIGINT; the event logger flushes writers on SIGTERM.
- Reserves `DRAIN_RESERVED_CPUS` during metrics tokenization drain so the host remains responsive.
- Deletes the old top-level timeout/drain paths and compatibility aliases. Timeout values are strictly positive; `null` is the only unlimited/off value where supported.

## Configuration

```yaml
settings:
  timeouts:
    run_timeout_s: null
    interrupted_teardown_grace_s: 30.0
    service_ready_timeout_s: 30.0
    warmup_drain_timeout_s: 240.0
    performance_drain_timeout_s: null
    accuracy_drain_timeout_s: null
    metrics_drain_timeout_s: null
```

`run_timeout_s` excludes accuracy scoring/finalization, tracked by [#464](https://github.com/mlcommons/endpoints/issues/464). Each audit subrun currently receives its own setup-inclusive benchmark deadline; an overall audit-sequence deadline is tracked by [#466](https://github.com/mlcommons/endpoints/issues/466).

## Follow-ups

- [#449](https://github.com/mlcommons/endpoints/issues/449) — promote warmup to a first-class phase type with per-phase configuration.
- [#459](https://github.com/mlcommons/endpoints/issues/459) — consolidate run-outcome state into one abort/outcome model.
- [#464](https://github.com/mlcommons/endpoints/issues/464) — add a deadline for accuracy scoring and finalization.
- [#465](https://github.com/mlcommons/endpoints/issues/465) — parameterize run-timeout coverage across lifecycle stages.
[...]

PR #462, opened 2026-08-22
fix: fail fast on endpoint response stalls

roborluo:
## Summary

- add an opt-in no-progress deadline for in-flight endpoint requests
- expose it as `settings.timeouts.no_progress_timeout_s` plus a `--no-progress-timeout` CLI alias
- arm the guard only while work is in flight; retire it as soon as the phase drains

## When this is useful

Use this for automated runs where a request can be accepted but the endpoint then becomes silent — for example, a TensorRT-LLM disaggregated executor or KV-transfer stall that never reaches the normal terminal-error path. It is engine-agnostic: it catches the same client-visible silent failure through vLLM, a frontend, or transport. Without it the benchmark blocks until its outer wall-time limit, because the phase drain waits on responses that never arrive and `drain_timeout` defaults to unlimited.

Disabled by default. It starts after work is issued, runs only while requests are in flight, and resets on an observed stream chunk or final result. It does not diagnose or restart the backend — it makes the benchmark fail with a clear error. For non-streaming endpoints, configure it above the full expected request latency. For TensorRT-LLM disaggregated serving the documented starting value is 300 s, matching the executor `hang_detection_timeout`.

## Changes since the last review

Rebased onto latest main and moved the setting into the `Timeouts` model added by #409 (@viraatc).

| Review point | Change |
| --- | --- |
| **[P1]** `completed_uuids` retained every completed UUID | Only `register_skipped` adds tombstones now, so it no longer grows O(total requests) alongside `uuid_to_index` |
| **[P2]** disabled path did work per response | Gated on the feature being enabled, and the receiver reuses the arrival timestamp it already takes instead of reading the clock twice |
| **[P2]** task + timer per 0→1 transition | Single self-rearming `loop.call_later` TimerHandle — no task, no `Event`, no per-iteration `wait_for` |
| **[P2]** full-timeout sleep could detect silence ~2x late | Remaining deadline is computed from the last progress stamp |
| `cohort` wording | Removed from the docs and from the local variable in `issue()` |
| help/description drift | Schema description is the short form; `docs/config/DESIGN.md` holds the tuning detail |

**Measured cost when disabled** (A/B against the base commit): **+3.5 ns per response, +9.4 ns per request**. With `stream_all_chunks: false` (the default) the main process sees 2 messages per request, not one per token.

**Measured cost of the arming change** under the production eager task factory: arm+cancel per request went from **2205 ns to 535 ns (4.1x)**.

### Also fixed while here

A receiver transport error could overwrite an earlier `NoProgressError`. A stalled endpoint often drops its connection too, so the real diagnosis was being masked by a generic "receiver failed" message. First error now wins, with a regression test.

### Not adopted

The suggestion to use one phase/session-lifetime watchdog driven by an activity event. The guard is still armed on the 0→1 in-flight transition and retired on drain — now with a `TimerHandle` rather than a task. This keeps the timer heap clear once a phase drains, and measured 4.1x cheaper than the reviewed version. Happy to switch to a resident watchdog if that is preferred.

## Validation

- `pytest tests/unit/config/test_schema.py tests/unit/commands/test_benchmark.py tests/unit/load_generator/test_async_session.py` — **425 passed**
- `pre-commit run --all-files` — all hooks pass except mypy, which reports 3 pre-existing `os.sched_*affinity` errors on macOS in `cpu_affinity.py` / `token_metrics.py`; neither file is touched by this PR and the symbols exist on Linux
- `python scripts/regenerate_templates.py --check`
- **AGA disaggregated held-response canary, job 578946**: failed as intended with `Endpoint made no response progress for 10.0s with 1 request(s) in flight`, **10.002 s** after the phase started against a 10 s deadline. The 10 s value is fault-injection coverage, not the deployment recommendation.

Timer lifecycle is covered in both directions: the perf issue cap (`stop_current_phase`) deliberately leaves the guard armed so a stall during drain is still caught, while drain completion, `stop()`, a phase change, and session teardown each retire it.


docs/load_generator/DESIGN.md
```diff
@@ -173,7 +174,7 @@ class BenchmarkSession:
 **`run(phases)`** lifecycle:
 
 1. Publish `SessionEventType.STARTED`
-2. Start receiver coroutine (`_receive_responses`)
+2. Start receiver coroutine (`_receive_responses`). When the no-progress timeout is set, each active in-flight cohort arms a liveness watchdog; it exits as soon as the cohort drains. The watchdog fails the session only when an in-flight request has no streamed chunk or final result for the configured interval.
```
arekay-nv: Lets not use `cohort` here.

src/inference_endpoint/config/schema.py
```diff
@@ -648,6 +648,26 @@ def _parse_duration_suffix(cls, v: object) -> object:
...
+    ] = Field(
+        None,
+        gt=0,
+        description=(
+            "Fail a run when requests are in flight but no "
+            "response chunk or completion arrives for this many seconds. Disabled "
+            "by default; set above the longest expected interval between response "
+            "progress (full request latency for non-streaming endpoints) and, for "
+            "TensorRT-LLM disaggregated serving, match hang_detection_timeout."
+        ),
```
arekay-nv: Lets keep a single copy in the help/description so there isn't drift over time.

src/inference_endpoint/load_generator/session.py
```diff
@@ -651,6 +773,7 @@ def _handle_response(self, resp: QueryResult | StreamChunk) -> None:
                 )
 
             if phase_issuer is not None and query_id in phase_issuer.uuid_to_index:
+                phase_issuer.completed_uuids.add(query_id)
```
arekay-nv: [P1] Avoid retaining every normally completed UUID here. `completed_uuids` lives for the entire phase alongside `uuid_to_index`, so this adds a second O(total requests) hash index. At the repository's 50k+ QPS target, long phases can accumulate millions of decoded UUID strings and multi-GB incremental memory pressure, eventually affecting cache locality or causing OOM. Keep tombstones only for exceptional/synthetic completions, or track active IDs and remove them on completion.

src/inference_endpoint/load_generator/session.py
```diff
@@ -563,20 +606,99 @@ async def _drain_inflight(
...
+                    # Cancel a strategy blocked awaiting a semaphore that will
+                    # never be released.
+                    if self._strategy_task and not self._strategy_task.done():
+                        self._strategy_task.cancel()
+                    break
+                self._handle_response(resp)
+                # This is a session-level liveness guard: any response proves
+                # the endpoint/transport is making progress. In particular,
+                # warmup responses may arrive after the performance phase starts.
+                self._record_response_activity()
```
arekay-nv: [P2] Please keep the default-disabled path inert. This unconditionally calls `monotonic_ns()` and writes `_last_response_progress_ns` for every chunk and final response even when `no_progress_timeout_s` is `None`; streaming frame rate can substantially exceed request QPS. Gate activity tracking on the feature being enabled and reuse the timestamp already taken while handling the response.

src/inference_endpoint/load_generator/session.py
```diff
@@ -563,20 +606,99 @@ async def _drain_inflight(
...
+    def _record_response_activity(self) -> None:
+        self._last_response_progress_ns = time.monotonic_ns()
+
+    def _on_inflight_started(self) -> None:
+        """Start a no-progress watchdog for a newly active request cohort."""
+        if self._no_progress_timeout_s is None:
+            return
+        if self._progress_watchdog_task and not self._progress_watchdog_task.done():
+            self._progress_watchdog_task.cancel()
+        self._progress_watchdog_task = asyncio.create_task(self._watch_no_progress())
```
arekay-nv: [P2] This creates a new asyncio task and timer for every 0-to-1 in-flight transition, then cancels it on 1-to-0. Concurrency-one and sparse workloads therefore allocate/cancel a watchdog task per request, directly adding scheduler/timer-heap work to the guarded hot path. Prefer one phase/session-lifetime watchdog driven by an activity event and updated deadline.

src/inference_endpoint/load_generator/session.py
```diff
@@ -563,20 +606,99 @@ async def _drain_inflight(
...
-                logger.warning("Issuer recv() returned None — transport closed")
-                self._stop_requested = True
-                self._drain_event.set()  # Unblock _drain_inflight
-                # Cancel the strategy task if it's blocked (e.g., ConcurrencyStrategy
-                # awaiting sem.acquire() that will never be released).
-                if self._strategy_task and not self._strategy_task.done():
-                    self._strategy_task.cancel()
-                break
-            self._handle_response(resp)
+            await asyncio.sleep(timeout_s)
```
arekay-nv: [P2] Sleeping a full timeout on every iteration can detect silence almost 2x later than configured: if progress occurs shortly after arming, the first wake observes it and then sleeps another full interval. Compute the remaining duration from `last_progress + timeout` or use an event-driven deadline. That also avoids keeping stale timer work around longer than necessary.

arekay-nv (changes requested):
The opt-in liveness guard is useful, but this implementation adds avoidable work and state to the load generator's hot paths. The most serious issue is retaining every completed UUID in a second phase-lifetime container, which is an O(total requests) memory regression at the repository's 50k+ QPS target. The disabled path also performs work for every response, and low-concurrency workloads can create/cancel a watchdog task per request. Please keep the disabled path inert and use bounded active-request state plus a single event/deadline-driven watchdog.

src/inference_endpoint/config/schema.py
```diff
@@ -648,6 +648,26 @@ def _parse_duration_suffix(cls, v: object) -> object:
     ] = Field(None, gt=0)
     scheduler_random_seed: int = Field(42, description="Scheduler RNG seed")
     dataloader_random_seed: int = Field(42, description="Dataloader RNG seed")
+    no_progress_timeout_s: Annotated[
```
viraatc: recommend rebasing onto latest main, 
then we can instead add this to the new Timeouts model

roborluo: Rebased

roborluo:
Thanks for the comments @viraatc and @arekay-nv . All issues resovled other than one. 

For the per transition watchdog, I choosed a different approche than your suggested ,so flagging it here. 

the guard is still armed  on the 0→1 in-flight transition and retired on drain — but it's now a  single `loop.call_later` TimerHandle instead of an asyncio task. No task,   no Event, no per-iteration `wait_for. 

Also   measured arm+cancel per request, under the eager task factory this repo  uses in production:
 >    Task   (reviewed version):  2205 ns
 >   Timer  (now):                535 ns   → 4.1x

It also leaves the timer heap clear once a phase drains, which covers the  "stale timer work" part of your comment. 

let me know if that make sense to you, Thanks a lot!

src/inference_endpoint/load_generator/session.py
```diff
@@ -566,27 +615,111 @@ async def _drain_inflight(
...
+                    # Cancel a strategy blocked awaiting a semaphore that will
+                    # never be released.
+                    if self._strategy_task and not self._strategy_task.done():
+                        self._strategy_task.cancel()
+                    break
+                # One clock read per response, shared by the event records and
+                # the liveness stamp.
+                now_ns = time.monotonic_ns()
+                self._handle_response(resp, now_ns)
+                if self._no_progress_timeout_ns is not None:
```
viraatc: **High: healthy streaming requests can falsely time out.** Progress advances only for messages reaching the main process, but the production default is `client.stream_all_chunks=false`, so OpenAI/SGLang workers forward only the first `StreamChunk` and the final `QueryResult`. A healthy generation whose first-to-final interval exceeds `no_progress_timeout_s` is therefore aborted despite continuous token output. The new streaming test injects every chunk directly and misses this integration behavior. Please observe liveness where workers see all chunks, or send bounded/coalesced progress heartbeats; forcing all-chunk IPC would add the hot-path overhead this default intentionally avoids.

src/inference_endpoint/load_generator/session.py
```diff
@@ -431,13 +465,17 @@ async def run(
...
+                if self._fatal_error is not None:
+                    raise self._fatal_error
                 if self._stop_requested:
                     break
                 if on_phase_start is not None:
                     on_phase_start(phase)
                 result = await self._run_phase(phase)
                 if result is not None:
                     phase_results.append(result)
+                if self._fatal_error is not None:
```
viraatc, line 477: **Medium: preserve the interrupted report before propagating this failure.** This raises `NoProgressError` out of `BenchmarkSession`; `_run_benchmark_async` drains and builds an interrupted `Report` in `finally`, but then rethrows before returning a `BenchmarkResult`. Consequently `run_benchmark` never calls `finalize_benchmark`, so standard artifacts such as `performance/result_summary.json` and `report.txt` are not written even though `final_snapshot.json` was drained successfully. Carry the fatal outcome alongside the report, finalize it, then return the nonzero outcome.

docs/config/DESIGN.md
```diff
@@ -71,6 +71,12 @@ Immutable snapshot of all parameters needed to execute a run.
 
 Once constructed, `RuntimeSettings` cannot be modified. All consumers receive the same instance.
 
+### No-progress deadline
+
+`settings.timeouts.no_progress_timeout_s` is a client-side liveness deadline, `None` by default. Once a phase has an in-flight request, `BenchmarkSession` fails the run if no streamed chunk or final response arrives within the interval. It does not apply before the server accepts a request, and it does not replace an engine's own KV-transfer deadline.
```
viraatc: **Low: this acceptance boundary is not implemented.** The timer starts in `PhaseIssuer.issue()` immediately after the local issuer/ZMQ push; the client contract has no server-acceptance acknowledgment. Worker scheduling, connection acquisition, and local transport queueing can therefore consume this deadline before the server accepts the request. Please describe it as starting after local issuance, or add a real acceptance signal before advising operators to exclude pre-accept time when sizing the timeout.

viraatc (commented):
Review Council: three actionable findings.

viraatc (approved):

src/inference_endpoint/load_generator/session.py
roborluo: Thought about it.   by default it is off. I think once we start to see stall or hangging anywhere, we can enable it.  So the guidance is to set it generously enough to make sure the server is pretty much hangging already like 10 mins /15mins since we receive last request

src/inference_endpoint/load_generator/session.py
roborluo: You're right — this should match the drain-timeout/interrupt convention

docs/config/DESIGN.md
roborluo: good catch, will fix  the doc

docs/CLI_QUICK_REFERENCE.md
```diff
@@ -300,6 +300,7 @@ settings:
   timeouts:
     run_timeout_s: null # Whole-run watchdog; null = off
     performance_drain_timeout_s: null # null = unlimited
+    no_progress_timeout_s: 30 # Endpoint liveness deadline; null = off
```
nvzhihanj: One small nitpick on the name, recommend more explicit naming so user can understand without reading into the doc. Maybe `endpoint_response_idle_timeout_s` or `endpoint_stream_idle_timeout_s`?

Btw, we should probably instruct that this is better turned off (or a very high value) when running non-streaming cases (e.g. offline run or non-streming benchmark). Note that T2V can takes up to 200-300 seconds for the response to come back on the endpoint.

nvzhihanj: I.e. the default value can be null (if it's not already)

roborluo: Sure, will rename it. 

Also I thought about it. The non streaming part. I would recoomend setting it like 600/900 seconds. Also in TRTLLM, there is something similar checking if the server is stalling or not, And its timeout is 300 seconds.

roborluo: Also the 600/900 seconds recoomended value can be used **regarldess streaming/non-streaming** in my experience.  

I will make it more clear  in case some one set it too small

docs/load_generator/DESIGN.md
```diff
@@ -296,13 +297,17 @@ async def _receive_responses(self):
...
+        if self._no_progress_timeout_ns is not None:
+            self._last_response_progress_ns = time.monotonic_ns()
 ```
 
 Uses `recv()` exclusively — no `poll()` spin. The ZMQ fd is registered with
 the event loop, so `recv()` wakes exactly when a response is available with
 zero CPU overhead. Each `recv()` call yields to the event loop, ensuring
 strategy coroutines (call_at callbacks, semaphore waiters) are never starved.
 
+The no-progress guard rides this same path to catch a client-visible worker or transport stall. It is not a per-request engine timeout: any chunk or final response counts as progress, and only a silent in-flight request fails the run. The receiver only stamps a timestamp, and only when the deadline is set — streaming frame rate is QPS x output length, so per-chunk work is paid per token. The armed `call_later` timer re-derives its own deadline from that stamp when it fires, and re-arms for the remainder if progress raced it.
```
arekay-nv:
```suggestion
The no-progress guard rides this same path to catch a client-visible worker or transport stall. It is not a per-request engine timeout: any chunk or final response counts as progress, and only a fully silent endpoint fails the run. The receiver only stamps a timestamp, and only when the deadline is set — streaming frame rate is QPS x output length, so per-chunk work is paid per token. The armed `call_later` timer re-derives its own deadline from that stamp when it fires, and re-arms for the remainder if progress raced it.
```

src/inference_endpoint/commands/benchmark/execute.py
```diff
@@ -929,6 +935,23 @@ def _on_phase_start(phase: PhaseConfig) -> None:
...
+                        # then run_benchmark fails with this message.
+                        logger.error(
+                            "Endpoint stalled — finalizing partial results "
+                            "before failing the run: %s",
+                            e,
+                        )
+                        stall_error = str(e)
+                        result = SessionResult(
+                            session_id=session_id,
+                            phase_results=[],
```
arekay-nv, line 951: Should we be discarding all the results here? Isn't the intent to salvage the run as much as possible.

arekay-nv (approved):
Thanks for putting this!
Can you add a test that verifies the behavior for one long stuck request while other requests are making progress. From my understanding, this will still wait till all the requests are completed except the stuck one (regardless of how long it stays stuck), and then the timer will trigger after the specified timeout. If that is the case (and correct me if i misunderstood), it seems like overkill to track progress. 
Also, can you clarify the phase data being cleared on timeout is intentional.

src/inference_endpoint/commands/benchmark/execute.py
roborluo: This mirrors the existing run-timeout abort path. We still drain and preserve the metrics report (when usable) and event log; the empty SessionResult only means phase metadata/scoring inputs are unavailable after an aborted session.

roborluo:
> Thanks for putting this! Can you add a test that verifies the behavior for one long stuck request while other requests are making progress. From my understanding, this will still wait till all the requests are completed except the stuck one (regardless of how long it stays stuck), and then the timer will trigger after the specified timeout. If that is the case (and correct me if i misunderstood), it seems like overkill to track progress. Also, can you clarify the phase data being cleared on timeout is intentional.

Yes, that is intentional: this is a session-level liveness guard, not a per-request deadline. While other requests are making progress, the session is still live; once they drain, the remaining stuck request is the only in-flight work and the timeout fires. Tracking progress avoids falsely failing an active session. I’ll add a test for that scenario and clarify the docs.

PR #468, opened 2026-08-25
[docs] Benchmark suite and ruleset proposal

arekay-nv:
## What does this PR do?

Adds a proposed Benchmark Suite spec: a git-versioned catalog of models, datasets, and per-benchmark contracts, with a cohort ruleset that freezes the valid benchmark IDs, suite version, and seed sets. A submission names only cohort, seed set, and benchmark (plus submitter-owned concurrency/endpoints); the client inlines the suite contract and the checker validates the run against it. Worked YAML examples cover DeepSeek-R1 and gpt-oss-120b (including multi-dataset accuracy, generation overrides, and gates). No code changes — this is the design source of truth for a follow-on schema/POC.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass

### DIFF
--- a/docs/bench_suite.md
+++ b/docs/bench_suite.md
@@ -1,0 +1,320 @@
+# Benchmark Suite (status:proposed)
+
+The Benchmark Suite is a Git-versioned record of reusable models, datasets, and the complete contract for each benchmark. A Ruleset is the approved grouping published for one cohort: benchmark IDs, a benchmark suite version, and seed sets. A submission names a cohort, seed set and benchmark, and the checker verifies the run against that benchmark contract.
+
+## Definitions
+
+- Benchmark suite - the git-versioned collection of models, datasets, and benchmarks available in the repository. The suite provides specifications for entire benchmarks (models, datasets, requirements etc), but can also be used for non-submission benchmarking. It is meant as a repository for supported items - everything in the suite at any given time should be supported by the client to conduct benchmarking runs.
+- Cohort - a logical collection of submissions made during a certain duration. A cohort is a submission round defined by an identifier such as `2026-09-C0` . The current proposal is to have two cohorts per month, so `C0` and `C1` are the two cohorts for each month.
+- Ruleset - defines the valid submissions for a cohort. Each ruleset is tied to a cohort and defines the benchmarks that are valid for that cohort. While datasets, models, benchmark updates can happen at any time, the inclusion in a ruleset for a cohort signals that the update is now a valid submission via rolling submission. A ruleset also includes other cohort specific items such as the seed sets that are valid for submissions in that cohort.
+- Phase - each benchmark execution is defined as a series of phases. Currently they can be divided into four groups:
+  - Pre-processing - executed to generate the complete configuration files, as well as perform any sanity checks (for instance current repo matches the SHA for the cohort tag ).
+  - Performance - measured execution of the performance dataset with the user specified concurrency.
+  - Accuracy - correctness checks performed over the accuracy dataset and scored by the designated scorer and aggregated to component scores. Each accuracy dataset can be run multiple times (repeats) and there can be different ways of aggregating the scores from different runs (best of K or average over K). As there can be multiple accuracy datasets, we can also specify aggregation across accuracy datasets (unweighted average), or alternatively leave the scores per dataset separate.
+  - Post-processing - performs integrity checks on the datasets (SHA verification), number of samples issued, model checkpoints. It will also ensure that the accuracy scores satisfies the gates specified in the benchmark specification. There can also be requirements on the performance phase such as minimum duration, minimum number of samples issued, time-to-first-token constraints etc.
+
+## Submitter file
+
+The submitter will use a minimalist file to specify the submission:
+
+```yaml
+config.yaml   # submission configuration; replaces today's submission_ref{model, ruleset}
+  type: submission
+  submission_ref:
+    cohort:  2026-09-C0  # frozen selection (YYYY-MM-C0/C1, submission rules §4.2)
+    seed_set_id: B
+    benchmark:     gpt-oss-120b
+    # TODO - add path overrides.
+  endpoint_config:
+    endpoints: [ https://my-host/v1 ]
+    api_type:  openai           # must equal the benchmark's locked api_type
+    # response_mode comes from benchmark.execution
+  settings:
+    load_pattern: # submission-owned; concurrency is the only supported online type
+      type: concurrency
+      target_concurrency: 64 # one Pareto point; re-run across values to trace the curve
+```
+
+### Folder structure:
+
+The `type: submission` entry in the yaml file above directs the front-end to load the submission details from the benchmark suite which is stored in the following folder structure:
+
+```
+benchmark_suite/
+├── suite.yaml             version; the loader globs the three directories below
+├── models/<id>.yaml       id, reference_id, tokenizer?, processor?, chat_template?,
+│                          metadata{}
+├── datasets/<id>.yaml     id, description, samples, source{url, sha256},
+│                          prompts{}, ground_truth_or_tests{}, evaluation_recipes{}?,
+│                          metadata{}
+└── benchmarks/<id>.yaml   id, family, model, api_type,
+                           execution{}       # response_mode - streaming vs non-streaming
+                           preprocessing{}   # input transform + optional validation hooks
+                           performance{}     # datasets, generation, runtime/sample/latency
+                                             # constraints
+                           accuracy{}        # datasets, recipes, generation,
+                                             # repeat_aggregation
+                                             # dataset_aggregation, thresholds
+                           postprocessing{}  # report, integrity, and compliance gates
+rulesets/<cohort>.yaml
+├── cohort, benchmark_suite_version
+├── seed_sets[]    # id, scheduler_rng_seed, sample_index_rng_seed, model_seed
+└── benchmarks[]   # complete benchmark IDs; no accuracy overrides
+
+
+```
+
+## Sample entries:
+
+### Cohort specification(rulesets/2026-09-C0.yaml):
+
+```yaml
+rulesets/2026-09-C0.yaml
+  cohort: 2026-09-C0                      # submission round; YYYY-MM-C0 (1st Wed) / -C1 (3rd Wed)
+  benchmark_suite_version: "1.0"
+
+  seed_sets:                              # a submission selects exactly one, by id
+    - id: A
+      scheduler_rng_seed:    16159082839903944936
+      sample_index_rng_seed: 2747215439041700203
+      model_seed:            42
+    - id: B
+      scheduler_rng_seed:    <generated at ruleset publication>
+      sample_index_rng_seed: <generated at ruleset publication>
+      model_seed:            <generated at ruleset publication>
+
+  benchmarks:
+    - deepseek-r1
+    - gpt-oss-120b
+
+```
+
+## DeepSeek-R1
+
+### Dataset:
+
+```yaml
+datasets:
+  - id: mlperf_deepseek_r1
+    description: >-
+      Custom dataset curated by MLCommons for DeepSeek R1, specifically for the
+      MLPerf Inference benchmark
+    samples: 4388
+    source:
+      # provenance; not fetch-enforced
+      url: https://<public-host>/mlperf_deepseek_r1_dataset_4388_fp8_eval.jsonl
+      # verified against bytes loaded
+      sha256: <64 hex>
+    prompts: { column: question }
+    ground_truth_or_tests: { column: ground_truth }
+    evaluation_recipes:
+      mlperf_exact_match:
+        scorer: legacy_mlperf_deepseek_r1
+        ground_truth: ground_truth
+```
+
+### Model
+
+```yaml
+models:
+  - id: deepseek-r1
+    reference_id: deepseek-ai/DeepSeek-R1-0528
+```
+
+### Benchmark:
+
+```yaml
+benchmarks:
+  - id: deepseek-r1
+    family: text
+    model: deepseek-r1
+    api_type: openai
+    execution: { response_mode: streaming } # benchmark-wide; applies to every phase
+
+    preprocessing:
+      input_transform: chat_messages
+
+    performance:
+      datasets: [{ dataset: mlperf_deepseek_r1 }]
+      generation:
+        temperature: 0.0
+        top_p: 1.0
+        top_k: 1
+        max_new_tokens: 20000
+      requirements:
+        min_duration_ms: 600000
+        min_sample_count:
+          <TBD> # authoritative online value;
+          # freeze before the Ruleset is published
+
+    accuracy:
+      datasets:
+        - dataset: mlperf_deepseek_r1
+          evaluation_recipe: mlperf_exact_match
+          num_repeats: 1
+      generation:
+        temperature: 0.0
+        top_p: 1.0
+        top_k: 1
+        max_new_tokens: 20000
+      dataset_aggregation: { method: single_dataset }
+      thresholds:
+        exact_match: { reference: 81.3582, lower_factor: 0.99 }
+        TOKENS_PER_SAMPLE:
+          { reference: 3886.2274, lower_factor: 0.9, upper_factor: 1.1 }
+
+    postprocessing:
+      gates:
+        [
+          dataset_integrity,
+          workload_integrity,
+          accuracy_gate,
+          requirements_lock,
+        ]
+```
+
+## GPT-OSS-120b full specification
+
+```yaml
+# MODEL
+models:
+  - id: gpt-oss-120b
+    reference_id: openai/gpt-oss-120b
+
+# DATASETS
+datasets:
+  - id: aime25
+    description: AIME 2025 (AIME2025-I + AIME2025-II, test split)
+    samples: 30
+    source: { url: <TBD>, sha256: <TBD> }
+    prompts: { column: question }
+    ground_truth_or_tests: { column: answer }
+    evaluation_recipes:
+      mlperf_exact_match:
+        {
+          scorer: pass_at_1,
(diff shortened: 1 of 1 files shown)


docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+  - Pre-processing - executed to generate the complete configuration files, as well as perform any sanity checks (for instance current repo matches the SHA for the cohort tag ).
+  - Performance - measured execution of the performance dataset with the user specified concurrency.
+  - Accuracy - correctness checks performed over the accuracy dataset and scored by the designated scorer and aggregated to component scores. Each accuracy dataset can be run multiple times (repeats) and there can be different ways of aggregating the scores from different runs (best of K or average over K). As there can be multiple accuracy datasets, we can also specify aggregation across accuracy datasets (unweighted average), or alternatively leave the scores per dataset separate.
+  - Post-processing - performs integrity checks on the datasets (SHA verification), number of samples issued, model checkpoints. It will also ensure that the accuracy scores satisfies the gates specified in the benchmark specification. There can also be requirements on the performance phase such as minimum duration, minimum number of samples issued, time-to-first-token constraints etc.
```
arav-agarwal2, line 14: The way I'm reading this, the document intends for the load generator to have a ton of submission-related checks.

Who is implementing these checks, and what's the goal of the submission checker given them? Is our goal to verify that these checks were made by the loadgen checker?

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+## Submitter file
+
+The submitter will use a minimalist file to specify the submission:
+
+```yaml
+config.yaml   # submission configuration; replaces today's submission_ref{model, ruleset}
+  type: submission
+  submission_ref:
+    cohort:  2026-09-C0  # frozen selection (YYYY-MM-C0/C1, submission rules §4.2)
+    seed_set_id: B
```
arav-agarwal2, line 25: Note: Target cohort should completely specify a seed set, so I'm unsure what seed set format you're expecting here. Also, what does "A" and "B" mean here? I assume it means the earlier and later available seed sets for a given submission, but it pays to be precise.

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+      target_concurrency: 64 # one Pareto point; re-run across values to trace the curve
+```
+
+### Folder structure:
+
+The `type: submission` entry in the yaml file above directs the front-end to load the submission details from the benchmark suite which is stored in the following folder structure:
+
+```
+benchmark_suite/
+├── suite.yaml             version; the loader globs the three directories below
```
arav-agarwal2, line 44: I"m assuming this is all of the possible configuration for loadgen, which would be specified by the cohort-id.

Where would these details be stored? Do we know who has the responsibility for making sure that this is accurate?

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+├── models/<id>.yaml       id, reference_id, tokenizer?, processor?, chat_template?,
```
arav-agarwal2, line 45: What does <id> refer to here? Is this a model-specific id and, if-so, where would that list be defined?

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+│                          metadata{}
+├── datasets/<id>.yaml     id, description, samples, source{url, sha256},
```
arav-agarwal2, line 47: Same as models

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+│                          prompts{}, ground_truth_or_tests{}, evaluation_recipes{}?,
+│                          metadata{}
+└── benchmarks/<id>.yaml   id, family, model, api_type,
```
arav-agarwal2, line 50: Same as models

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+          scale: percentage
+          reference: 83.13
+          lower_factor: 0.99 # resolved threshold: 82.2987 (display 82.30)
+      requirements:
+        min_accuracy_sample_count: 4395
+
+    postprocessing:
+      gates:
+        [
+          dataset_integrity,
```
arav-agarwal2, line 294: If this needs to be verified by the submission checker after running the benchmark, these files should be present during submission. Is there a PR in the rules repo for getting these into the submission file structure? I couldn't find any.

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+        ]
+```
+
+## Submission flow:
+
+1. Submitter writes a config yaml file that identifies the cohort, benchmark, seed and optional overrides (model paths, dataset paths etc) as well as the client specific settings (endpoint urls, worker configs, network configs, api keys etc).
+2. The front end will perform initial validation on the submission metadata to check valid cohort, benchmark membership, and verify repository versions match the remote tags for that cohort.
+3. Once validated, the specification is inlined into the yaml file to reveal an expanded configuration file which specifies exactly what needs to be executed. This ensures that the benchmark execution is correct by construction. The generated yaml file will be saved as part of the output artifacts.
+4. The generated yaml is passed on to the execution phases which will run the phases in order and produce associated artifacts. For instance, the dataloader will produce an implementation specific SHA for the data loaded which will be logged.
+5. Towards the end of the execution of the benchmark phases, post-processing step will apply the checks needed to verify submission integrity:
```
arav-agarwal2, line 307: Is post-processing something that is done by the submission checker as created by MLC, or is this done by loadgen? If the former, we need rules added to the rules repo so everyone knows what checks are being done on submissions.

arav-agarwal2 (commented):
I don't think I can approve it just yet as it's a proposal, but my main worry is that I don't see a policies PR suggesting the changes to the submission directory and submission checker rules for this system.

arav-agarwal2:
Is this intended for 1.0? If so, this needs to be far more detailed in terms of checks and overall schema for each of the files, and there needs to be accompanying rules PRs that another organization agrees with.

arekay-nv:
> Is this intended for 1.0? If so, this needs to be far more detailed in terms of checks and overall schema for each of the files, and there needs to be accompanying rules PRs that another organization agrees with.

Yes, this is intended for 1.0. But this is purely an implementation change, nothing from the rules needs to change. The goal is to 1) centralize all the benchmark specification - right now we do not have a single source of truth for benchmarks in a given round/cohort 2) ensure that we can capture all aspects of what the existing rules specify to be automated and 3) enable us to add new benchmarks rapidly by updating the specification. 

There are two sides to the verification of a submission - client side (performed before submission) and MLC side (after submission). Both should use the same yaml configs but should be independent so that we can more robust and identify any disambiguation. 

This change is designed not have any rules PR and will be an implementation only change unless we find gaps in the specification and we can then file issues/PRs for rules.

arekay-nv:
> I don't think I can approve it just yet as it's a proposal, but my main worry is that I don't see a policies PR suggesting the changes to the submission directory and submission checker rules for this system.

This shouldn't affect the submission directory structure or the submission checker rules. Those _should_ remain the same.

docs/bench_suite.md
arekay-nv: We already have some post-processing steps in the existing setup (steady state checks) and some expected to be added (steady state). The goal is to ensure that a submission is as close to compliance as possible and there are no issues that are overlooked (incorrect checkpoint, insufficient number of samples etc). 
Submission checker would independently make those checks based on the benchmark specification which is more reliable as it is done on MLC side compared to the loadgen side checks which can be bypassed.

docs/bench_suite.md
arekay-nv: There is a cohort definiton example below that includes seeds
```
seed_sets:                              # a submission selects exactly one, by id
    - id: A
      scheduler_rng_seed:    16159082839903944936
      sample_index_rng_seed: 2747215439041700203
      model_seed:            42
    - id: B
      scheduler_rng_seed:    <generated at ruleset publication>
      sample_index_rng_seed: <generated at ruleset publication>
      model_seed:            <generated at ruleset publication>

```

docs/bench_suite.md
arekay-nv: Yes, this would be a model specific id. Note that this is the list. We would not need to maintain a separate document as the models specified here are what can be used in a benchmark, which is what can be specified in a cohort for submission.

docs/bench_suite.md
arekay-nv: Same as above. Datasets defined here are what can be used in a benchmark and then in a cohort.

docs/bench_suite.md
arekay-nv: This is one level below a cohort. A cohort will define the set of benchmarks for submission.

docs/bench_suite.md
arekay-nv: Do you mean `dataset_integrity` and peers? These are passes that will need to be implemented. But this is similar in structure to what we have for scorers/extractors for accuracy datasets.

docs/bench_suite.md
arekay-nv: These are independent of cohort. A cohort only defines which of these is valid for a submission into that cohort. 
A task force will work independently to update the model/dataset/benchmark specs for its benchmark, and then merge those changes in. Then propose to open it up for submission at which point it will be added to the next cohort.

docs/bench_suite.md
anandhu-eng: @arekay-nv ,  If we are blocking a submission due to any of the post-processing steps/checks by the submission checker, i think it would be better to mention the checks in the rules so that submitters would be aware of it. They probably would be searching for these guardrails in endpoints policies repo rather than the endpoints package repo

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+    # response_mode comes from benchmark.execution
+  settings:
+    load_pattern: # submission-owned; concurrency is the only supported online type
+      type: concurrency
+      target_concurrency: 64 # one Pareto point; re-run across values to trace the curve
+```
+
+### Folder structure:
+
+The `type: submission` entry in the yaml file above directs the front-end to load the submission details from the benchmark suite which is stored in the following folder structure:
```
anandhu-eng, line 40: Does the benchmark suite come along with the current endpoints repo, or are we planning it as a seperate rpository?

docs/bench_suite.md
nv-alicheng, line 45: What is 'processor' here?

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+        min_sample_count: 6396
+        min_context_length: 65536
+        max_ttft_latency_ms: 3000
+        max_tpot_latency_ms: 80
+
+    accuracy:
+      datasets:
+        - dataset: aime25
+          evaluation_recipe: mlperf_exact_match
+          generation: # sparse merge over the phase generation block
```
nv-alicheng, line 260: I'm iffy about sparse merging - it's bitten us in the past when doing sparse merges on configs.

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+      dataset_aggregation: { method: single_dataset }
+      thresholds:
+        exact_match: { reference: 81.3582, lower_factor: 0.99 }
+        TOKENS_PER_SAMPLE:
+          { reference: 3886.2274, lower_factor: 0.9, upper_factor: 1.1 }
+
+    postprocessing:
+      gates:
+        [
+          dataset_integrity,
```
nv-alicheng, line 169: Where would these gates be implemented?

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+        - dataset: mlperf_deepseek_r1
+          evaluation_recipe: mlperf_exact_match
+          num_repeats: 1
+      generation:
+        temperature: 0.0
+        top_p: 1.0
+        top_k: 1
+        max_new_tokens: 20000
+      dataset_aggregation: { method: single_dataset }
+      thresholds:
```
nv-alicheng, line 161: Where do the keys in this block come from? Can we have a list of valid thresholds? When are these threshold checked and by what?

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+
+The `type: submission` entry in the yaml file above directs the front-end to load the submission details from the benchmark suite which is stored in the following folder structure:
+
+```
+benchmark_suite/
+├── suite.yaml             version; the loader globs the three directories below
+├── models/<id>.yaml       id, reference_id, tokenizer?, processor?, chat_template?,
+│                          metadata{}
+├── datasets/<id>.yaml     id, description, samples, source{url, sha256},
+│                          prompts{}, ground_truth_or_tests{}, evaluation_recipes{}?,
```
nv-alicheng, line 48: ground_truth_or_tests{} can be misinterpreted - can you do a similar breakdown with descriptions of each field like you did below with the benchmarks yaml?

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+    reference_id: openai/gpt-oss-120b
+
+# DATASETS
+datasets:
+  - id: aime25
+    description: AIME 2025 (AIME2025-I + AIME2025-II, test split)
+    samples: 30
+    source: { url: <TBD>, sha256: <TBD> }
+    prompts: { column: question }
+    ground_truth_or_tests: { column: answer }
```
nv-alicheng, line 191: Based on the example here, this is a dict with either the key 'column' or 'tests', which is never explicitly stated anywhere, and the name of the value implies either this is a state enum, or that the keys will be 'ground_truth' or 'tests'. Can you rename and add some docs here?

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+## GPT-OSS-120b full specification
+
+```yaml
+# MODEL
+models:
+  - id: gpt-oss-120b
+    reference_id: openai/gpt-oss-120b
+
+# DATASETS
+datasets:
```
nv-alicheng, line 185: I'm curious, how would an agentic dataset be written in this yaml? Would you need to differentiate between turns and trajectories?

docs/bench_suite.md
nvzhihanj: Hi folks, the purpose is to build clarity in the repo what is valid in a certain period of time (cohort) and this will not be captured in the policy repo

The pain point now is many new submitters or even old submitters are confused what a valid submission is, if they don't attend meeting weekly. It cannot be understood easily by reading the policy repo or the submission checker. Implementing the structure to constrain a "valid submission suite" will help us avoid many human mistakes during the submission process.

nvzhihanj: The post-submission checker should still be implemented, but they can assume the definition in these implemented ruleset and benchmark suite files, so we have one single source of truth to be quoted.

docs/bench_suite.md
nvzhihanj: Just to be clear, we are not rotating seed per cohort right (would be helpful to link this to where we define how frequent the seed is changing)

I vaguely remember we discuss per 2-month or per quarter but I couldn't find it

docs/bench_suite.md
nvzhihanj: What does suite.yaml define? From the comment I am not too sure

docs/bench_suite.md
```diff
@@ -0,0 +1,320 @@
...
+```
+
+## Sample entries:
+
+### Cohort specification(rulesets/2026-09-C0.yaml):
+
+```yaml
+rulesets/2026-09-C0.yaml
+  cohort: 2026-09-C0                      # submission round; YYYY-MM-C0 (1st Wed) / -C1 (3rd Wed)
+  benchmark_suite_version: "1.0"
```
nvzhihanj, line 74: Is this 1.0 corresponds to endpoints 1.0? 
Is the suite supposed to change dynamically or stay the same in 1-2 cohorts? (Asking mainly to clarify what suite.yaml specifies)

docs/bench_suite.md
arekay-nv: Yes, it is a two-cohort seed [rotation policy](https://github.com/mlcommons/endpoints_policies/pull/102)

docs/bench_suite.md
arekay-nv: It will be part of the endpoint repo. That would enable us to snapshot the benchmark specification for a cohort by tagging a branch and there are no external dependencies to reproduce a run.

docs/bench_suite.md
arekay-nv: For now it is just a container and contains the version of the suite schema. If we make breaking changes to it, we can then update the version.

docs/bench_suite.md
arekay-nv: No, this is the suite schema version. Probably over-complicating things for now and we can assume there is a single version-less schema. Will take out.

docs/bench_suite.md
arekay-nv: Both - the specification should be sufficient to have both independently implement a correct checker and agree on the outcomes. If they don't we need to either identify the buggy implementation and fix it or harden the specification.

docs/bench_suite.md
arekay-nv: The way it is currently implemented, loadgen doesn't need to deal in turns. The agentic issuer is given trajectories and it would handle the issuing of turns for each trajectory before moving on.

docs/bench_suite.md
arekay-nv: As post-processing checks in the client for submissions.

docs/bench_suite.md
arekay-nv: Right now they are just understood to be there - but we can expose those from the metrics implementation and use a consistent naming scheme across them. The metrics recorder would have all the computed metrics available to the checkers which can express those constraints.

PR #479, opened 2026-08-26
fix(metrics): weight gpt-oss accuracy by dataset sample count to matc…

arekay-nv:
The combined gpt-oss-120b accuracy was an unweighted mean of the three per-dataset scores (aime25/gpqa/livecodebench). That over-weights small datasets (AIME's 30 problems counted equally with LiveCodeBench's 1055) and diverges from the MLCommons reference submission score. Switch to a unique-problem-count weighted mean:

    overall = sum(score_d * unit_samples_d) / sum(unit_samples_d)

which reproduces the reference exact_match (83.374 vs the old 80.95 on the same run). Rename the aggregator to samples_weighted_average_accuracy; the accuracy_results.json key stays "average_accuracy" for consumer stability.

Also fix the gpt-oss accuracy plot (results_plots.py), which still computed an unweighted overall and would otherwise disagree with the report/JSON headline.

- Single-dataset models (DeepSeek-R1, BFCL, ...) are unchanged (weight cancels).
- Weight guard: absent/None unit_samples -> 1.0 (legacy artifacts); a present but non-positive/non-numeric weight is treated as corrupt and skipped.
- Tests: MLPerf reference lock (83.374), plot weighting, guard branches, and the average_accuracy JSON-key contract.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor/cleanup

## Testing

- [ ] Tests added/updated
- [ ] All tests pass locally

## Checklist

- [ ] Pre-commit hooks pass

### DIFF
--- a/src/inference_endpoint/commands/benchmark/accuracy.py
+++ b/src/inference_endpoint/commands/benchmark/accuracy.py
@@ -41,7 +41,9 @@ from inference_endpoint.async_utils.services.metrics_aggregator.token_metrics im
 from inference_endpoint.config.schema import DatasetType, ScorerMethod, TestMode
 from inference_endpoint.dataset_manager.dataset import Dataset
 from inference_endpoint.evaluation import Extractor
-from inference_endpoint.evaluation.accuracy_results import average_accuracy
+from inference_endpoint.evaluation.accuracy_results import (
+    samples_weighted_average_accuracy,
+)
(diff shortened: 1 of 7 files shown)


zihaok (approved):
lgtm, Thanks!

nvzhihanj (approved):

PR #493, opened 2026-09-03
feat(metrics): report E2E average interactivity

### LINKED ISSUE #492: [Feature]: Add agentic interactivity metric
### Motivation

Agentic workloads are going to use a different metric definition for interactivity `e2e_avg_interactivity` defined by sum(OSL over all turns)/sum(E2EL over all turns) which is a scalar.

### Proposed Solution

Generate the agentic interactivity as a separate metric reported for agentic workloads in the output report.

### Alternatives Considered

_No response_

### Additional Context

_No response_

hvagadia:
## Summary

- compute `e2e_avg_interactivity` for every report as total output tokens divided by summed sample latency in seconds
- include the metric in `performance/result_summary.json` and the human-readable summary

## Testing

- `uv run pytest -m unit tests/unit/metrics`
- `uv run pre-commit run --all-files`

Closes #492

### DIFF
--- a/tests/unit/metrics/test_report_builder.py
+++ b/tests/unit/metrics/test_report_builder.py
@@ -209,6 +209,7 @@ class TestFromSnapshot:
         # No duration -> from_snapshot leaves throughput unset.
         assert report.qps is None
         assert report.tps is None
+        assert report.e2e_avg_interactivity is None
         # Series with count==0 should produce empty dicts.
         assert report.ttft == {}
         assert report.latency == {}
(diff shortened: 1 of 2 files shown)


tests/unit/metrics/test_report_builder.py
```diff
@@ -369,7 +372,9 @@ def test_display_tps_na_when_unmeasured(self):
         report = _build_report(_make_registry(n_samples=0))
         lines: list[str] = []
         report.display(fn=lines.append, summary_only=True)
-        assert "TPS: N/A" in "\n".join(lines)
+        output = "\n".join(lines)
+        assert "TPS: N/A" in output
+        assert "E2E average interactivity: N/A" in output
```
arekay-nv, line 396: Can you add a test to cover failed requests as well - i think the sample_latency would overcount the latencies (all latencies including those that failed) but the OSL doesn't account for that - it only counts successful responses. So the two parts of the ratio mismatch in size.

arekay-nv (commented):
Please also update the docs to refer to the added field.

nvzhihanj (approved):

leopck (approved):

PR #495, opened 2026-09-04
fix: retry transient SWE-bench launch failures

zihaok:
## What does this PR do?

This PR introduces retry logic to handle the swe bench service pyxis runtime's possible failure. http request failure, pyxis runtime related environment failure. This doesn't help model to achieve higher score, it simply makes the evaluation environment stable and reproducible. 

## Type of change

- [X] Refactor/cleanup

## Testing

- [X] Tests added/updated
- [X] All tests pass locally
- [X] Manual testing completed

## Checklist

- [ ] Pre-commit hooks pass

### DIFF
--- a/src/inference_endpoint/evaluation/swe_bench_scorer.py
+++ b/src/inference_endpoint/evaluation/swe_bench_scorer.py
@@ -32,6 +32,8 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 _SAFE_RUN_ID_RE = re.compile(r"^[A-Za-z0-9_-]+$")
+_HTTP_TIMEOUT_MAX_ATTEMPTS = 10
+_HTTP_TIMEOUT_RETRY_DELAY_S = 5.0
 
 
 class _RejectRedirectHandler(urllib_request.HTTPRedirectHandler):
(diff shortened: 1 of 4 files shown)


nvzhihanj (approved):

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
```diff
@@ -137,28 +153,57 @@ def run_srun_step(
...
+                failure_path.touch()
+            raise RunnerError(
+                "Pyxis infrastructure failure before the command completed"
+            ) from exc
+
+        status = status_path.read_text().strip()
+        if status == f"finished:{result.returncode}":
+            return result
+
+        output = (result.stdout or "").strip()
```
arekay-nv: Read from stderr as well since some pyxis errors would be on the stderr instead of stdout or pipe the output in the call above.

src/inference_endpoint/evaluation/swe_bench_scorer.py
```diff
@@ -181,19 +183,30 @@ def _http_json(
...
-            ) from exc
-        except urllib_error.URLError as exc:
-            raise SetupError(
-                f"SWE-bench service is unreachable at {url}: {exc.reason}"
-            ) from exc
+        for attempt in range(1, _HTTP_TIMEOUT_MAX_ATTEMPTS + 1):
+            try:
+                with _NO_REDIRECT_OPENER.open(req, timeout=timeout_s) as resp:
+                    body = resp.read()
+                break
```
arekay-nv: Wouldn't this create duplicate requests - do we know that the timeout happened because the request was terminated?

hvagadia (approved):

src/inference_endpoint/evaluation/swebench_service/swebench_service/pyxis_environment.py
zihaok: add stderr

src/inference_endpoint/evaluation/swe_bench_scorer.py
zihaok: removed, turns out that this is not needed. to make pyxis environment robust

PR #504, opened 2026-09-11
feat(metrics): full-run OSL over all agentic turns for accuracy validation

### LINKED ISSUE #500: Agentic inference: include tail turns in mean OSL used for accuracy validation
Agentic inference stops performance tracking when the first active user finishes its final assigned trajectory. Turns issued before this boundary remain included, even if they complete later; turns issued afterward are excluded from performance metrics. Official runs continue executing the remaining turns for accuracy. See [Tail Management](https://github.com/mlcommons/endpoints/blob/main/examples/10_Agentic_Inference/README.md#tail-management).

However, the reported mean output sequence length (OSL) is calculated over the performance window, while per-turn mean OSL is also an [accuracy requirement](https://github.com/mlcommons/endpoints/blob/main/examples/10_Agentic_Inference/README.md#accuracy).

The number and composition of excluded tail turns depend on concurrency. Consequently, different concurrency points can report different mean OSL values simply because they include different subsets of turns—even if the model produces identical outputs for every turn. This introduces sampling bias into the accuracy check and can affect whether a Pareto point meets the required OSL range.

We should calculate the mean OSL used for accuracy validation over **all generated turns in the configured workload, including the tail**.

Proposed behavior:

- Preserve the existing performance cutoff and performance-window OSL for throughput calculations.
- Report a separate full-run OSL statistic after all configured trajectories finish, including any dataset repeats.
- Use consistent token-counting rules for both statistics, including reasoning and tool calls.
- Document which field to use for the OSL accuracy requirement, and clearly identify incomplete runs.
- Add a regression test using fixed per-turn outputs at different concurrencies: full-run mean OSL must remain identical even when the performance-window mean changes.

The full-run mean should be calculated as total output tokens divided by total included assistant turns, rather than averaging per-user or per-trajectory means.

arekay-nv:
## Summary

Resolves #500.

Agentic-inference runs stop performance tracking when the first active user finishes its final trajectory, so turns issued after that boundary are excluded from the windowed output-sequence-length (OSL). Because the excluded tail depends on concurrency, the reported mean OSL differs across Pareto points even when the model produces identical output — which biases the **per-turn mean OSL accuracy requirement** (see the [Tail Management](examples/10_Agentic_Inference/README.md#tail-management) section).

This PR adds a **separate full-run OSL** computed over **every** completed turn — tail and dataset repeats included — for use as the accuracy-gate statistic, while leaving the windowed OSL untouched for throughput.

## What changed

- **`commands/benchmark/full_run_osl.py`** (new): a post-finalize, streaming, prefix-gated scan of `events.jsonl` that rolls up OSL over all in-population COMPLETE turns. The population is bounded to the performance phase via `sample_idx_map.json` and **fails closed** (skips) when that map is unavailable, so accuracy-phase turns can never be counted as performance OSL. Reports `n_turns_counted` / `n_empty` / `n_errors` / `n_undecodable` so a partial or all-blank run stays visible rather than looking uncomputed.
- **`metrics_aggregator/tokenization.py`**: `output_tokenization_input` — the single shared rule for how a model output (including reasoning and tool calls) is rendered for counting. Both the windowed `OslTrigger` and the full-run scan use it, so the two numbers are directly comparable.
- **`metrics_aggregator/token_metrics.py`**: `BatchTokenizer.count_sync` for finalize-side (post-event-loop) counting that matches the async path exactly; teardown is guarded so the in-process (`n_workers=0`) tokenizer never terminates unrelated child processes.
- **`commands/benchmark/execute.py`**: computes and attaches the block at finalize, inside the report-writing `try/finally` (skipped on abort so a partial tail is never reported as complete).
- **`metrics/report.py`**: new optional `output_sequence_lengths_full_run` field, plus a `report.txt` line labeled as the accuracy OSL (with an explicit note when it is missing or the run is incomplete).
- **`examples/10_Agentic_Inference/README.md`**: documents `output_sequence_lengths_full_run.output_sequence_lengths.avg` as the field to read for the OSL accuracy gate — a run is invalid for the gate when this is `null`, `n_turns_counted == 0`, or the run is not `complete`.

## Testing

```
uv run pytest -m unit \
  tests/unit/commands/test_full_run_osl.py \
  tests/unit/metrics/test_report_builder.py \
  tests/unit/async_utils/services/metrics_aggregator/test_output_tokenization_rule.py
```

Includes the issue #500 (f) regression guard: the full-run mean stays identical when the performance window closes at different points, even as the windowed mean moves. (Pre-existing macOS-only IPC-socket / CPU-affinity test failures are unrelated to this change.)

## Notes

- The single `Sync` commit can be squashed/renamed on merge.
- Draft: opening for review. A few optional test-hardening follow-ups (an isolation test for a raising `output_tokenization_input`, a direct test of the tokenizer-teardown guard) are not yet included.

### DIFF
--- a/src/inference_endpoint/commands/benchmark/full_run_osl.py
+++ b/src/inference_endpoint/commands/benchmark/full_run_osl.py
@@ -1,0 +1,231 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Full-run OSL — output token lengths over every turn, including the tail.
+
+The performance window closes when the first user finishes its final assigned
+trajectory, and turns issued after that point never reach the metrics
+aggregator at all (they are dropped at row creation). The windowed OSL is
+therefore a concurrency-dependent subsample: at higher concurrency the window
+closes earlier, so a different — and systematically shorter — subset of turns
+is measured, even when the model produces identical output for every turn.
+
+This module recomputes OSL over every completed turn from ``events.jsonl``,
+which retains the full ``TextModelOutput`` (reasoning and tool calls included)
+regardless of tracking state. Counting goes through the same
+``extract_tokenization_input`` rule the perf-side ``OslTrigger`` uses, so the
+two statistics are comparable rather than merely adjacent.
+
+See mlcommons/endpoints#500.
+"""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Callable, Collection
+from pathlib import Path
+from typing import Any
+
+import msgspec
+
+from inference_endpoint.async_utils.services.metrics_aggregator.token_metrics import (
+    BatchTokenizer,
+)
+from inference_endpoint.async_utils.services.metrics_aggregator.tokenization import (
+    MessageInput,
+    TextInput,
+    extract_tokenization_input,
+)
+from inference_endpoint.core.record import EventRecord, EventType
+from inference_endpoint.core.types import TextModelOutput
+from inference_endpoint.metrics.report import series_metric_dict
+
+logger = logging.getLogger(__name__)
+
+# Matched against each line's leading bytes so the 6+ GB of prompt and response
+# payload in a large run is never decoded just to classify the record.
+_COMPLETE_MARKER = b'"event_type":"sample.complete"'
+_PREFIX_BYTES = 256
+# AgenticInferenceInlineScorer's dataset_name; the key its turns live under in
+# sample_idx_map.json.
+_PERFORMANCE_PHASE = "performance"
+
+TokenCounter = Callable[[MessageInput | TextInput], int]
+
+
+def _in_population(
+    record: EventRecord,
+    performance_uuids: Collection[str] | None,
+) -> bool:
+    """Whether one COMPLETE record belongs to the full-run OSL population.
+
+    This is the policy boundary for "all turns", and the one place to change
+    if that definition moves.
+
+    ``events.jsonl`` is written for the whole run, so it also holds the
+    accuracy phases' completions (SWE-bench). Bounding the population by the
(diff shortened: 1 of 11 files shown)


leopck (approved):
tested with high concurrency runs previously was heavily impacted after this PR fix, my mean OSL is stable on high concurrency and matching with low concurrency runs

src/inference_endpoint/commands/benchmark/full_run_osl.py
```diff
@@ -0,0 +1,231 @@
...
+            # Output preparation stays inside the per-turn try: a malformed
+            # tool-call payload can raise in extract_tokenization_input, and one
+            # bad turn must not void the whole statistic.
+            try:
+                tokenization_input = extract_tokenization_input(record.data)
+                if tokenization_input is None:
+                    n_empty += 1
+                    continue
+                lengths.append(count_tokens(tokenization_input))
+            except Exception as e:  # noqa: BLE001 - one bad turn must not void the stat
```
hvagadia, line 147: Could we make partial tokenization explicit in the validity of the full-run OSL metric?

The `n_errors` / `n_undecodable` counters and warning are useful, but the mean is still calculated from the surviving turns and displayed as "accuracy, all turns". The documented gate checks only that the block exists, `n_turns_counted > 0`, and the report is `complete`, so it can accept a partial mean.

I reproduced two outputs of 400 and 800 tokens (correct mean: 600): failing to count the second produces a mean of 400 with `n_errors == 1` and `report.complete == true`. Removing the second COMPLETE record produces the same mean with all diagnostic counters at zero, even though its UUID remains in the performance map.

Please preserve the partial mean/counters for diagnostics, but expose a machine-readable completeness/validity indicator on this metric, clearly label partial OSL in `report.txt`, and exclude it from the OSL accuracy gate. Reconciling observed completion UUIDs against the expected performance population would also expose missing records. This can leave the benchmark/performance report available while making it clear that the full-run OSL is unusable for accuracy. A regression covering a failed or missing turn would pin this behavior.

hvagadia (approved):

PR #506, opened 2026-09-15
feat(deps): add rouge extra for ROUGE scoring deps

anandhu-eng:
## Problem

`RougeScorer` (`eval_method: "rouge"`) lazily imports `nltk`, `evaluate` and `rouge_score` and raises an `ImportError` telling users to install them — but no extra ever declared them. The three shipped Llama configs using `eval_method: "rouge"` fail at scorer construction on a stock install, and the only install instructions lived in `examples/05_Llama_Examples/README.md`.

## Change

- New `rouge` extra declaring `nltk==3.10.3`, `evaluate==0.4.6`, `rouge-score==0.1.2`.
- `test` extra inherits it via `inference-endpoint[rouge]` — same self-reference pattern as `inference-endpoint[sql]` — so CI covers the import surface with no extra job.

Resolution is clean: 4 new packages, `evaluate` pulls the already-pinned `datasets==5.0.1`, and **no existing pin moves**. Unlike `bfcl`, no `[tool.uv].conflicts` fork is needed.

## Verified

- `uv lock` adds only `absl-py`, `evaluate`, `nltk`, `rouge-score`; no version changes to existing packages.
- `uv sync --frozen --extra rouge` succeeds; all three import at the pinned versions.
- The lazy-import guard in `evaluation/scoring.py` now resolves (`_evaluate`/`_nltk` non-`None`), so the `ImportError` path is no longer hit.

## Note

Installing the packages is not sufficient for an **offline** image: `nltk` needs the `punkt`/`punkt_tab` corpora and `evaluate.load("rouge")` fetches its metric script from the HF Hub. Prefetching both is handled separately in `Dockerfile.dev`.


attafosu (approved):
LGTM

arekay-nv (approved):
Thanks!

PR #508, opened 2026-09-17
feat(config): register MLPerf Endpoints v1.0 seed set as a ruleset round

arekay-nv:
## Summary

Registers the published **MLPerf Endpoints v1.0 seed set** (cohort `2026-10-C1`, set A) as a ruleset round, so a submission can bind to the official seeds through `submission_ref.ruleset`.

Seeds come from [`seedset.yaml`](https://github.com/mlcommons/endpoints_policies/blob/5279b845b8492b02742a66d31cee09d9512f7e1d/seedset.yaml) in `mlcommons/endpoints_policies` (added by [endpoints_policies#117](https://github.com/mlcommons/endpoints_policies/pull/117)), pinned to a commit rather than a branch so the transcription stays re-verifiable.

| Registry name | Round |
| --- | --- |
| `mlperf-endpoints-v1.0-2026-10-C1-A` | cohort-and-set-qualified |
| `mlperf-endpoints-current` | convenience alias |

No config-schema changes: the seeds reach `settings.runtime` through the existing `_apply_ruleset_seed_overrides` path that already pins the MLPerf Inference round seeds.

## Design notes

**Endpoints rotates independently of the Inference rounds.** MLCommons publishes a seed set every two publication cohorts, each set is adoptable for four, and a submission keeps its bound set for its full Pareto-update window. `ENDPOINTS_ALL` therefore registers *every* published set rather than replacing the previous one — an older set must keep resolving after a newer one is published, or a submission already bound to it stops loading.

**The version string carries both the cohort ID and the set ID.** `seed_sets` is a list keyed by `id` ("a submission selects exactly one, by id"), so a cohort may publish more than one set. Registration uses `setdefault`, so a second set in the same cohort would otherwise be silently dropped and resolve to whichever came first.

**`benchmark_rulesets={}`.** Endpoints gates a Pareto curve rather than per-model TTFT/TPOT, so there are no per-model rules to declare. The legacy per-model `apply_user_config` path correctly refuses this round; there is a test pinning that.

## `model_seed` is deliberately not pinned

The cohort publishes a third seed, `model_seed: 9315206023656308754`, which this PR records as a comment but does not wire up. Its purpose is **undocumented upstream** — the string appears in neither `endpoints_rules.md` nor `endpoints_submission_rules.md` at the pinned commit, only in `seedset.yaml`, which gives no definition.

Submission rules §4.6 says the set drives *"request-issue / sample order, **and the per-query salt**"* — three RNG purposes for three published seeds. This round maps the first two, which leaves the per-query salt RNG unaccounted for.

**Open question for the WG:** what does `model_seed` seed? If it is the salt seed, `endpoints_rules.md` §2.1.1 has a MUST-level requirement currently unmapped by the reference client. (Separately: the value is a 64-bit *unsigned* integer, so if it is intended as an OpenAI `seed` request field it exceeds the signed-int64 range that field accepts.)

## Testing

10 new tests in `tests/unit/config/rulesets/mlcommons/test_rules.py`:

- Seed values match the published set; registered under both names; every published set stays registered; cohort versions unique
- Seeds differ from every other registered ruleset (`ALL_ROUNDS` + `EDGE_CURRENT`) — catches a copy-paste from the wrong source
- End-to-end: binding `submission_ref` to either registry name pins the seeds onto `settings.runtime` and reseeds warmup, and survives a `config.yaml` round-trip. `scheduler_rng_seed` exceeds int64, so this is the branch where a future bound on that field would break this round while every other test stayed green
- The per-model `apply_user_config` path refuses the round
- `ENDPOINTS_CURRENT` is re-exported from the package

`uv run pytest` passes in full. `pre-commit run --all-files` passes except three pre-existing macOS-only mypy errors (`os.sched_setaffinity` / `os.sched_getaffinity` in `cpu_affinity.py` and `token_metrics.py`) in files this PR does not touch.

Seed constants were verified digit-for-digit against the pinned upstream blob.

## Second commit: loud failure on an unregistered ruleset

An unregistered `submission_ref.ruleset` previously raised only for `type: submission`; every other test type logged a warning and carried on with the default seeds `42`/`42`. That hides a typo in exactly the case it matters — the run reports itself as bound to a ruleset, dumps a `config.yaml` naming it, and issues load from seeds nobody chose.

It is now an error for every test type, and the message lists the registered rulesets:

```
submission_ref.ruleset 'mlperf-endpoints-v1.0-2026-10-C1' is not registered, so the
run cannot be bound to its RNG seeds and would silently fall back to the defaults.
Registered rulesets: mlcommons-current, mlperf-edge-current, mlperf-edge-v0.1,
mlperf-endpoints-current, mlperf-endpoints-v1.0-2026-10-C1-A, mlperf-inference-v5.1,
mlperf-inference-v6.1
```

The near-miss shown (`-C1` vs the registered `-C1-A`) is the realistic mistake this guards against, and is why the alternatives are listed rather than only the rejected value.

This affects **all** rulesets, not only the Endpoints seed set. No YAML template or example referenced an unregistered ruleset, so nothing on disk changes behaviour. Two existing tests that use a placeholder ruleset name are `type: submission` cases asserting a `benchmark_mode` error, which raises before the ruleset lookup and so are unaffected; the one test that asserted the old leniency is replaced with parametrized offline/online failure tests.

## Verification on a cluster

Ran the branch on an `aga` login node against the built-in echo server (no GPU needed — seed pass-through happens during config resolution). The input config deliberately set **wrong** seeds, so this exercises override rather than mere presence:

| | Input config | Dumped `config.yaml` |
| --- | --- | --- |
| `scheduler_random_seed` | `7` | `10487924139932647040` |
| `dataloader_random_seed` | `9` | `586478644936801402` |
| `warmup_random_seed` | unset | `586478644936801402` (reseeded from sample-index) |

Both registry names resolve and pin identically (`mlperf-endpoints-current` and `mlperf-endpoints-v1.0-2026-10-C1-A`), each `exit 0` with 20/20 samples. A third run omitting the seeds entirely produced the same pinned values, confirming a submission config need only name the ruleset. `performance/result_summary.json` carries the same seeds in its `run_config` block.


nvzhihanj (approved):

zihaok:
Hi Rashid, can you post an example of how ruleset is specified in endpoint.yaml

arekay-nv:
> Hi Rashid, can you post an example of how ruleset is specified in endpoint.yaml

Here is what i used to test the seeds getting passed through:
```
name: seed-passthrough-test
type: offline

submission_ref:
  model: test-model
  ruleset: mlperf-endpoints-v1.0-2026-10-C1-A    # ← the only thing that matters

model_params:
  name: test-model
  max_new_tokens: 8
  streaming: "on"

endpoint_config:
  endpoints:
    - http://127.0.0.1:8765
  api_type: openai

datasets:
  - path: tests/assets/datasets/dummy_1k.jsonl
    type: performance
    parser:
      prompt: text_input

settings:
  runtime:
    n_samples_to_issue: 20
    scheduler_random_seed: 7      # deliberately wrong
    dataloader_random_seed: 9     # deliberately wrong
  client:
    num_workers: 1
    max_connections: 8
    warmup_connections: 0
```

